PR #11, opened 2015-12-03
two proposals for more elegant syntax

pinxue:
1. use $(expr) to replace \(expr) for string literal interpolation.
2. add const as an alias for let.

### DIFF
--- a/proposals/0008-add-const-as-alias-of-let.md
+++ b/proposals/0008-add-const-as-alias-of-let.md
@@ -1,0 +1,30 @@
+# Add const as alias of let
+
+* Proposal: [SE-0008](https://github.com/apple/swift-evolution/proposals/0008-add-const-as-alias-of-let.md)
+* Author(s): [Wu Yang](https://github.com/pinxue)
+* Status: **Review**
+* Review manager: TBD
+
+## Introduction
+
+Add const as alias of let.
+
+## Motivation
+
+const is more related to var than let, which may makes source code a little easier to read.
+
+## Proposed solution
+
+Add const as an alias of let.
+
+## Detailed design
+
+Programmer may write const instead of let in source code. Mixing of const and let is forbidden though to avoid confusion.
+
+## Impact on existing code
+
+It is harmless new syntax sugar, thus no impacts.
+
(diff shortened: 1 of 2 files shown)


hartbit:
I dont agree with both:

1. The \\(expr) syntax is actually very elegant to me because it makes interpolation syntactically equivalent to any other escaping or special characters.

2. Swift does not promote the use of keyword aliases nor should it. Swift programmers are now used to using let which is both short and similar to the keyword in other functional languages.

In general, these two proposals seem to come from a wish to make Swift syntax more similar to languages the author is more fluent in. I don't think this is a good reason for accepting then.

Danappelxx:
I agree with @harbit - these changes are purely syntactical, which is are opinionated and don't have enough motivation to make them worthwhile (IMO).

soffes:
I agree with @hartbit as well.

Changing string interpolation from \\ to $ doesn't seem ideal to me. Most languages have some character to denote that it's a string interpolation. Using \\ is the most consistent since it's the same as other escape characters. This would also break tons of Swift code.

A keyword alias would only cause confusion in my opinion.

proposals/0008-add-const-as-alias-of-let.md
```diff
@@ -0,0 +1,30 @@
...
+
+Add const as an alias of let.
+
+## Detailed design
+
+Programmer may write const instead of let in source code. Mixing of const and let is forbidden though to avoid confusion.
+
+## Impact on existing code
+
+It is harmless new syntax sugar, thus no impacts.
```
soffes: I'd argue two keywords to do the same thing would be confusing to new Swift programmers.

DougGregor:
I love the enthusiasm here, but the intent is for changes to be discussed on the swift-evolution mailing list first. Pull requests with complete proposals are a later stage in the process, once an idea has been discussed and refined. Please see the Swift Evolution Process document (process.md) in this repository for more information.

soffes:
@pinxue love to see proposals from others in the community though! :heart:

pinxue:
Only if I can get mailing list page displayed...

PR #17, opened 2015-12-03
fix the API guideline link

ShonFrazier:
Was:
https://swift.org/api-guidelines

Should (probably?) be:
https://swift.org/blog/swift-3-api-design/


ShonFrazier:
nope. wrong link.

PR #18, opened 2015-12-03
fix the API guideline link

ShonFrazier:
Was:
https://swift.org/api-guidelines

Should (probably?) be:
https://swift.org/documentation/api-design-guidelines.html


ShonFrazier:
dup

PR #28, opened 2015-12-05
0007 for-loops: Format code

ken0nek:


vtourraine:
This code snippet is included as an example of real-life Swift code (from “a search of github's Swift gists”). The original source isn’t linked, because this is merely an example, and I assume that we don’t want/need to bring attention to the author.

So the fact that this code formatting is a bit awkward and non-standard is relevant to the point that this proposal is trying to make. I think it should be left as it is.

DougGregor:
I think the author's intent was to show code as it appears in the wild, not to show the best use of this feature. I'm going to close this; please contact the author of the proposal if you believe the proposal would be stronger with this change.

PR #32, opened 2015-12-06
Proposal to add StaticString.UnicodeScalarView

kballard:
It is occasionally useful to be able to produce a substring of a `StaticString`
that can be passed to APIs expecting a `StaticString`. For example, extracting
the filename from `__FILE__`. But there is no way to do this today, as
`StaticString` does not provide any means by which to create a new instance
beyond the trivial nullary `init()` initializer (which creates an empty
string).


kballard:
There's already an implementation of this as apple/swift#277

PR #33, opened 2015-12-06
Proposal to replace 'typealias' with 'associated' for associated type declarations

loiclec:
This has been discussed in the swift-evolution mailing list, and has received positive feedback so far :)

### DIFF
--- a/proposals/NNNN-replace-typealias-associated.md
+++ b/proposals/NNNN-replace-typealias-associated.md
@@ -1,0 +1,107 @@
+# Replace `typealias` keyword with `associated` for associated type declarations
+
+* Proposal: [SE-NNNN](https://github.com/apple/swift-evolution/blob/master/proposals/NNNN-replace-typealias-associated.md)
+* Author(s): [Loïc Lecrenier](https://github.com/loiclec)
+* Status: **Review**
+* Review manager: TBD
+
+## Introduction
+
+The `typealias` keyword is currently used to declare two kinds of types:
+
+1. Type Aliases (alternative name for an existing type)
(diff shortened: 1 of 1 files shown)


radex:
+1 for the proposal, keyword name, and proposed approach. I think it would be great if initial support could be landed in Swift 2.2 — doing deprecation and removal in two steps is such a great pattern. It eases the transition, especially as Swift matures, but still leaves us with room to change things without outright breaking of a lot of code.

lattner:
+1, to @DougGregor for consideration.

tjw:
I very much like the idea of splitting up these two concepts, but "associated" seems vague to me, since it doesn't specify the kind of relationship, just that there is one. A few possible spellings might be "type", "typearg", "needs_type", or anything that reads as there being a placeholder in the protocol that would be filled by a type-shaped thing.

wiruzx:
+1 for the `type` keyword. Scala uses `type` for [abstract types](http://docs.scala-lang.org/tutorials/tour/abstract-types.html) too.

anandabits:
This conflicts with the proposal to use a type keyword for type-level members unifying static and class.  If we are going to have a type keyword we'll only be able to use it in one way or the other, not both.  I don't have an opinion, just pointing this out.

tjw:
Unifying those (and getting rid of `static`, which I've never liked) would be good too. Using `type` for that would work, but something like `Self func foo(...)` or `func Self.foo(...)` also seem like they'd work (and make it clear that the type being referred to is the one defining the current scope.

DougGregor:
Please keep the discussion on the swift-evolution mailing list :)

PR #36, opened 2015-12-06
Add missing question mark for optional type

Shimmen:
The example implies that prefix is an optional String, although it's declared as a non-optional String. I assume it's a typo.


lattner:
Yep, thanks!

PR #43, opened 2015-12-08
Proposal: Remove Partial Application of Non-Final Super Methods

bitjammer:
This simplifies the implementation for dynamic super method dispatch and should have a small impact. I've sent out the rough idea to the mailing list and will link here.

### DIFF
--- a/proposals/0008-remove-partial-application-super.md
+++ b/proposals/0008-remove-partial-application-super.md
@@ -1,0 +1,131 @@
+# Remove Partial Application of Non-Final Super Methods (Swift 2.2)
+
+* Proposal: [SE-0008](https://github.com/apple/swift-evolution/blob/master/proposals/0008-remove-partial-application-super.md)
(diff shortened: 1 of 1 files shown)


lattner:
I commented on the list, but I'll reproduce it here for archival:

" Currying will be removed in Swift 3.0 so, rather than invest more engineering in those mechanisms, I propose that we disallow partial application of non-final methods through super, except where the self parameter is implicitly captured.”

This isn’t true.  The proposal you’re referring to removes the “func f(a : Int)(b : Int)” syntax, but methods will still be curried.  It will still be perfectly legal to do:

class Foo {
  func bar() {}
}

let fn = Foo.bar
fn()


So, I think that in a perfect world, the implementation would lazily generate curry thunks for super partial applications, and we would keep the current behavior.  However, if it is a significant implementation burden, I’d support doing a survey of existing code to see how widely used this is.  I suspect it is rare, and if you can prove this, we may be able to knowingly break compatibility in this case in  the service of getting the ABI stability.

PR #57, opened 2015-12-12
proposal to remove end of the line Semicolon

jonasman:


jonasman:
Preview: https://github.com/jonasman/swift-evolution/blob/master/proposals/remove-semicolons.md

masters3d:
Please provide the link to the mailing list for the discussion of this proposal. 
https://swift.org/contributing/#participating-in-the-swift-evolution-process
https://lists.swift.org/pipermail/swift-evolution/

jakob:
Personally I like semicolons. When writing @IBActions, I often have code that looks like the following:

```swift
@IBAction func doSomething(sender: NSObject?) {
    guard window.makeFirstResponder(nil) else { NSBeep(); return }
    guard let button = sender as? NSButton else { NSBeep(); return }
    
    // do something
}
```

Spreading that code out over multiple lines would make it harder to read.

DougGregor:
This needs to be discussed on swift-evolution first.

PR #58, opened 2015-12-13
Failable numeric conversion initializers

anandabits:


Lantua:
Most numeric types have failable initializer from String.
It seems to follow your motivation.

If ones try to read from loosely-typed data, they could as well read optionally from string, rather than read into Int and cast it later into Int16.

anandabits:
Loosely typed is not the same as "stringly" typed.  Plist data is a good example where you will end up with an NSNumber.

Furthermore when parsing structured text data application level code usually relies on parser libraries which will provide wide numeric types even if the application model expects and requires narrower numeric types.  The application does not necessarily have access to the raw string data.

Lantua:
Thanks for the information!

I tried to move my comment into mailing list, but I'm still very new to it. Do you happen to have a good source on how to use it?

lattner:
Just taking a look at this now, I'm surprised that these are marked "throws" instead of being a failable init.  Given that these can fail in exactly one obvious way, wouldn't it make more sense to make them be failable?  @dabrahams, do you have an opinion on this?

kballard:
I agree with @lattner. The only distinction I can think of in the failing case is indicating whether it overflowed or underflowed, but I don't think that info is very useful, and this API will be easier to use if it returns optionals rather than throwing (plus, this proposal doesn't even define what the thrown error will be, which means this proposal as written isn't even providing overflow/underflow info).

anandabits:
Thanks for the feedback folks.  After coming back to this I agree with you and have updated the proposal to use failable initializers.  

The primary purpose of a throwing initializer would be to capture the out-of-range value in the error message for debugging or logging.  This is easy to do on top of the failable initializers where necessary and would probably add an undesirable performance hit in common cases where it isn't necessary.

dabrahams:
I also agree with @lattner

lattner:
Thanks!

PR #59, opened 2015-12-15
Updated counter-argument section to more accurately represent the counter argument.

nickmshelley:


DougGregor:
The review for this proposal has concluded already.

PR #60, opened 2015-12-15
Conversion protocol conventions

anandabits:


lattner:
Has this been discussed on swift-evolution yet?  If not, please start a discussion there.

anandabits:
Yes, i posted a draft a day or two ago.  It received a few positive replies but not a lot of discussion.  Is that ok or should I try to restart the thread now that we're in the business week?

erica:
The proposal has been discussed, dissected, and reconstructed on swift-evolution, and was then updated in considerable depth.

PR #64, opened 2015-12-16
proposal for scoped access level

ilyathewhite:


anandabits:
Thanks Ilya.  I support the concept in general but would like to see it presented differently and in much more detail than it is here.

PR #69, opened 2015-12-19
Swift Language Version Build Configuration

bitjammer:


lattner:
David, did you consider syntax along the lines of:

#if swift(>2.2)

The syntax you propose implies an exact match.  Is there a range-match equivalent?

bitjammer:
We threw around the idea of indicating the implied ">=" part but not symbolically. I think you're right, the type of comparison should be explicit to prevent confusion. I'm not sure about ranges – would we use the same syntax as `..<` or `...`?

What do you think about opening up the operators to `<`, `>`, `<=`, `>=`, `==`, `!=`, and having something like this?

```swift
#if swift >= 2.2
  // new stuff, should work with newest Swift
#elseif swift >= 1.5
  // older stuff
#endif
```

or, if some prefer,

```swift
#if swift < 2.2
  // old stuff
#else
  // should work with newest Swift
#endif
```

bitjammer:
Actually, I think I like your `#if swift(>= 2.2)` better.

bitjammer:
Actually, after sketching out both implementations, I think `swift >= 2.2` looks pretty good, unless it's odd for it to not be a "function call". I brought this up on the e-mail thread.

lattner:
I'd recommend something like: #if swift(>=2.2), because this needs to be a boolean value that fits into the larger build configuration grammar.  Exposing the condition outside of the parens means that build configurations would be extended to the full integer/fp domain.

bitjammer:
Agreed - I was willing to expand the domain but it's probably not worth it for this. I amended the proposal to use the `#if swift(>=2.2)` syntax, I still keep coming back to that. I've amended the implementation to match - just updating tests now.

PR #73, opened 2015-12-22
Coroutine for Swift

SusanDoggie:
Coroutine can be used for generate a list of values without creating a array. And it's a useful mechanism for lazy evaluation.


lattner:
Hi @SusanDoggie, per our swift-evolution guidelines, please start a thread discussing this on the swift-evolution mailing list before sending a pull request.  This suggestion has already come up, and has been deferred due to its close inter-relation with concurrency support, which is out of scope for Swift 3.

PR #78, opened 2015-12-29
Clarify from northern hemisphere bias.

timburgess:


lattner:
I see what you mean, we should clarify this, but "Q4" is just confusing and unclear..

kud1ing:
Suggestions:
* "end of 2016"
* "September - November 2016"

lattner:
I changed it to "late 2016" in b7dee14, thanks.

timburgess:
:+1:

PR #84, opened 2016-01-05
Add alternative considered re: Kotlin Scala syntax

anandabits:


lattner:
I think it is great to include this in the "alternatives considered" section, but I'd suggest moving the "More importantly, this approach was not chosen" section much earlier.  It is really the crux of the matter.

anandabits:
Good idea @lattner, done.

lattner:
Thanks!

PR #85, opened 2016-01-05
Proposal for public(objc) modifier

kballard:


DougGregor:
I'm not seeing much in the way of support for this on the list; the only replies were mild "against"s. Did I miss something?

kballard:
Well, the only person who responded at all was Félix Cloutier, and it seems like he didn't even understand what I was proposing (his second email indicates that he was misinterpreting the proposal). So really it seems like there wasn't much response at all.

I have to wonder how many people besides me are actually running into this issue at all. It's something that annoys me almost every time I implement a library-level API in Swift that needs to be exposed to Obj-C. But maybe other people either just aren't doing this a lot, or are just providing an Obj-C–compatible API even at the Swift level instead of having a better Swift-only API. Or maybe they just don't recognize the issues caused by exposing the Obj-C version of the API to Swift (it's not just visual clutter in autocomplete; I had a coworker actually write Swift code that called my `__`-prefixed Obj-C wrapper instead of calling the proper Swift version because it was in the autocomplete list and he didn't realize what was going on).

DougGregor:
Yeah, that's what it looked like to me. We don't want to accept proposals into the system until there appears to be some critical mass supporting the proposal.

kballard:
This is rather annoying, because right this moment I'm implementing an Obj-C wrapper for a Swift API and I'm going to end up with literally dozens of methods that are inappropriately exposed to Swift.

DougGregor:
It doesn't make sense to bring up a proposal that doesn't have some critical mass of support form the community. Maybe it got missed, and reposting again will draw enough attention.

PR #93, opened 2016-01-10
Proposal: Add replace(_:with:) function to the stdlib

kballard:


adrfer:
@kballard, what about a version of `replace` that would return the new value instead of mutating it? Well, as of now, according to the [API Design Guidelines](https://swift.org/documentation/api-design-guidelines.html), the suggested name would be `replaced`.

Do you think it's worth updating your proposal to include a non-mutating counterpart for `replace`?

Also, I came across this [proposal](https://github.com/apple/swift/blob/master/docs/proposals/Inplace.rst)... but not sure how current it ir though.

kballard:
I have no idea what you mean by a version of `replace` that returns the new value instead of mutating it. The new value is the second parameter to the `replace(_:with:)` function. If you want to just return the new value instead of mutating it, then the function is literally doing nothing at all, e.g. `replace(&a, with: b)` would just become `b` without a function call. The whole point of the function is the mutation; without it, there's no function at all.

adrfer:
@kballard I apologize for my confusion. Guess I read too many times your proposals [Add scan, takeWhile, dropWhile, and iterate](https://github.com/apple/swift-evolution/pull/95) and [Add function SequenceType.find(_:)](https://github.com/apple/swift-evolution/pull/94) and ended up treating the `replace` function in the same context, unrelated to properties.

DougGregor:
There hasn't been enough discussion on this to form adequate consensus to move forward with a proposal.

Coeur:
@kballard When reading a name like _replace_, I would expect it to return the new value, not the old one, which makes the feature counter-intuitive. I would advice to name the feature _exchange_.

Coeur:
Oh, and what about handling two `inout` values at the same time?

kballard:
We already *have* a function called `swap`, and it takes two `inout` values (and returns nothing). It's rarely used, because most of the time where you'd want to use it, one of your values is actually a temporary instead of a mutable variable (hence the need for `replace(_:with:)`). Which is why in my Swift code I use `replace(_:with:)` pretty often, but I think I've only ever used `swap(_:_:)` once. Similarly in Rust code `replace()` is used pretty frequently and `swap()` almost never.

PR #95, opened 2016-01-10
Add scan, takeWhile, dropWhile, and iterate to the stdlib

kballard:


proposals/0000-scan-takewhile-dropwhile.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Introduction
+
+Add 3 new `SequenceType` functions `scan(_:combine:)`, `takeWhile(_:)`, and
+`dropWhile(_:)`, with overrides as appropriate on `CollectionType`,
+`LazySequenceType`, and `LazyCollectionType`, as well as a global function
+`iterate(_:apply:)`.
+
+Swift-evolution thread:
+[Proposal: Add scan, takeWhile, dropWhile, and iterate to the stdlib](https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20151228/004690.html)
```
dabrahams: I’d like to encourage people to use gmane links (in this case, http://news.gmane.org/find-root.php?message_id=1451347170.2635180.478054954.2AD975B7%40webmail.messagingengine.com) because they are so much more useful.

proposals/0000-scan-takewhile-dropwhile.md
```diff
@@ -0,0 +1,138 @@
...
+* Status: **Awaiting review**
+* Review manager: TBD
+
+## Introduction
+
+Add 3 new `SequenceType` functions `scan(_:combine:)`, `takeWhile(_:)`, and
+`dropWhile(_:)`, with overrides as appropriate on `CollectionType`,
+`LazySequenceType`, and `LazyCollectionType`, as well as a global function
+`iterate(_:apply:)`.
+
```
dabrahams: One of my concerns here is that while map, filter, and reduce are mostly
“lingua franca” among programmers everywhere, these others are much less
commonly-known and some of them aren't very descriptive of what they do.
There are lots more like these in the FP lexicon.  It was easy to draw
the line at map/filter/reduce (which also don't follow the usual grammar
for API guidelines), but if we take the ones in this proposal, where do
we stop?

proposals/0000-scan-takewhile-dropwhile.md
```diff
@@ -0,0 +1,138 @@
...
+Add the following extension to `SequenceType`:
+
+```swift
+extension SequenceType {
+  /// Returns an array containing the results of
+  ///
+  ///     p.reduce(initial, combine: combine)
+  ///
+  /// for each prefix `p` of `self`, in order from shortest to longest.
```
dabrahams: It would be good to say explicitly whether the empty prefix is included, and maybe even to be explicit that the longest prefix is `self` itself.

proposals/0000-scan-takewhile-dropwhile.md
```diff
@@ -0,0 +1,138 @@
...
+  /// For example:
+  ///
+  ///     (1..<6).scan(0, combine: +) // [0, 1, 3, 6, 10, 15]
```
dabrahams:
```swift
(1..<6).reducedPrefixes(0, combine: +)
```

might be a more descriptive name, if we decide `scan` is too esoteric.  I don’t love it, though.

proposals/0000-scan-takewhile-dropwhile.md
```diff
@@ -0,0 +1,138 @@
...
+}
+```
+
+Modify the declaration of `SequenceType` with two new members:
+
+```swift
+protocol SequenceType {
+  // ...
+  /// Returns a subsequence by skipping elements while `dropElement` returns
+  /// `true` and returning the remainder.
```
dabrahams:
```swift
/// Returns the suffix of `self` beginning with the first element not satisfying `dropElement`
```

seems like a better description (saying it returns a subsequence adds nothing that isn’t in the signature).  But then  why wouldn’t we just invert the predicate and call this as

```swift
x.suffixFrom(firstElementSatisfying: somePredicate)
```

?

proposals/0000-scan-takewhile-dropwhile.md
```diff
@@ -0,0 +1,138 @@
...
+  func dropWhile(@noescape dropElement: (Self.Generator.Element) throws -> Bool) rethrows -> Self.SubSequence
+  /// Returns a subsequence containing the elements until `takeElement` returns
+  /// `false` and skipping the remainder.
+  func takeWhile(@noescape takeElement: (Self.Generator.Element) throws -> Bool) rethrows -> Self.SubSequence
```
dabrahams: similar remarks here, using `prefixTo`.

proposals/0000-scan-takewhile-dropwhile.md
```diff
@@ -0,0 +1,138 @@
...
+`dropWhile(_:)` will perform the filtering when `startIndex` is accessed.
+
+Add a global function:
+
+```swift
+/// Returns an infinite sequence of lazy applications of `apply` to the
+/// previous value. For example:
+///
+///     iterate(1, apply: { $0 * 2 }) // yields: 1, 2, 4, 8, 16, 32, 64, ...
+func iterate<T>(initial: T, apply: T -> T) -> IterateSequence<T>
```
dabrahams: Would it be better to express this by creating a `repeatForever` and using `scan`?  That seems like it would yield more expressive code.

dabrahams:
Kevin, this PR is clearly ready to be accepted, but I’m going to bring these issues up in review if it’s accepted in this form so I wanted to give you a chance to make any changes you may wish to, in response to my comments, before I accept.  Just let me know what you want me to do.

proposals/0000-scan-takewhile-dropwhile.md
kballard: I haven't heard any suggestions yet for alternate names that I thought were any better.

`dropWhile(_:)` is patterned after `dropFirst(_:)`. Personally I'd actually be in favor of renaming them both to `skipWhile(_:)` and `skipFirst(_:)` to avoid the potential confusion around whether they're mutating, but as long as we have `dropFirst()` then `dropWhile()` seems appropriate. Or perhaps that should be `drop(while:)` now? I'm not sure.

For `takeWhile(_:)`, the closest existing method is `prefix(_:)`, but `prefixWhile(_:)` (or `prefix(while:)`) doesn't make sense as "prefix" is a noun, not a verb (well, it is a verb too, but that means something different). I suppose we could go with `prefixHaving(_:)` (or should that now be `prefix(having:)`?) for that one.

And for `scan(_:combine:)`, I really don't know what else to call it. I'm open to suggestions, but since this one doesn't have an obvious plain english name, I'm inclined to say sticking with FP terminology makes the most sense.

Same thing with `iterate(_:apply:)`, I'm not sure what else to call it.

proposals/0000-scan-takewhile-dropwhile.md
kballard: Good idea.

proposals/0000-scan-takewhile-dropwhile.md
kballard: Good idea.

proposals/0000-scan-takewhile-dropwhile.md
kballard: Yeah, that's a bit awkward. Really the idea of this function is it's `reduce(_:combine:)` but it includes all of the intermediate results. But of course `reduceIncludingIntermediateResults(_:combine:)` is much too unwieldy.

proposals/0000-scan-takewhile-dropwhile.md
kballard: I don't like the emphasis on suffixing here. When you think "suffix" you typically think about working backwards from the end, but this function is really about skipping elements at the start. And in fact the name `suffixFrom(firstElementSatisfying:)` could be construed as starting with the first element from the end that satisfies that predicate, instead of the first element from the start.

As for the bit about returns a subsequence, you may as well say that `map(_:)` shouldn't say "Return an `Array` ..." since that's part of the type signature.

proposals/0000-scan-takewhile-dropwhile.md
kballard: I don't like inverting the predicate here. The focus here is on retaining the initial elements according to some predicate, and it feels very backwards to flip the predicate around so it describes what you aren't retaining. In my experience, usually the predicate with this function is used to describe a characteristic of the elements you want to keep, not a characteristic of the elements you want to skip.

I could be convinced to go with `prefixHaving(_:)` (or is it `prefix(having:)` now?), though I'm still partial to `takeWhile(_:)` because that's the naming that all of the precedent I'm familiar with uses.

proposals/0000-scan-takewhile-dropwhile.md
kballard: `scan(_:combine:)` returns an array, you can't use it on an infinite sequence. The lazy version returns a lazy sequence, but the stdlib philosophy (last I checked) was to not perform lazy sequence operations without the user specifying `.lazy` somewhere, so this theoretical `repeatForever()` would have to be like `zip()` in that it may operate lazily itself but it still exposes eager sequence operations. So this turns the expression into `repeatForever(f).lazy.scan(0, apply: { $0 * 2 })`, which is a bit of a mouthful and isn't really the most obvious way to accomplish this. And I'm sure if I were to propose a `repeatForever(_:)` there'd be pushback from people saying that we really should just have a `.cycle()` method (which I suggested on the ML before and had a lot of debate over infinite sequences) instead, which would turn it onto the even more unwieldy `CollectionOfOne(f).cycle().lazy.scan(0, apply: { $0 * 2 })`.

All that said, I do want to introduce a `.cycle()` method, but `iterate(_:apply:)` is a specific pattern that is relatively common to encounter when doing FP-style manipulation of sequences (and works very nicely with `.takeWhile(_:)`) that it seems worth having, if for no other reason than the fact that it would save everyone who wants this type of functionality from having to redefine it themselves in terms of `CollectionOfOne(f).cycle().lazy.scan(_:apply:)`.

kballard:
@dabrahams I've updated the PR, including a lengthy Alternatives section that covers name suggestions.

gribozavr:
> the stdlib philosophy (last I checked) was to not perform lazy sequence operations without the user specifying .lazy somewhere

That's not quite how it is.  The rule we have been following is to perform the operation lazily, unless we would need to capture a user-defined closure in order to do so, or unless the lazy operation would be much more expensive.  Thus, `zip()`, `reversed()`, `enumerated()` are lazy, while `map()`, `filter()` and `sorted()` are eager.

kballard:
@gribozavr Ok you're right, I spoke inaccurately. But `scan(_:combine:)` would have to capture a closure, so it would need the `.lazy` in that expression.

gribozavr:
Yeah, I don't disagree, just clarifying a rule that we apparently never wrote down.

gribozavr:
Actually, we did: https://github.com/apple/swift/blob/master/docs/StdlibRationales.rst#lazy-functions-that-operate-on-sequences-and-collections

proposals/0000-scan-takewhile-dropwhile.md
```diff
@@ -0,0 +1,222 @@
...
+  func takeWhile(takeElement: (Self.Generator.Element) -> Bool) -> LazyTakeWhileSequence<Self.Elements>
+}
+```
+
+The types `LazyScanSequence`, `LazyDropWhileSequence`, and
+`LazyTakeWhileSequence` all conform to `LazySequenceType`.
+
+Extend `LazyCollectionType` with collection variants for `dropWhile(_:)` and
+`takeWhile(_:)` (but not `scan(_:combine:)` because there's no way to recover
+the value from an index without re-scanning from the start):
```
gribozavr: You can still conform to forward collection, though.  You can keep the current value in the index itself.

kballard: That also requires that the index have a reference to the underlying collection as well, so it can increment itself, which requires that the collection itself be held by-reference by the collection wrapper (it could be stored inline and just copied into each index, but arbitrary collections might be quite large so that would be an expensive index).

I seem to recall seeing at one point a proposal to change how collection indexes work so the get incremented/decremented via methods on their associated collection. I don't see it in the swift-evolution repo so I assume it was somewhere on the ML. In any case, given a change like that, it would be a lot easier to store the value in the index because it wouldn't have to keep track of the associated underlying collection.

gribozavr: There's a thread on swift-evolution alread, and I am planning to propose officially it soon: https://github.com/gribozavr/swift-evolution/blob/new-collections/proposals/NNNN-collections-move-indices.md

gribozavr: I would recommend to proactively make this change from sequence to collection.  The current Dictionary implementation has to keep a reference to the dictionary, this case is not that different.

kballard: Dictionary is already implemented with a COW reference internally though. It seems unfortunate to introduce an internal reference type inside of the `scan` implementation when it's otherwise not required. But I suppose it's probably better to do that now to get the API in place, and the internal reference type can then be removed at such time as your proposal is accepted and implemented.

kballard:
@dabrahams You merged this without assigning a number to it. I'm going to submit a PR with @gribozavr's changes, so I'll also give it number 0045 (since that's the next available one).

proposals/0000-scan-takewhile-dropwhile.md
kballard: Hrm, looking at `LazyFilterIndex`, that actually stores the underlying collection inline in the index. That's a bit surprising. I guess that provides precedent for doing the same thing with `LazyScanCollection.Index` though.

kballard:
PR submitted as #195

dabrahams:
on Mon Mar 07 2016, Kevin Ballard <notifications-AT-github.com> wrote:

> @dabrahams You merged this without assigning a number to it. I'm going
> to submit a PR with @gribozavr's changes, so I'll also give it number
> 0045 (since that's the next available one).

I just took care of that part.

-- 
-Dave

PR #96, opened 2016-01-10
Add proposal for abstract classes

dscreve:


Danappelxx:
Is the only difference between abstract classes and protocols (with extensions) the ability to have default values for properties? If so, I don't think it makes sense to create a new type+keyword just for that.

muhasturk:
In real life, protocols with extensions seems enough.

NachoSoto:
A protocol can't have storage, that's the advantage that an abstract base class would add.

DougGregor:
Please continue this discussion on swift-evolution.

DougGregor:
... where we discuss what it would take to implement stored properties for protocols.

AlexanderKaraberov:
As for me it can bring additional accidental complexity to the language core. We can wonderfully
 model abstract classes with protocol extensions which have default method implementations +  computed properties and computed type properties + new initializers.
@NachoSoto I think Swift emphasize protocol-oriented vs. OOP approach. And abstract base classes which complect state (storage) with behaviour instead of protocol extensions is not very good...

swiftTailor:
For me abstract classes, methods and properties make sense. Take the class inheritance example of the swift programming language documentation with the
- Vehicle
- Bicycle
- Car
- AutomaticCar

The class Vehicle has the properties currentSpeed and description which can be used by the inherited classes but it makes no sense to instantiate the class Vehicle.
To prevent that the class should be marked as abstract.
The method makeNoise would also be a cantidate to be marked as abstract to force that inherited classes have to implement that method.
Protocols do only define the interface without any implementation. Therefore they are not sufficient to build partly implemented classes which can only be used as base classes and should never be instantiated.
Because swift has often its own style - which I really like - and uses other keywords than used in other programming languages "abstract" could also be replaced by "incomplete", "partlyimplemented", "pure base" or similar.

PR #102, opened 2016-01-15
Clean reset, with updated README file only. I'll do another process p…

erica:
Doug:

This *should* be a clean pull request with just the README.md changes.


lattner:
Awesome, thanks Erica!

PR #103, opened 2016-01-15
Tail Call Optimization attribute and modifier

griotspeak:
Add a `@tailrec` attribute and `tail` modifier allowing the compiler to emit errors when tail call optimization  will not be used.


kostiakoval:
Is there are See swift-evolution thread for where this features was discussed? Can you include it ? 

"Swift-evolution thread: link to the discussion thread for that proposal" in [template.md](https://github.com/apple/swift-evolution/blob/master/0000-template.md)

griotspeak:
Thanks for pointing that out.

iUtvikler:
Could you add newlines above and below the *Alternatives considered* section?

Also i'm not sure I understand the **Proposed Solution**.

It was my understanding that the *tail* modifier would be a modifier on the return statement where you do the tail call.

```swift
func sumRecursively(numbers: [Int], _ total: Int = 0) -> Int {
    guard let head = numbers.first else {
        return total
    }
    
    let tail = Array(numbers.dropFirst())
    
    tail return sumRecursively(tail, head + total)
}
```
[Example from natashatherobot.com](https://www.natashatherobot.com/functional-swift-tail-recursion/)

And i thought the *@tailrec* function attribute was only needed on a public API that wants to be externally tail-callable.

Can`t think of an example of that..

lattner:
@griotspeak It was great to meet you TJ!  I'm sorry that this is not in scope for Swift 3, but there will certainly be another release...

DougGregor:

PR #106, opened 2016-01-18
Correct name of CollectionDefaultIterator

natecook1000:
It looks like it's `CollectionDefaultIterator` in the source, not `DefaultCollectionIterator`.


gribozavr:
We actually settled on `DefaultCollectionIterator`, so it would be best to correct the source :)

natecook1000:
Oh good, I like that better! :+1:

PR #110, opened 2016-01-18
Import Objective-C Constants as Swift Types

SlaunchaMan:
A way to add some niceties to how Objective-C constants are imported into Swift.

* [Original E-Mail](https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20160111/006893.html)
* [Replies](https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20160118/006904.html)


DougGregor:
When you provide another update, please squash.

PR #119, opened 2016-01-24
Variables in case labels with multiple patterns

therealbnut:
Proposal to "Declare variables in 'case' labels with multiple patterns".

Thanks for your time

[0023-declare-variables-in-case-labels-with-multiple-patterns.md](https://github.com/therealbnut/swift-evolution/blob/therealbnut-case-match/proposals/0023-declare-variables-in-case-labels-with-multiple-patterns.md)


therealbnut:
yay :) thanks @lattner

PR #120, opened 2016-01-24
Behavior of comments next to operators

jder:
Specifies that comments next to operators should be ignored when determining whether there's whitespace on the left or right of the operator.

While swift-evolution discussions (and the preceding swift-dev discussion) had minimal participation, there seemed to be a wide agreement that something should be done about this.


lattner:
Accepting this and adding to the to-be-scheduled pile, thanks!

PR #123, opened 2016-01-28
Adding proposal to modernize debug identifiers

erica:
Just leaving this here for now.

### DIFF
--- a/proposals/00xx-modernizing-debug-identifiers.md
+++ b/proposals/00xx-modernizing-debug-identifiers.md
@@ -1,0 +1,97 @@
+# Modernizing Swift's Debugging Identifiers
+
+* Proposal: TBD
+* Author(s): [Erica Sadun](http://github.com/erica)
+* Status: TBD
+* Review manager: TBD
+
+## Introduction
+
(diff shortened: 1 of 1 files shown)


lattner:
Thanks Erica,

Doug and I discussed this.  He is now on-board with adding #line, #column, etc as you have described.  I would suggest moving the #sourcelocation field to a "possible future extensions" section to make it clear that we can always add that as a follow-on if/when we decide to tackle a standardized SourceLocation type.

Also, to address the ambiguity with #line, we should make the rule be tied to whether #line is the first token on the line.  This is the policy for other whitespace sensitivity we have in the language.

I think that #symbol is great, but it will also be worth discussing whether we keep #function as it currently stands and add #symbol, or whether we do as you suggest and remove #function.

We can bikeshed on the exact naming of the symbols but #dsohandle with a lower case H seems more natural (analogous to typealias and associated type), but completely other spellings are also welcome :-)

erica:
I will adjust accordingly and perform my standard "delete entire repo, refork, and add" because this is the only way I know how to provide a clean pull request.

-- E

lattner:
Works for me, thanks again!

PR #126, opened 2016-01-29
Proposal to modernize debug identifiers

erica:
Updated, clean pull request for modernizing debug identifiers from "screaming snake case" to lower case.


erica:
Can you please kill 123? Thanks!

-- E

PR #127, opened 2016-01-29
Adjusting  declarations for type decoration proposal

erica:
Added proposal to amend `inout` declarations so they decorate the type and not the parameter label.


lattner:
subsumed by 128

PR #128, opened 2016-01-29
Adjusting declarations for type decoration proposal

erica:
Trying again. Because I stink at git.


lattner:
Closing at erica's request

PR #133, opened 2016-01-31
Runtime-safe array subscripting

salutis:


avaidyam:
I noticed you switched the `Setting values` and `Getting values` comments around.

Also, out of curiosity, why not simply do something like this:

```swift
var array = [0, 1, 2]

// Getting values
array.safeAccess[0] // same as array[0]
array.safeAccess[3] // out of bounds, evaluates to nil

// Setting values
array.safeAccess[0] = 42 // same as array[0] = 42
array.safeAccess[3] = 42 // out of bounds, no operation
```

Where `safeAccess` could be implemented like so:
```swift
public struct SafeAccessor<T: CollectionType> {
    let value: T

    public subscript(index: Int) -> Int? {
        get { 
            return try? self.value[index]
        }
        set (newValue) {
            try? self.value[index] = newValue
        }
    }
}

public extension CollectionType {
    public func safeAccess() -> SafeAccessor<Self> {
        return SafeAccessor<Self>(self)
    }
}
```

It's all pseudocode, but essentially, it can be implemented using a reflected accessor, and doesn't need to be language level. I'll try to flesh this out later.

gribozavr:
The `try? self.value[index]` syntax is not implementable for all collections without an undue burden that will affect performance in all cases (including `collection[index]`).  It is not always possible to detect an invalid index reliably.

avaidyam:
@gribozavr Here's what you want:
```swift
public extension CollectionType {
	public subscript (safe index: Index) -> Self._Element? {
		get {
			return self.indices.contains(index) ? self[index] : nil
		}
	}
}

public extension MutableCollectionType {
	public subscript (safe index: Index) -> Self._Element? {
		get {
			return self.indices.contains(index) ? self[index] : nil
		}
		set(newValue) {
			if let val = newValue where self.indices.contains(index) {
				self[index] = val
			}
		}
	}
}
```

I've tested that it works exactly according to your spec.

```swift
var array = [0, 1, 2]
array[safe: 0] // same as array[0]
array[safe: 3] // out of bounds, evaluates to nil
array[safe: 0] = 42 // same as array[0] = 42
array[safe: 3] = 42 // out of bounds, no operation
```

gribozavr:
That would cause the safe subscript to be O(n), which is unacceptable.

avaidyam:
`self.indices.contains()` is the only new call here; I don't see where the documentation states its overhead.

avaidyam:
In fact, it looks like the implementation is right here:

```swift 
/// O(1) implementation of `contains()` for ranges of comparable elements.
extension Range where Element : Comparable {
  @warn_unused_result
  public func _customContainsEquatableElement(element: Element) -> Bool? {
    return element >= self.startIndex && element < self.endIndex
  }

  // FIXME: copied from SequenceAlgorithms as a workaround for
  // https://bugs.swift.org/browse/SR-435
  @warn_unused_result
  public func contains(element: Element) -> Bool {
    if let result = _customContainsEquatableElement(element) {
      return result
    }

    for e in self {
      if e == element {
        return true
      }
    }
    return false
  }
}
```

I'm taking this to mean `_customContainsEquatableElement` is O(1). It doesn't seem like the safe subscript is O(n) in this case.

gribozavr:
Not all indices are comparable, and, that overload of contains() does not work from a generic context.

avaidyam:
@gribozavr Are there cases you see O(n) with this safe subscript implementation? I think for the majority of cases, it would be O(1). Of course, I'm thinking of `Set`s, `Array`s, and `Dictionary`s.

gribozavr:
Even Array would be O(n) since, as I said, that overload of contains() can't be called from a generic context.

Also, for invalid indices, the comparison operation is allowed to return whatever (or even trap!), and the actual subscript operation might trap as well.

avaidyam:
@gribozavr I see what you mean. I'll think about this a little more.

dabrahams:
I am inclined to reject this PR because no use-cases are shown or discussed, and it doesn’t say what problem it’s solving.

DougGregor:
This proposal has insufficient detail: it does not provide the declarations of the subscript operator to be added, provide motivating use cases, or explore the alternative at all. I suggest revising the proposal and posting it to swift-evolution again as a "[Proposal draft]" to see if it draws more attention with the additional detail.

PR #137, opened 2016-02-02
Revise SE-0003 for reconsideration of 'var' patterns

bitjammer:


lattner:
Thanks for updating this @bitjammer !

PR #140, opened 2016-02-04
Implementing `Comparable` On `NSOperatingSystemVersion`

rsmoz:


rsmoz:
@parkera Any plans to take a look at this? If not, I'm honestly cool with that; I don't want to clog up the evolution process with someone that's pretty insignificant. But I also think it's something that could be dealt with relatively quickly, and maybe open the door for further proposals that seek to add some Swifty extensions to Foundation types. Just wondering if I should close the PR.

parkera:
As far as the actual content of the proposal: I think we should do the `Comparable` adoption, but probably not the new init method.

The main issue is one of process; we have a long-standing API review process for Foundation which requires a significant internal review before we change anything. I'm still trying to figure out how to integrate great suggestions like this one from the open source community into that process. 

Basically, we need to schedule someone to move the proposal through both the internal and external process. It takes a significant amount of time. However, like everyone, we have a lot of things on our plate, so it's been challenging to keep up with everything.

So I don't want to close this PR, because I think we should do it. We just need some additional time to figure out the best way to go about it.

parkera:
And for any Apple people watching, I've filed a radar so we can keep track of it in our internal system as well. 25378378.

rsmoz:
Ok, that makes sense. Should I modify the proposal to remove the init?

parkera:
Sure, I think that'd be good.

rsmoz:
@parkera Happy to see [this](https://github.com/apple/swift-evolution/blob/master/proposals/0069-swift-mutability-for-foundation.md)! In light of it, would it make sense to close my proposal? It seems like yours could potentially include a `Comparable` `NSOperatingSystemVersion`.

parkera:
Actually I didn't include changes for NSOperatingSystemVersion in that, so I'd like to keep track of this one somehow. I'll come back to it.

lattner:
@parkera I think we should close this PR, since we do not handle framework extensions and changes with the swift-evolution process.  Ok with you?

parkera:
Ok.

parkera:
@rsmoz We'll continue to track this ourselves as an ER.

lattner:
ER == Enhancement Request

lattner:
Thanks Tony!

PR #145, opened 2016-02-09
New proposal

jeschot:
Proposal for a default func== in Swift


lattner:
Please continue the discussion on swift-evolution.  At this point there are only two responses, which is not a very detailed discussion.

PR #162, opened 2016-02-22
Modernizing Attribute Case and Attribute Argument Naming

erica:
Two isolated instances of snake case remain in the Swift attribute grammar. This proposal updates those elements to bring them into compliance with modern Swift language standards and applies a new
consistent pattern to existing attributes.


lattner:
I think this is subsumed by PR183.

PR #163, opened 2016-02-22
Unifying `split`/`join`

moiseev:
- `split` obtains `separator` as first argument label
- `join` gets renamed to `joined`


moiseev:
@gribozavr please take a look.

PR #176, opened 2016-02-27
Proposal: deprecate the default keyword.

narrativium:
Created NNNN-deprecate-the-default-keyword.md file, a Swift proposal for deprecating the 'default' keyword.


lattner:
This has been discussed before: https://github.com/apple/swift-evolution/blob/master/commonly_proposed.md

lattner:
Also, Doug's proposal isn't specific to the default keyword.  Many other things (e.g. import) are hit as well.

PR #179, opened 2016-03-01
Conventionalizing `stride` semantics

erica:
Swift offers two stride functions, stride(to:, by:) and stride(through:, by:). This proposal introduces a third style and renames the existing to and through styles.


lattner:
Subsumed by PR184.

PR #181, opened 2016-03-01
Evolving Swift Process Document tweaks

erica:
Bringing this back into a new pull request because Joe Groff taught me how to branch! This is the same document discussed several months ago.


lattner:
PR186 subsumes this.

PR #185, opened 2016-03-01
Decoupling Floating Point Strides from Generic Implementations

erica:
Swift strides create progressions along “notionally continuous one-dimensional values” using
a series of offset values. This proposal replaces the Swift’s generic stride implementation
with seperate algorithms for integer strides (the current implementation) and floating point strides.


lattner:
@dabrahams ?

PR #188, opened 2016-03-04
Mixins

Anton3:
All current suggestions have been addressed, and no more have been posted in two days.


lattner:
I haven't actually looked at the proposal, but please squash all history into a single commit.

PR #192, opened 2016-03-05
Mixins

Anton3:


Pearapps:
I _really_ like this.

proposals/NNNN-mixins.md
```diff
@@ -0,0 +1,499 @@
...
+var y = B.x
+func g(arg: Type) = B.f
+```
+
+Example:
+
+```swift
+mixin A { var x: Int = 0 }
+mixin B { var x: Int = 1 }
+struct C : A, B { var y = B.x }
```
Danappelxx: I think this is a typo? This should still be a compilation error (it's not clear whether it's using A.x or B.x), unless the proposal is `var x = B.x` (in which case this a typo that needs to be updated).

lattner:
This is a very interesting idea, but clearly out of scope for Swift 3 at this point.  @DougGregor how would you like to handle this?

DougGregor:
This is an interesting proposal that is out of scope for Swift 3. Please bring it up again when we start scoping the next major Swift release.

Evertt:
Would now be a good time?

Anton3:
Nope, it's been postponed to Swift 4 Phase 2 (another 6 months).

>

Evertt:
Ah okay, thanks for letting me know. :-)

Bartzy:
@Anton3 Is there a summary somewhere of these details for features?

Anton3:
I thought this topic was mentioned specifically. It wasn't, but anyway it's
an additive proposal, and they are outside of Phase 1 scope. Read more here:
https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20160725/025676.html
https://github.com/apple/swift-evolution

proposals/NNNN-mixins.md
```diff
@@ -0,0 +1,501 @@
...
+    B.super.init()
+  }
+}
+```
+
+Note that `super` syntax can also be used for calling supertype members, as in classes.
+
+### Mixin inheritance
+
+If a single mixin is mixed-in using different paths, then only one copy of it will be kept in the final struct/class. Example:
```
Evertt: I believe this is a bad idea. I believe the developer should **always** have to explicitly define whether or not the base mixin should be copied twice or if only one copy should be used.

Here's an example where you might only want 1 copy of the base mixin:

```swift
mixin Mover {
    var xPosition: Int
}

mixin Jumper: Mover { ... }
mixin Walker: Mover { ... }

class Human: Walker, Jumper {
    ...
}
```

It is clear that even though a human can be both a walker and a jumper, a human can only be in one place at a time, so it should only inherit one copy of the `xPosition: Int` of Mover. This means that both `Walker` and `Jumper` will work with the same variable `xPosition`.

Here's a case where you might want 2 copies of the base mixin (shamelessly stolen from the mailing list):

```swift
mixin UniversityPerson {
    var computerAccount: Account
}
mixin Teacher: UniversityPerson { ... }
mixin Student: UniversityPerson { ... }

class TeachingAssistant: Teacher, Student { 
    ... 
}
```

Here a TeachingAssistant might actually have two different computerAccounts. One for his studying activities and one for his teaching activities. So in this case the developer would need to rename one or both of the versions of the inherited variable.

There's no reliable way to make assumptions about when things should be copied once or multiple times, which is why I propose to force the developer to explicitly define it. Here's a possible syntax we could use for that:

```swift
class TeachingAssistant: Teacher[computerAccount as teacherAccount],
                         Student[computerAccount as studentAccount] { 
    // We explicitly conveyed that we need
    // two different copies of computerAccount
}

class Human: Walker[xPosition],
             Jumper[xPosition] {
    // We explicitly conveyed that we want walker and jumper
    // to work with the same copy of the variable xPosition
    // [xPosition] is just syntactic sugar for [xPosition as xPosition]
}
```

And we could easily mix and match that like so:

```swift
class C: M1[someVar],
         M2[someVar],
         M3[someVar as someVar2],
         M4[someVar as someVar2],
         M5[someVar as someVar3] {
    // M1 and M2 use the same copy of someVar
    // M3 and M4 also use the same copy,
    // but different from M1 and M2
    // and M5 uses an entirely different copy again
}
```

Anton3: Please, read "two kinds of inheritance in a diamond pattern" section.

Evertt: I did and it says:

> By default, in a diamond pattern `A -> (B, C) -> D, D` contains a single copy of `A` for usage by `B`, `C` and `D`.

I'm proposing to not have a default strategy and to always make the developer be explicit about their intent. Also I believe the syntax you proposed to resolve the conflicts in a diamond pattern does not handle the case where you have more than 2 mixins inheriting from the same "base" mixin. The syntax that I propose does handle that case.

HT154: Forgive me if I'm off track, but wouldn't it be possible to address this by overriding the property in one child mixin and using a separate property for storage? This way, only the shared strategy is necessary and there's no need for additional syntax. Here it is applied to the teaching assistant example:

```swift
mixin UniversityPerson {
    var computerAccount: Account
}
mixin Teacher: UniversityPerson {
    final var teacherAccount: Account
        // the proposal does not mention access control, but this ought to be private

    override var computerAccount: Account {
        get { return teacherAccount }
        set { teacherAccount = newValue }
    }
    ...
}
mixin Student: UniversityPerson { ... }

class TeachingAssistant: Teacher, Student { 
    // Teacher.computerAccount and Student.computerAccount are distinct
    ... 
}
```

Evertt: @HT154 Okay, so what do I get when I write the following?

```swift
let john = TeachingAssistant(...)
print(john.computerAccount.accountName)
```

Which computer account will be printed?

HT154: I see. Looks like I missed the forest for the trees.

@Evertt Your proposed syntax seems like it could get out of hand when dealing with mixins with many properties. I think the proposal's existing syntax is nearly adequate as-is to express what yours does. I also propose extending the proposal's syntax to allow multiple mixins that share the same base to share the same renamed binding.

```swift
mixin Base { var x: Int }
mixin A: Base {}
mixin B: Base {}
mixin C: Base {}
mixin NotBase { var x: Int }
mixin D: NotBase {}
mixin E: NotBase { var y = NotBase.x } // Is it possible to perform renaming even if there are no conflicts?

struct Q: A, B, C, D {
    var xAB = [A.x, B.x]
    var xAB { A.x B.x } // possible variations on the syntax
    var xC = C.x
    var zDE = [D.x, E.y] // D.x and E.y are both derived from NotBase.x
    // var xAD = [A.x, D.x] // error: A and D aren't derived from the same mixin
}
```
It may also be desirable to give these rename declarations the ability to be marked `override` and/or provide replacement implementations for functions/getters/setters.

The teaching assistant example would be resolved like this:
```swift
mixin UniversityPerson {
    var computerAccount: Account
}
mixin Teacher: UniversityPerson { ... }
mixin Student: UniversityPerson { ... }

class TeachingAssistant: Teacher, Student { 
    var studentAccount = Student.computerAccount
    var teacherAccount = Teacher.computerAccount
    ... 
}
```

I don't agree that the renaming step should be a required part of the mixing-in process if there are no ambiguities. However, I do that that (re)naming mixin members should be mandatory when encountering an ambiguous situation like the diamond inheritance pattern. At a minimum, it would look like this (same mixin declarations as above):
```swift
struct R: A, B {
    var x = [A.x, B.x] // required
}
struct S: A {} // no ambiguities here, no need to (re)name; var x = A.x is implicit
```

Evertt: @HT154 

> I also propose extending the proposal's syntax to allow multiple mixins that share the same base to share the same renamed binding.

I really like the syntax that you propose. It really brings the best of both worlds together.

> ```swift
> mixin E: NotBase { var y = NotBase.x } // Is it possible to perform renaming even if there are no conflicts?
> ```

I wouldn't see why not.

> It may also be desirable to give these rename declarations the ability to be marked `override` and/or provide replacement implementations for functions/getters/setters.

Yes I would like there to be some kind of mark, but I'm not sure `override` would be the best one for renaming the variables. Since you're not overriding their implementation, but merely renaming them I would say let's create a new mark called `rename`. So it would look like `rename var studentAccount = Student.computerAccount`.

About replacement implementations I would say, only if the mixin marked them as `open`, but I think that was already part of the original proposal.

Anton3: @Evertt @HT154 
I tried another approach that does not use sybtyping, see [some examples](https://gist.github.com/Anton3/b82d3dfbd07599410d234c43e94738aa).
There, mixins are just a sugar for proxy methods. It gets a bit more interesting with abstract types.

Evertt:
@Anton3 so 6 months ago you said:

> Nope, it's been postponed to Swift 4 Phase 2 (another 6 months).

So uh, how's it going? Is there somewhere I can follow progress on this proposal? I really think that if this proposal would get accepted it would help to make everybody's code so much more modular.

Anton3:
@Evertt AFAIKS, there is no way such a new concept would fit even into Swift 4 Phase 2 scope. Apart from that, I see several issues with the proposal draft:

1. Unsolved problems. I know no language where problems with conflicts at mixin merging are solved cleanly. We should either research this brand new concept, or accept some limitations, with which we should live forever.
2. The proposal suggests a new monolithic concept, which is difficult to break into independent parts, such that every part brings some relief to the developers.
3. There is already a similar concept of protocols in the language. Protocols and mixins would have a lot in common, probably more than wanted.

One of the suggestions was to add mixins' features to protocols. But then we still need to talk about conflict resolution. I suggested an idea to provide some syntactic sugar to forward members of a property to the enclosing type. That way I hoped to avoid creating a new entity of mixins and to some extent avoid conflicts.

Evertt:
> 1. Unsolved problems. I know no language where problems with conflicts at mixin merging are solved cleanly. We should either research this brand new concept, or accept some limitations, with which we should live forever.

I thought that the conflict-resolution proposals that you and I and @HT154 talked about in this thread were actually getting pretty close, don't you think?

> 2. The proposal suggests a new monolithic concept, which is difficult to break into independent parts, such that every part brings some relief to the developers.

That's true. So you're saying the scope is just too big for where swift is focussing on right now?

> 3. There is already a similar concept of protocols in the language. Protocols and mixins would have a lot in common, probably more than wanted.

Also true, but I wouldn't want to say that that's a reason not to implement mixins. I believe mixins are way better than protocols in building your app from clean little blocks. Maybe I would rather have mixins + old-school interfaces than have protocols.

Evertt:
So uh, is the Swift team gonna do anything with this?

DougGregor:
I can bring it up with the core team, but I strongly suspect that the addition of a new, top-level declaration kind that's a little like protocols and a little like classes is a non-starter: it's a lot of new surface area and complexity for a small improvement over existing facilities. If you feel strongly about this, the best way to start is to re-pitch on swift-evolution

PR #195, opened 2016-03-07
Tweak the scan-takewhile-dropwhile proposal

kballard:


dabrahams:
@kballard Sorry, looks like you need to resolve conflicts here.

kballard:
@dabrahams PR updated.

dabrahams:
Hi @kballard,

Since the review is going to start in a couple of days I met with my team to discuss this proposal.  We like the ideas and request the following tweaks, which we think will improve the proposal and its chances of acceptance:

1. Update the proposal to account for language and standard library changes, most notably, the first argument label rules and API names have been overhauled since the proposal was written.  Ideally you’d base the standard library on what’s in the [swift-3-indexing-model branch](https://github.com/apple/swift/tree/swift-3-indexing-model). We think this is really essential before the review starts.
2. In the “Detailed design” section it would help for you to explain the remark about “matching expected behavior.” Even the members of the standard library team were scratching our heads over this one until we figured out it’s about our policy of not making multiple passes using a user-defined closure without an explicit appearance of `.lazy`
3. We would prefer that you change `scan`’s first argument label from `initial` to `startingWith`, and propose we update `reduce` to use the same argument label.  Whatever we do, `reduce` and `scan` should be consistent in this respect.
4. In lieu of `iterate`, we'd prefer to see two overloads of `unfold` as follows:

    ```swift
    func unfold<T, State>(
      _ initialState: State, applying nextElementAndState: (State)->(T, State)?
    ) -> UnfoldSequence<T>

    func unfold<T>(
      _ initialElement: T, applying nextElement: (T)->T?
    ) -> UnfoldSequence<T> {
      return unfold(initialElement) { state in nextElement(state).map { (state, $0) } }
    }
    ```

Of course, if you disagree with any of these changes you should not make them, and we can discuss them on-list during the review period. 

Thanks,
Dave

kballard:
@dabrahams The problem with the `unfold()` you suggested is that which I described in my proposal, namely that the overload that's equivalent to `iterate()` ends up evaluating the closure one step in advance. By that I mean the following code:

```swift
let seq = unfold(0, applying: { print("step"); return $0 + 1 })
var gen = seq.generate()
print(gen.next())
```

will end up printing

> ```
step
0
```

even though the only element retrieved is the initial element, i.e. the one that doesn't require evaluating the closure to produce.

**Comment by kballard** [comment]
Also to note, #261 already updated the proposal for the Swift 3 API names, though that doesn't address first argument labels.

**Comment by dabrahams** [comment]
on Tue Apr 26 2016, Kevin Ballard <notifications-AT-github.com> wrote:

> @dabrahams The problem with the unfold() you suggested is that which I described
> in my proposal, namely that the overload that's equivalent to iterate() ends up
> evaluating the closure one step in advance. By that I mean the following code:
>
> let seq = unfold(0, applying: { print("step"); return $0 + 1 })
> var gen = seq.generate()
> print(gen.next())
>
> will end up printing
>
>     step
> 0
>
> even though the only element retrieved is the initial element, i.e. the one that
> doesn't require evaluating the closure to produce.

1. I'm not sure that's a problem.

2. Regardless, you can implement it any way you desire.  We're just
   asking that the signatures match up.

-- 
Dave

**Comment by dabrahams** [comment]
Oh, and to follow up, here’s a little implementation I whipped up:

```swift
func unfold<T, State>(
  _ initialState: State,
  applying nextElementAndState: (State)->(T, State)?) -> AnySequence<T> {
  var state = initialState
  return AnySequence {
    AnyIterator {
      if let (element, newState) = nextElementAndState(state) {
        print("element:", element, "newState:", newState)
        state = newState
        return element
      }
      print("nil!")
      return nil
    }
  }
}

struct CollectionOfTwo<T> : Collection {
  init(_ first: T, _ second: T) {
    self.elements = (first, second)
  }
  
  var startIndex: Int { return 0 }
  var endIndex: Int { return 2 }
  subscript(i: Int) -> T { return i == 0 ? elements.0 : elements.1 }
  
  let elements: (T, T)
}

func unfold<T>(_ initialElement: T, applying nextElement: (T)->T?) -> AnySequence<T> {
  let notLazyEnough = unfold(initialElement) {
    state in nextElement(state).map { (state, $0) }
  }
  
  return AnySequence(
    CollectionOfTwo(
      AnySequence(CollectionOfOne(initialElement)),
      notLazyEnough.lazy.dropFirst()
    ).lazy.flatten()
  )
}
```

dabrahams:
on Tue Apr 26 2016, Kevin Ballard <notifications-AT-github.com> wrote:

> Also to note, #261 already updated the proposal for the Swift 3 API names,
> though that doesn't address first argument labels.

I think you may need to take another pass at it.  I still see plenty of
`SequenceType`s and `Generator`s in there.

-- 
Dave

kballard:
@dabrahams I submitted #275 with the changes. I elected not to rename `initial` to `startingWith` because I think that renaming the argument for `reduce` is a bit out of scope of this proposal, but I made the other requested changes.

Also, FYI, your little implementation is a bit buggy (it still evaluates the function one step too early once you get past the first element in the sequence, and it terminates one step too soon if `nextElement` returns `nil`).

dabrahams:
on Thu Apr 28 2016, Kevin Ballard <notifications-AT-github.com> wrote:

> @dabrahams I submitted #275 with the changes. I elected not to rename initial to
> startingWith because I think that renaming the argument for reduce is a bit out
> of scope of this proposal, 

It's only out of scope because it's not in the proposal.

> but I made the other requested changes.

Well, as I said in my comment, IMO reduce and scan should use the same
argument label, whatever it is.  You could have changed it to match
reduce's argument label.
>
> Also, FYI, your little implementation is a bit buggy (it still evaluates the
> function one step too early once you get past the first element in the sequence,
> and it terminates one step too soon if nextElement returns nil).

In that case, I'm looking forward to seeing a correct one—and the tests
that prove it. ;-)

-- 
Dave

kballard:
It does match reduce's argument label.

I pulled from master yesterday (16a875930) and `reduce` is declared as

```swift
@warn_unused_result
public func reduce<T>(_ initial: T, combine: @noescape (T, ${GElement}) throws -> T) rethrows -> T
```

And that matches the declaration of `scan` here:

```swift
@warn_unused_result
public func scan<T>(_ initial: T, @noescape combine: (T, Self.Iterator.Element) throws -> T) rethrows -> [T]
```

(although now that I list them out, I notice that `@noescape` was moved, which I didn't account for when updating the proposal)

dabrahams:
on Fri Apr 29 2016, Kevin Ballard <notifications-AT-github.com> wrote:

> It does match reduce's argument label.
>
> I pulled from master yesterday (16a875930) and reduce is declared as
>
> @warn_unused_result
> public func reduce<T>(_ initial: T, combine: @noescape (T, ${GElement}) throws -> T) rethrows -> T
>
> And that matches the declaration of scan here:
>
> @warn_unused_result
> public func scan<T>(_ initial: T, @noescape combine: (T, Self.Iterator.Element) throws -> T) rethrows -> [T]
>
> (although now that I list them out, I notice that @noescape was moved, which I
> didn't account for when updating the proposal)

Ah, thanks, then!

FWIW, the reason we asked for a change to the argument labels all around
is that we only have time for so many reviews before we have to lock
down Swift 3.  A proposal just for changing these two labels is a pretty
inefficient use of community resources.

-- 
Dave

kballard:
Fair enough. That said, since the current `reduce` method does not have an external parameter name for the first argument, I wasn't sure if you were suggesting that it (and `scan`) should be given an external parameter name, or if you just wanted the internal parameter name to be changed.

dabrahams:
on Fri Apr 29 2016, Kevin Ballard <notifications-AT-github.com> wrote:

> Fair enough. That said, since the current reduce method does not have an
> external parameter name for the first argument, I wasn't sure if you were
> suggesting that it (and scan) should be given an external parameter name, or if
> you just wanted the internal parameter name to be changed.

Good point; now I'm not sure either! :-)

-- 
Dave

moiseev:
IIRC here is the argument for the first argument label:

```Swift
xs.scan(0) { $0 + 1 }
```

Since `scan` is an active verb, it reads more like an action on its first argument. Name itself falls into the _term-of-art_ category, so is OK to be used, but when combined with the proper label, the invocation would be unambiguous:

```Swift
xs.scan(startingWith: 0) { $0 + 1 }
```

Same reasoning applies to `reduce`.

kballard:
I consider that to be an argument in favor of not using trailing-closure syntax for functional methods (i.e. the [Rule of Kevin](http://ericasadun.com/2015/11/17/a-handful-of-swift-style-rules-swiftlang/)).

```swift
xs.scan(0, combine: { $0 + 1 })
```

dabrahams:
on Fri Apr 29 2016, Kevin Ballard <notifications-AT-github.com> wrote:

> I consider that to be an argument in favor of not using trailing-closure syntax
> for functional methods (i.e. the Rule of Kevin).
>
> xs.scan(0, combine: { $0 + 1 })

I don't see how that clarifies anything.

  x.verbPhrase(argumentWithoutLabel, ...)

makes argumentWithoutLabel look like the direct object of the
verbPhrase, and clearly we are scanning xs in this case, not 0.

-- 
Dave

kballard:
I guess you're right. I just feel like adding the argument label here might be too unnecessarily verbose.

PR #199, opened 2016-03-09
Proposal: Listing all possible values of a type

brentdax:
This is a proposal for the long-discussed, frequently-requested, never-reviewed "give us a way to get all enum cases" feature.

### DIFF
--- a/proposals/0000-allvalues.md
+++ b/proposals/0000-allvalues.md
@@ -1,0 +1,157 @@
+# Listing all possible values of a type
+
+* Proposal: [SE-NNNN](https://github.com/apple/swift-evolution/blob/master/proposals/NNNN-name.md)
+* Author(s): [Brent Royal-Gordon](https://github.com/brentdax)
+* Status: **Awaiting review**
+* Review manager: TBD
+
+## Introduction
+
+This proposal adds a `ValuesEnumerable` protocol to the standard library.
+Conforming to this protocol requires your type to offer a static `allValues`
+property, vending a collection which includes all permitted values of the 
+type. The Swift compiler will automatically derive an implementation for 
+this property in certain cases, particularly for simple `enum`s.
+
+Swift-evolution thread: Several, most recently [Pre-proposal: CaseEnumerable protocol (derived collection of enum cases)](https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20160111/006876.html)
+
+## Motivation
+
+Sometimes, particularly when using enums, developers want to fetch all 
+possible instances of a type.
+
+This desire manifests in many different ways with varying degrees of 
+hackiness. Some developers [create an array, but have to keep it up to date](http://stackoverflow.com/questions/24033782/how-to-get-all-values-of-an-enum-in-swift).
+Others [keep a count of the cases and use `RawRepresentable` to construct them](http://stackoverflow.com/questions/27094878/how-do-i-get-the-count-of-a-swift-enum),
+presuming they are all consecutive and start at 0. Different developers use 
+different names for this operation. And all possible solutions require you 
+to either specify redundant information, rely on implementation details, or 
+both.
+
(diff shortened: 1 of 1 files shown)


jessesquires:
Hey @brentdax ! :smile: 

How is this different from #114?

brentdax:
Partly, #114 seems to have stalled and I thought the feature deserved a nudge. The two proposals are quite similar, but they do differ in minor but significant ways:

1. There is an underlying difficulty with both proposals: `allValues` should belong to an associated type conforming to `Collection where .Iterator.Element == Self` (to use the new terminology), but this cannot currently be expressed in Swift. The two proposals address this problem in different ways.

    1. In this proposal, `allValues` is defined in the protocol and is (perhaps temporarily) an `Array<Self>`. In #114, `allValues` is not defined in the protocol, but is added to all types marked with the protocol, which means the protocol can't be used for generic programming. I believe it's more important to model the relationship; #114 believes it's more important to have looser typing so that implementations can use more efficient collection types.

    2. In this proposal, the eventual change to the type of `allValues` is explicitly included as part of the proposal. In #114, a workaround (the introduction of an `AnyCollection<Element>` type) is mentioned as a future direction. That's partly because this proposal was drafted after the Generics Manifesto, while #114 was drafted before.

2. In this proposal, the semantic meaning of `allValues` is defined (for an `Equatable` type, all possible instances must `==` at least one of its elements; for other types, all possible instances must `~=` at least one of its elements). #114 makes no attempt to define this, which leaves its potential meaning for structs, classes, and protocols unclear.

3. In this proposal, the `allValues` property is defined to return values in their `Comparable` order if one is defined, or in source order otherwise. (That means that, if an enum is `Comparable`, Swift may not always be able to derive its `allValues` property.) In #114, the `allValues` property is defined to return values in source order. 

4. This proposal sets minimum requirements for auto-deriving an `allValues` property, but allows derivation above that to be implementation-defined to avoid having to formally propose future extensions. #114 specifically defines the enums which should have an `allValues` property derived and lists extensions of that capability as something to be added through the evolution process.

5. This proposal includes doc comments on the APIs it proposes. #114 includes more discussion of alternatives and extensions.

In general, this proposal is more focused on defining the meaning of the `ValueEnumerable` protocol, while the automatic derivation of an `allValues` property is almost an afterthought. #114 is more focused on automatically deriving the `allValues` property, with the `ValueEnumerable` protocol primarily serving as a way to invoke that behavior.

I prefer my approach, but both of them will get us to similar places.

jessesquires:
Ahh, I see. 

I think it would be more productive and efficient to collaborate with @jtbandes on #114, discuss the differences on the mailing list (most recent response on the original thread [here](https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20160307/012177.html), from today), and then update #114 according to everyone's feedback. :smile:

jtbandes:
I'd be happy to discuss this further and update #114 if we can come up with something better. Please feel free to bring this up again on the mailing lists. I might not have enough time to contribute anything to the discussion in the next week or two, but I'll get back to it eventually.

brentdax:
@jtbandes Let me know when you're a little less busy and I'll post something on the list. (Or just post it yourself.)

jtbandes:
Re. 2: I'm happy to adopt your explanations of the semantic meaning, although we should think carefully about the wording, specifically usage of "must", if we're allowing users to write their own implementations (i.e. if `allValues` is exposed in the protocol).
Re. 4: I'm happy to adjust the wording to clarify that derivations not covered by the proposal are implementation-defined, but I'm skeptical that significant changes could be made without going through the evolution process.
Re. 5: That's a fairly minor point. I think we could combine them and offer both, or just drop the "future directions".

Re. the rest: I'm drafting an email to swift-evolution so we can keep discussing and hopefully come to some conclusions in the near future. I'll gladly share the draft with you, and I think it'll make sense to have both our names as co-authors on the proposal.

PR #210, opened 2016-03-12
Referencing Objective-C key-paths

hartbit:
After discussing this proposal on the mailing list, with only positive feedback, I am now ready for having it reviewed.

### DIFF
--- a/proposals/XXXX-objc-keypaths.md
+++ b/proposals/XXXX-objc-keypaths.md
@@ -1,0 +1,63 @@
+# Referencing Objective-C key-paths
+
+* Proposal: [SE-XXXX](https://github.com/apple/swift-evolution/blob/master/proposals/XXXX-objc-keypaths.md)
+* Author(s): [David Hart](https://github.com/hartbit)
+* Status: TBD
+* Review manager: TBD
+
+## Introduction
+
+In Objective-C and Swift, key-paths used by KVC and KVO are represented as string literals (e.g., `"friend.address.streetName"`). This proposal seeks to improve the safety and resilience to modification of code using key-paths by introducing a compiler-checked expression.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


DougGregor:
Please clarify the semantics of key paths that refer into collections, e.g., the example:

    #keypath(Person.friends.firstName)

how is #keypath supposed to determine that "firstName" is meant to be a property on the element type of the "friends" collection? Is it tied to some KVC convention? Conformance to the Collection protocol? Something else? Otherwise, the proposal looks ready to go up for review.

hartbit:
Would it be enough to say that paths that reference:

- a type conforming to SequenceType are allowed to add a key to reference properties on that type and properties on the Element type
- a type conforming to NSArray, NSDictionary, NSSet are allowed to add a key to reference properties on that type

Which leaves a whole for referencing properties on types inside a NSArray/NSDictionary/NSSet.

Any ideas?

hartbit:
I added necessary details.

DougGregor:
Three minor comments and then I'd like to get this scheduled for review:
1) Per recent discussions about the naming of the `#` expressions/directives, we should call this `#keyPath`
2) I think the SequenceType conformance makes sense; the second bullet (about NSArray/NSDictionary/NSSet) isn't necessary because its behavior falls out from NSArray/NSDictionary/NSSet conforming to (or not conforming to) SequenceType.
3) There is some implicit bridging going on here that could use some detailed design. If I refer to "Person.lastName.uppercased", that's a method on the value type String. At runtime, we're depending on getting the uppercaseString method on NSString. This may be as simple as saying that we follow the _ObjectiveCBridgeable conformance for any value type encountered along the way.

hartbit:
1) I changed it to `#keyPath`
2) I simplified that whole paragraph to only mention the conformance to `SequenceType`.
3) I hope you don't mind, but I copy pasted your third remark as-is under the `uppercased` example for potential implementors to see (it's chinese for me :))

hartbit:
@DougGregor Do you need anything else before it's merged?

DougGregor:
My apologies for the delay; haven't have much -evolution time lately.

DougGregor:
Let's just start it now...

PR #215, opened 2016-03-18
Remove explicit use of `let` from Function Parameters

Nirma:
First draft of proposal.


NinoScript:
If accepted, I think it would also be worthwhile to discuss if we could also remove `var` and `let` from switch cases with associated values.

Meanwhile, here are some typos I found in your draft:

- **keword** in `The let keword would have to be deleted` should be **keyword**.
- **paramenters** in `not label function paramenters as let` should be **parameters**.
- **redundent** in `Leave the redundent syntax in place` should be **redundant**.

Nirma:
@NinoScript Thanks for catching that! I have corrected the typos and added a little more detail to the motivation section and the section about impact on existing code.
Hows it looking now?

PR #216, opened 2016-03-18
Add proposal to abolish IUO type from Swift.

cwillmor:
I'd like to formally submit this proposal for consideration. It's been under discussion in draft form on swift-evolution for the past few days.


cwillmor:
Thank you!

lattner:
Thank you for driving this!

PR #219, opened 2016-03-21
"Make pointer nullability explicit using Optional"

jrose-apple:
As discussed and largely approved of on swift-evolution. I've left the UnsafeBufferPointer issue as an "open issue" in order to get more feedback during the formal review period; I feel it's important to get more opinions from the people using UnsafeBufferPointer today, since it's a relatively niche API.


jrose-apple:
Updated after some discussion to go with the simplest answer for UnsafeBufferPointer's API.

PR #222, opened 2016-03-22
Correct typo in #selector

onmyway133:
At first, I thought that `_:at:` is enough. But it seems it is actually a typo. This corrects that


DougGregor:
The method is call insertSubview(_:atIndex:) in Swift 2 and insertSubview(_:at:) in Swift 3. I'll accept the PR because people are likely to be using this with Swift 2.2.

PR #223, opened 2016-03-22
Fixed a comment.

marcusrossel:
Changed an occurrence of `==` to `=`, as this style is used in the other comments.


lattner:
Thanks!

PR #224, opened 2016-03-23
Suggested handling of polymorphic selectors

jhaungs:
This pull request is per email with Doug Gregor.

### DIFF
--- a/proposals/0022-objc-selectors.md
+++ b/proposals/0022-objc-selectors.md
@@ -135,12 +135,20 @@ syntax at the completion of the public review.
 ## Polymorphic Selectors
 
 It's common to use identical method names in different classes when the methods do the same conceptual thing.
-For example, several view classes might use the method name `panned` as the action selector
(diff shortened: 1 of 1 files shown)


DougGregor:
I think the terms "polymorphic" and "monomorphic" are a bit confusing here. Can we avoid those terms, and instead talk about this as an approach for dealing with cases where there is no (single) natural Swift method to refer to in `#selector`? It's relevant to both the "provided-by-several-disjoint-classes" and the "dynamically-generated-and-has-no-Swift-declarations" use cases. (Thanks!)

lattner:
@DougGregor please review the update here, and close or merge.  Thanks!

lattner:
@DougGregor please review the update here, and close or merge, this is an edit to a proposal not a new proposal. Thanks!

lattner:
This PR is stale and conflicts, I'm closing it now.  If there is still something that needs to happen, please create a new PR.  Thanks!

PR #231, opened 2016-03-27
my xcode fixed it like that

ilaipi:


lattner:
The example in the document is correct, it is trying to point out the manually curried syntax.  Your suggestion means something else.

PR #234, opened 2016-03-28
adjustments to SE-0025 after the first review

ilyathewhite:


lattner:
It looks like there are some merges included in the commit history.  Please squash them, thanks!

PR #237, opened 2016-03-31
Added notes about further discussion on swift-evolution.

tjw:


lattner:
typo "yields"

tjw:
Oops -- thanks!

PR #238, opened 2016-03-31
Add implicit numeric conversions to out-of-scope list

brentdax:
This came up again on the list, and @lattner indicated it should be added to the repository.

I wasn't sure whether it should be added to the Commonly Proposed Changes list or the Out of Scope list. I ended up picking Out of Scope because it seems like we do intend to implement it eventually. However, I also put a note in the commonly proposed list to remind people to check the out-of-scope list too.


lattner:
Awesome, thanks Brent!

PR #244, opened 2016-04-01
Enforcing order of defaulted parameters

jckarter:


LEOYoon-Tsaw:
I would like the opposite. Thanks to labeled parameters, reordering gives great freedom while keeping clarity. It's cocoa that need to adapt to Swift instead of the reverse.

jckarter:
@LEOYoon-Tsaw Thanks for your feedback, but it'd be easier for us to keep track of the discussion if you participated in the swift-evolution mailing list or review process. We don't follow discussions on these Github issues.

PR #253, opened 2016-04-12
Operator precedence

Anton3:
Replace syntax of operator definition, and replace numerical precedence with partial ordering of operators:

```swift
// Before
infix operator <> { precedence 100 associativity left }

// After
precedencegroup Comparative {
  associativity(left)
  precedence(> LogicalAnd)
}
infix operator <> : Comparative
```


Anton3:
Removed declarations of special operators from grammar. I added this feature myself, noone gave positive review of it, and now I think, that was a mistake. It was largely inconsistent. I changed the proposal back to a previous version.

PR #258, opened 2016-04-19
Add a link to the SE-0019 Proposal

chriseidhof:
To be consistent with all other proposals.


lattner:
Thanks @chriseidhof !

PR #261, opened 2016-04-20
Update SE-0045 for Swift 3 API Guidelines

kballard:
I believe I've applied the Swift 3 naming conventions correctly, but
please let me know if I missed anything.


kballard:
I also went ahead and renamed `takeWhile(_:)` to `prefix(while:)` since I expect that to have more community support.

lattner:
Fantastic, thank you @kballard !

PR #262, opened 2016-04-21
SE-0016: Amend to include OpaquePointer (post-acceptance).

jrose-apple:
The proposal author @MichaelBuckley, code owner @gribozavr, and I all consider this an omission from the original proposal. As particular evidence, OpaquePointer already has the reverse initializer, `init(bitPattern: Int)`.

See also apple/swift#2238.


DougGregor:
The core team has discussed this and agrees that it's an omission from the original proposal. Accepting the PR as a bug fix.

PR #266, opened 2016-04-23
Make Optional Requirement Objective-C-only

DougGregor:


lattner:
This LGTM, I'll review managerize it.

proposals/NNNN-optional-requirements.md
```diff
@@ -0,0 +1,185 @@
...
+us take the labels out of the function type:
+
+```swift
+@objc protocol NSTableViewDelegate {
+  var tableView(_:viewFor:row:): ((NSTableView, NSTableColumn, Int) -> NSView?)? { get }
+  var tableView(_:heightOfRow:): ((NSTableView, Int) -> CGFloat)? { get }
+}
+```
+
+By itself, that is a good feature. However, we're not dont, because we
```
jpsim: "we're note dont" should be "we're not done"

lattner: Fixed, thanks @jpsim!

PR #276, opened 2016-04-29
SE-0032 updated with swift 3 names

masters3d:


moiseev:
@masters3d do you mind going though this proposal once more to reflect recent changes in Swift 3?
- `@noescape` should now be attached to the type of the argument, not to the label
- first argument label rule means that `find(_:)` would become `find(predicate:)`, which is not the way it was designed.

masters3d:
@moiseev updated.

moiseev:
Thanks @masters3d!

PR #285, opened 2016-05-03
Add "Use pattern-matching in `if let`" to commonly_proposed.md.

jrose-apple:
Since it just came up again recently.

@DougGregor or @lattner, okay to merge?


lattner:
Adding this makes sense, but please include some points about why it doesn't work out well in practice:

1. Most devs don't think about things in "pattern matching" terms, they think about "destructuring".
2. The most common use case for "if let" is actually for optional matching.
3. Thus, going with this proposal punishes the common case syntactically, and also makes the learning curve for swift better.
4. The current design of "if case" unifies "pattern matching" around the 'case' keyword, which is googlable.

lattner:
pre approved with changes to bulk up the reason why it is a bad idea :-)

lattner:
104cdde addresses this.

jrose-apple:
Thanks, Chris; sorry for not getting back to this.

PR #286, opened 2016-05-03
Bring SE-0016's revision history in line with other proposals.

jrose-apple:
The proposal was tweaked post-acceptance in 4eae3025, but @dabrahams pointed out that we have a standard format for this now. No content change.

Look good, Dave?


dabrahams:
LGTM.

PR #288, opened 2016-05-03
Decoupling Floating Point Strides from Generic Implementations

xwu:
Erica and I are hoping to resurrect this proposal, updating it for changes made in the interim.

### DIFF
--- a/proposals/0050-floating-point-stride.md
+++ b/proposals/0050-floating-point-stride.md
@@ -68,7 +68,7 @@ public struct FloatingPointStrideToIterator<
     let quotient = (end - _start) / stride
     // FIXME: Maximum supported number of steps could be slightly larger
     _precondition(
-      quotient > 0 && quotient.ulp <= 1,
+      quotient >= 0 && quotient.ulp <= 1,
       "can't construct FloatingPointStrideToIterator: maximum supported number of steps exceeded")
     self._start = _start
     self._end = end
--- a/proposals/0050-floating-point-stride.md
+++ b/proposals/0050-floating-point-stride.md
@@ -40,17 +40,13 @@ The same issues occur with C-style for loops. This problem is a fundamental arti
 
 ## Detail Design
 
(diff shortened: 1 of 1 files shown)


dabrahams:
Why does this apply only to `BinaryFloatingPoint` and not `FloatingPoint` in general?

xwu:
Two reasons:

(1) Decimal types (such that may exist in the future) do not accumulate error using the "old" stride algorithm

(2) The maximum exactly representable contiguous integer value, or whatever you want to call it, can only be calculated using significandBitCount, which is a requirement of conformance to BinaryFloatingPoint and not FloatingPoint

xwu:
You are correct, though, that a hypothetical floating point type with radix 3 would not be adequately accommodated. However, the division of labor between FloatingPoint and BinaryFloatingPoint may itself need some re-examination if TernaryFloatingPoint ever becomes a thing.

xwu:
Some further thoughts on this: error arises because we try to represent the results of base 10 arithmetic in base 2. It is possible that someone may wish to create a ternary floating point type, but one sensible use for such a type would be representing the result of operations performed in base 3, not base 10. A TernaryFloatingPoint type that performs ternary arithmetic would not accumulate error on repeated multiplication.

dabrahams:
Decimal representation guarantees that the literal you write in decimal notation can be exactly represented, but it doesn’t prevent error from accumulating when the precision of the representation would need to be exceeded.  This accumulation of error is inherent to floating point with any fixed precision.

xwu:
Of course, d'oh. I'll reflect on how to compute the max exactly representable consecutive integer using the facilities found in FloatingPointType. It might not be pretty.

xwu:
I've replaced `BinaryFloatingPoint` with `FloatingPoint`. Calculations about how many steps are supported now rely on `((end - start) / stride).ulp <= 1`. In the context of a `StrideTo<Double>`, sometimes that means the number of iterations tops out at 2 ** 53 - 2 instead of 2 ** 53. That isn't so elegant but it practice it should do.

lattner:
@stephentyrone Can you give this a sanity check?  I'd like to make sure it looks basically ok to you before merging it, thanks!

stephentyrone:
@lattner I think this is good enough to proceed.  There's may be some discussion to be had about what the type of `_step` should be, but I think that swift-evolution is a good place for that.

lattner:
thanks!

erica:
Thank you all!

PR #290, opened 2016-05-04
Generic Tuple Extensions

CodaFi:


lattner:
@DougGregor your opinion is welcome on this.  To me, this (sadly) seems clearly out of scope for Swift 3.

DougGregor:
Sadly, yes, this is out of scope for Swift 3. The proposal looks roughly reasonably, but implementing it is a fairly significant undertaking because the assumption that only nominal types can conform to protocols is pervasive. So even if this proposal were perfect, it couldn't happen in Swift 3. Please bring this up again once we start scoping the next major Swift release.

PR #292, opened 2016-05-04
Update 0003-remove-var-parameters.md

huguesbr:
Adding missing `doSomethingWithVar` function implementation


lattner:
Ok, thanks.

huguesbr:
:+1

PR #295, opened 2016-05-05
Update SE-0045 with results of review

kballard:
`Sequence.prefix(while:)` and `Sequence.drop(while:)` were accepted, but
`scan(_:combine:)` and `unfold(_:applying:)` were rejected.


lattner:
Thank you!

PR #308, opened 2016-05-08
Adding link to Douglas Gregor's generics manifesto

austinzheng:
Adding a link to the formatted version of the *Generics Manifesto*. This allows folks perusing the repo to easily see exactly what changes are envisioned for Swift 3.0 (and later) with regards to generics functionality.

Note that I don't think there is a way to specify a relative link across repositories, so the link is an absolute link that points to the copy of the document in the Apple `swift` repository's docs folder.


DougGregor:
LGTM, thanks!

PR #310, opened 2016-05-09
SE-0048 and SE-0057: Move to implemented list

benrimmington:


lattner:
I have a few bugs left to fix in generic type aliases before I declare SE-0048 done.

lattner:
I moved 0057 in e8a4404, thanks!

PR #318, opened 2016-05-12
Updated failable numeric initializers proposal

anandabits:
Removed comments about Foundation API.


lattner:
Thanks @anandabits!

PR #319, opened 2016-05-13
Correct author name and links in SE-0081

pyrtsa:


pyrtsa:
Ping @lattner, this should be a no-brainer.

PR #324, opened 2016-05-16
Spell out the protocol changes for modernized object literals.

rjmccall:


erica:
Read and understood. No objections.

PR #325, opened 2016-05-16
A proposal to add the base property to standard slice types

moiseev:


dabrahams:
Next time you assign a number to a proposal, please update README.md also.

lattner:
@dabrahams That isn't really practical for contributors to do, nor is it the standard protocol - they don't know what order evolution proposals will get merged in, so they can't number them.

PR #330, opened 2016-05-17
Reconsider semantics for optional methods in Swift

carlosrodriguez85:


proposals/00NN-reconsider-optional-semantics.md
```diff
@@ -0,0 +1,86 @@
...
+
+Swift-evolution thread: [Modify optional method semantics for swift](https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20160509/016893.html)
+
+## Motivation
+
+As a everybody knows, `optional` keyword is present in Swift solely to maintain compatibility with objective-c protocols. Optional methods in protocols are neither a Swift capability, nor a wanted one.In fact, proposal [0070] intends to begin the transition to an optionals-free language, since optionals, as presented by Objective-C, are a manner to avoid interface segregation.
+
+Nonetheless, in many occasions, optionals are used to model customized behavior vs default one. For instance, if you take a look at the documentation of UITableViewDataSource or delegate, you’ll see that optional methods are not required, since the framework provides a default behavior that can be customized by implementing the corresponding optional methods.
+
+Consequently, is most cases, optional methods in Cbjective-C are a means to replace the semantics of *default* protocol method implementations, as supported through extensions in swift.
```
JustinJiaDev: Objective-C instead of Cbjective-C

lattner:
This should be discussed more in depth on swift-evolution before proposing this.  Specifically, it is not the case that optional methods imply a default implementation. The runtime is able to detect whether or not an optional method is implemented, which is the whole point in ObjC.

PR #334, opened 2016-05-19
Change the swift-evolution template link to use Gmane

kballard:
Gmane links are typically easier to read than the official archives, so
we should encourage its use.


lattner:
Are gmane links long-term stable?

kballard:
I'm pretty sure they are. Gmane's official purpose is as a mailing list archive, and it would be a poor archive indeed if its links weren't stable.

PR #337, opened 2016-05-20
Adding Adrian as author to 0095

austinzheng:
(see title)

### DIFF
--- a/proposals/0095-any-as-existential.md
+++ b/proposals/0095-any-as-existential.md
@@ -1,7 +1,7 @@
 # Replace `protocol<P1,P2>` syntax with `Any<P1,P2>`
 
 * Proposal: [SE-0095](0095-any-as-existential.md)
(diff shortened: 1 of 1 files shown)


lattner:
@DevAndArtist look ok to you?

anandabits:
This still needs to be updated to reflect the decision to use lowercase `any` rather than uppercase `Any`.

austinzheng:
@anandabits I'll open another PR after this one goes in.

DevAndArtist:
@lattner @austinzheng feel free. ;)

DevAndArtist:
@lattner I'm probably not available on May 28 and hardly available on May 29. (Time difference: 19:00 at my time equals 9-10:00 when Apple keynote starts.) But I don't think this should be problem.

austinzheng:
@lattner, whenever you have a moment this PR is good to go.

PR #343, opened 2016-05-24
Simplifying Condition Clauses

erica:
This proposal adjust Swift grammar to enable the arbitrary mix of expressions among conditions rather 
than constraining them before other conditions. Under this proposal, expressions are no longer limited
to where clauses after the initial list of Boolean conditions.

Swift-evolution thread:
[\[Pitch\] making where and ,	interchangeable in guard conditions](http://thread.gmane.org/gmane.comp.lang.swift.evolution/17926)


erica:
Although I can probably fix this up to specify that expressions must be Boolean, I'm going to leave this for now as it's not pressing. Thanks again.

PR #348, opened 2016-05-30
Updated SE-0099 for case-conditions and case-item-lists

erica:
Updated SE-0099 to make clear that case-conditions (used in while, if, and guard statements) will no longer support where-clauses but case-item-lists (used in switch statements) will continue to do so


erica:
@lattner These are the changes I emailed you about yesterday.

lattner:
Thanks Erica!

PR #349, opened 2016-05-31
Proposal: `return` consistency for single-expressions

DevAndArtist:
Any single-expression closure can omit the `return` statement. This proposal aims to make this feature more consistent in some other corners of the language.

Original swift-evolution thread: [\[Pitch\] Make `return` optional in computed properties for a single case](https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20160523/019260.html)


lattner:
Given that Swift 3 is winding down, we are in a mode of declining PRs for proposals that don't align with its goals.  Please bring this back up for discussion this fall, thanks for understanding.

PR #351, opened 2016-06-02
Adding SE-0095 version 2 proposal

austinzheng:
@lattner @DougGregor


DougGregor:
Thanks for the quick turnaround!

austinzheng:
@lattner Just wanted to check if this proposal still needs to be scheduled. It would be nice to have it in for Swift 3.

PR #352, opened 2016-06-02
Changed capitalizedString -> capitalized for consistency

fluidsonic:
It looks like a mistake where `capitalizedString` was changed to just `capitalized` two times but another time it remains `capitalizedString`.


DougGregor:
Good catch, thanks!

PR #355, opened 2016-06-04
[Gardening] Minor changes

austinzheng:
@lattner 

- Moved SE-0048 into the 'implemented for Swift 3' section
- Added an entry to the 'commonly rejected' list regarding `|`-style type constraints, per SE-0095's rationale


lattner:
Thanks!

PR #361, opened 2016-06-09
Fixed a few typos / inconsistencies

tonisuter:
This pull request contains corrections of a few typos and inconsistencies in various proposals.

### DIFF
--- a/proposals/0065-collections-move-indices.md
+++ b/proposals/0065-collections-move-indices.md
@@ -115,7 +115,7 @@ haven't lost the ability to express anything.
 This section covers the proposed structural changes to the library at
 a high level.  Details such as protocols introduced purely to work
 around compiler limitations (e.g. `Indexable` or `IndexableBase`) have
-been omitted.  For a complete view of the the code
+been omitted.  For a complete view of the code
 and documentation changes implementing this proposal, please see this
 [pull request](https://github.com/apple/swift/pull/2108).
 
@@ -245,7 +245,7 @@ violation in the general case.
 
 This section describes changes to methods, properties, and associated
 types at a high level.  Details related to working around compiler
-limitations have been omitted.  For a complete view of the the code
+limitations have been omitted.  For a complete view of the code
 and documentation changes implementing this proposal, please see this
 [pull request](https://github.com/apple/swift/pull/2108).
 
@@ -472,7 +472,7 @@ in the interest of full disclosure:
   manipulations end up looking like free function calls:
 
   ```swift
-  let j = index(after: i)           // self.successor(i)
+  let j = index(after: i)           // self.index(after: i)
   let k = index(j, offsetBy: 5)     // self.index(j, offsetBy: 5)
   ```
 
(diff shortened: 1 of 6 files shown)


proposals/0065-collections-move-indices.md
```diff
@@ -472,7 +472,7 @@ in the interest of full disclosure:
   manipulations end up looking like free function calls:
 
   ```swift
-  let j = index(after: i)           // self.successor(i)
+  let j = index(after: i)           // self.index(after: i)
```
tonisuter: I am not entirely sure about this change, but I think the existing comment makes no sense.

proposals/0069-swift-mutability-for-foundation.md
```diff
@@ -122,7 +122,7 @@ let myOtherDate = Date()
 myOtherDate.addTimeInterval(60) // Error, as expected
 ```
 
-It is important to remember that the `Date` API author still controls the methods available on the type, and does not have to provide mutability for every property (as they would in a C struct). For types where we want to provide limited mutability, we can make all properties `get` only and add `mutable` methods to tightly control state and maintain internal consistency. For example, in the case of `Date` the `NSTimeInterval` ivar is still private to the implementation while we provide a `mutable func` to add a time interval.
+It is important to remember that the `Date` API author still controls the methods available on the type, and does not have to provide mutability for every property (as they would in a C struct). For types where we want to provide limited mutability, we can make all properties `get` only and add `mutating` methods to tightly control state and maintain internal consistency. For example, in the case of `Date` the `NSTimeInterval` ivar is still private to the implementation while we provide a `mutating func` to add a time interval.
```
tonisuter: When talking about methods, I changed the text to use 'mutating' instead of 'mutable'. I think that makes more sense semantically and the keyword is called 'mutating'.

proposals/0069-swift-mutability-for-foundation.md
```diff
@@ -220,8 +220,6 @@ URLComponents | NSURLComponents
 URLQueryItem | NSURLQueryItem
 UUID | NSUUID
 
-These new `struct` types will be implemented in the Swift overlay. Immutable/mutable pairs (e.g. `Data` and `MutableData`) will become one mutable struct type.
```
tonisuter: I removed this sentence, because a very similar sentence appears right before the table.

proposals/0085-package-manager-command-name.md
```diff
@@ -103,7 +103,7 @@ updated the proposal to emphasize only `swift build`.
 We considered using `swift build` as the top level command for the package
 manager and moving the other verbs from being flags to being subcommands of
 `swift build`, instead of introducing a `package` command (e.g., `swift
-init`). We think this reads poorly and is less clear than making them `package`
+build init`). We think this reads poorly and is less clear than making them `package`
```
tonisuter: Again, I am not entirely sure about this change, but I think the author meant to write 'swift build init' instead of 'swift init'.

proposals/0065-collections-move-indices.md
parkera: It's just that the text of the proposal pre-dates the new index method (it used to be called successor).

proposals/0069-swift-mutability-for-foundation.md
parkera: This looks good.

proposals/0069-swift-mutability-for-foundation.md
parkera: Ok

lattner:
Thanks!

PR #362, opened 2016-06-10
Removing Where Clauses from For-In Loops

erica:
This proposal removes `where` clauses from `for-in` loops, where they are better expressed (and read) as guard conditions.

Swift Evolution Discussion: [\[Pitch\] Retiring `where` from for-in loops](http://thread.gmane.org/gmane.comp.lang.swift.evolution/20142)


lattner:
Thanks, I'll kick this off.

PR #365, opened 2016-06-14
Shorthand Argument Renaming

fblondiau:
Here is my final proposal, integrating suggestions.


lattner:
This seems subjectively a lot worse than the current $0 syntax that we have now (which is just MHO), but is empirically an inconsistent use of the # namespace (which is about compiler magic, not synthesized names).  I'm closing this in an attempt to optimize process, but if you'd like to continue discussing it on swift-evolution, then I'm happy to make the case there.

PR #366, opened 2016-06-14
Create NNNN-unboxing-anyindex.md

Haravikk:


Haravikk:
Could I get some clarification on why this is out of scope? Right now there's no point to AnyIndex having a public constructor without something like this, as there's not a lot you can do with an index once wrapped, except compare them, but that'd be a job for an `AnyComparable`. `AnyIndex` implies that a wrapped value should be usable as an index at some point, but that's currently not possible.

It should also be a pretty tiny change.

JustinJiaDev:
They are focusing on souce-breaking charges right now (till August). According to the core team, "out of scope" doesn't mean the proposal needs to wait until Swift 4. It is possible to add it in Swift 3.x.

Haravikk:
Thanks for the explanation!

It still feels like this could do with being in Swift 3.0 though rather than a later update, as the current situation means we've actually lost capabilities that `AnyIndex` used to give us, since it was still a fully functional index that could be advanced etc., whereas that's no longer the case (except for types within the stdlib itself).

DougGregor:

PR #369, opened 2016-06-18
Conditional Compilation Blocks proposal

erica:
# Enhancing the Platform Configuration Test Suite for Conditional Compilation Blocks

This proposal introduces configuration tests to differentiate platform conditions in conditional compilation blocks.

_This proposal was first discussed on-list in the [[Draft] Introducing Build Configuration Tests  for Platform Conditions](http://thread.gmane.org/gmane.comp.lang.swift.evolution/12140/focus=12267) thread and then updated and re-pitched in [[Draft] Enhancing the Platform Configuration Test  Suite for Conditional Compilation Blocks](http://thread.gmane.org/gmane.comp.lang.swift.evolution/20849)._

### DIFF
--- a/proposals/XXXX-platformconfig.md
+++ b/proposals/XXXX-platformconfig.md
@@ -1,0 +1,163 @@
+# Enhancing the Platform Configuration Test Suite for Conditional Compilation Blocks
+
+* Proposal: TBD
+* Author: [Erica Sadun](http://github.com/erica)
+* Status: TBD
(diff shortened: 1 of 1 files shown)


lattner:
I think that this is a fine proposal, but will need to wait for Swift 3.x where x > 0, since it is an additive proposal.  We'd like the community to stay as focused as possible on Swift 3.0 related work.  Thanks!

PR #371, opened 2016-06-19
Enum case stored properties proposal

kirsteins:
This proposal allows each enum to have stored properties that are shared among cases.

Swift Evolution Discussion: https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20160523/018875.html


lattner:
This is an additive feature that is out of scope for Swift 3.

PR #373, opened 2016-06-21
Proposal: Make non-escaping closures the default

tanadeau:


lattner:
Awesome, thank you for driving this @tanadeau !

PR #374, opened 2016-06-21
[WIP] Protocol oriented integers

moiseev:
This is **not** the final proposal yet. Still working on some details.


proposals/0000-updated-integers.md
```diff
@@ -0,0 +1,687 @@
...
+}
+```
+
+#### `SignedArithmetic`
+
+The `SignedArithmetic` protocol is for numbers that can be negated.
+
+```Swift
+public protocol SignedArithmetic : Arithmetic {
+  func negate() -> Self
```
dabrahams: Didn’t you want to change this to mutating and provide `negated()`?

proposals/0000-updated-integers.md
```diff
@@ -0,0 +1,687 @@
...
+operations are dispatched in `Arithmetic`, `==` and `<` operators for
+homogeneous comparisons are implemented as generic free functions invoking the
+`isEqual(to:)` and `isLess(than:)` protocol methods respectively.
+
+This protocol adds 3 new initializers to the parameterless one inherited from
+`Arithmetic`. These initializers allow to construct values of one type from
+instances of any other type conforming to `Integer`, using different
+strategies:
+
+  - Assume the value is representatble in `Self`
```
dabrahams: The above is still unclear.  What happens when the assumption is violated?

dabrahams:
Overall this is looking fantastic, Max.  Please accept my PR into your branch and address the [two](https://github.com/apple/swift-evolution/pull/374#discussion-diff-67971624) [items](https://github.com/apple/swift-evolution/pull/374#discussion-diff-67971670) above.  Then, as far as I’m concerned, it’s ready to go.

proposals/0000-updated-integers.md
moiseev: Yes. Only changed it in the prototype. Updated.

proposals/0000-updated-integers.md
moiseev: It is the implementation detail. We will trap, but should it be described here?

dabrahams: on Tue Jun 21 2016, Maxim Moiseev <notifications-AT-github.com> wrote:

>> +
>> +The `Integer` protocol is the basis for all the integer types provided by the
>> +standard library.
>> +
>> +The `isEqual(to:)` and `isLess(than:)` methods provide implementations for
>
>> +`Equatable` and `Comparable` protocol conformances. Similar to how arithmetic
>> +operations are dispatched in `Arithmetic`, `==` and `<` operators for
>> +homogeneous comparisons are implemented as generic free functions invoking the
>> +`isEqual(to:)` and `isLess(than:)` protocol methods respectively.
>> +
>> +This protocol adds 3 new initializers to the parameterless one inherited from
>> +`Arithmetic`. These initializers allow to construct values of one type from
>> +instances of any other type conforming to `Integer`, using different
>> +strategies:
>> +
>> +  - Assume the value is representatble in `Self`
>
> It is the implementation detail. We will trap, 

In that case you have to say it's a precondition on the caller.  The
point is that “assume the value is representable” might just mean
“truncate the bits” or something.

> but should it be described here?

-- 
Dave

lattner:
@moiseev Please email me (off list) when you're ready for this to be merged.  I'll take care of kicking off the review from there.

proposals/0000-updated-integers.md
```diff
@@ -0,0 +1,615 @@
...
+  /// indefinitely and then truncating to fit `Self`.
+  init<T : Integer>(extendingOrTruncating source: T)
+
+  /// Creates in instance of `Self` containing the closest representable
+  /// value of `source`.
+  init<T : Integer>(clamping source: T)
+
+  /// Returns n-th word, counting from the right, of the underlying
+  /// representation of `self`.
+  func nthWord(n: Int) -> UInt
```
brentdax: Is there a way to know how many words there are to get, other than `(bitWidth+sizeof(UInt)-1) / sizeof(UInt)`?

proposals/0000-updated-integers.md
```diff
@@ -0,0 +1,718 @@
...
+  static var isSigned: Bool { get }
+
+  var absoluteValue: AbsoluteValue { get }
+
+  func isEqual(to rhs: Self) -> Bool
+  func isLess(than rhs: Self) -> Bool
+
+  /// Creates an instance of `Self` that has the exact value of `source`,
+  /// returns `nil` otherwise.
+  init?<T : FloatingPoint>(exactly source: T)
```
natecook1000: Could we add a truncating `init<T : FloatingPoint>(_ source: T)` initializer as well?

moiseev: @natecook1000 There are multiple ways you can truncate a floating point into an integer. For now, I'd say, let's use `ceil`, `floor`, `trunc`, `round` etc. in combination with this initializer.
Having said that, I think designing a good Swifty API for floating point truncations might be a good proposal of it's own.

moiseev: /cc @stephentyrone, you probably have an opinion on this topic.

stephentyrone: It makes sense to have the truncating init as well; we already have it for all concrete stdlib integer and floating-point types as `init(_ source: ${FloatType})`, so this would just be putting it in the protocol.

For other rounding directions, rounding and then converting makes sense, though we'll need to add `ceil`, `floor`, etc to the `FloatingPoint` protocol to make that work smoothly (we should do that anyway).

natecook1000: Right—we need to still support `let x = Int(myFloatValue)`. It would be much nicer to have that conversion at the protocol level.

natecook1000:
Should this proposal describe how the new floating point system and extant numeric protocols fit in? As far as I can tell, `FloatingPoint` could inherit from `SignedArithmetic`. Do `SignedNumber` and `AbsoluteValuable` still exist after this has been implemented?

moiseev:
@natecook1000 it was designed with floating point in mind, it's just that floating point landed earlier. There are several places in the current implementation of floating point numbers where the code is commented out because there are no corresponding protocols yet. Answering your questions: yes, the `SignedArithmetic` should be adopted by floats and `SignedNumber` should disappear, as well as the `AbsoluteValuable` and the `~>` operator.

proposals/0000-updated-integers.md
moiseev: @brentdax Yes, although it is not included in the proposal: see the [prototype implementation](https://github.com/apple/swift/blob/a8c48d25efb1266604eab12617ae8eb595c53bb2/test/Prototypes/Integers.swift.gyb#L314). Please note, that the name in the prototype will most likely not survive.

Another thing worth mentioning: in order to be bignum friendly`nthWord` should support the case where its argument is greater than the word count. For example `(-1 as Int8).nthWord(42)` should return `UInt.max`, virtually extending the number indefinitely to the left.

proposals/0000-updated-integers.md
moiseev: Shame on me, I did not realize we already have a truncating initializer on integers. Will add it.

stephentyrone: They're hidden at the very end of FloatingPointTypes.swift.gyb.

PR #377, opened 2016-06-23
Revise SE-0077 according to the rationale

Anton3:
No ready to merge right now, because naming issues are not resolved.
Plus, opinions on chosen precedence group syntax are still incoming.

### DIFF
--- a/proposals/0077-operator-precedence.md
+++ b/proposals/0077-operator-precedence.md
@@ -5,6 +5,11 @@
 * Status: **Under revision** ([Rationale](#rationale))
 * Review manager: [Joe Groff](http://github.com/jckarter)
 
+**Revision history**
+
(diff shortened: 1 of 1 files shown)


lattner:
@Anton3 I think it is ready to merge now, the bike shedding on the names can be specifically cited as something that needs more discussion, which can be done in the review #2 thread.  Sound ok?

Anton3:
@lattner Ready to merge.

lattner:
Great, I'll schedule it for next week, thanks again!

lattner:
Oh, actually @jckarter took over review managerizing it.  Joe, do you want to schedule it, or should I?

jckarter:
I wanted to wait until @dabrahams and @gribozavr had a chance to review the proposed names.

jckarter:
Thanks for the revision @Anton3!

PR #381, opened 2016-06-25
Added literal syntax protocols proposal

anandabits:


anandabits:
@dabrahams Are there any changes you would like to see to this before we put it up for review?

dabrahams:
Well, I feel that the alternatives section legitimizes an existing misunderstanding of what these protocols mean.  What is cited as "the best explanation of the potential confusion” is complaining that the names are accurate and descriptive!  Also, it leaves out the two alternatives discussed that are actually accurate and *more* descriptive, `XXXLiteralExpressible` and `ExpressibleAsXXXLiteral`.  I might even prefer the latter pattern.  But it’s your proposal ;-)

anandabits:
Thanks for catching that @dabrahams.  It was written before you provided the explanation of why it was inaccurate.  I updated it to reflect the later discussion.  

It is my proposal, but I want it to be accurate and to propose what you feel is the best solution.  If you like `XXXLiteralExpressible` or `ExpressibleAsXXXLiteral` better than `XXXLiteral` I will make one of those the proposed solution rather than an alternative.  Let me know what you prefer.

DevAndArtist:
@anandabits you may consider to add `Syntax.Literal.*Protocol` version.

This is safe to use and actually does work.
```swift
func f<T: IntegerLiteralConvertible>() -> T {
    return T(integerLiteral: 43 as! T.IntegerLiteralType)
}
```

Proof: 

Literal magic

anandabits:
@DevAndArtist I don't want to include every alternative mentioned during discussion in the proposal (there have been too many to list them all).  But since you are specifically requesting this one I have added it.

dabrahams:
on Fri Jul 01 2016, Matthew Johnson <notifications-AT-github.com> wrote:

> Thanks for catching that @dabrahams.  It was written before you
> provided the explanation of why it was inaccurate.  I updated it to
> reflect the later discussion.
>
> It is my proposal, but I want it to be accurate and to propose what
> you feel is the best solution.  If you like `XXXLiteralExpressible` or
> `ExpressibleAsXXXLiteral` better than `XXXLiteral` I will make one of
> those the proposed solution rather than an alternative.  Let me know
> what you prefer.

Frankly I'm ambivalent.  The tricky `Syntax.` namespace prefix becomes
superflous and verbose IMO with `ExpressibleAsXXXLiteral`.  Do we want
to namespace these things or do we want the most English-sounding usage?

Should we go with `ExpressibleAs.XXXLiteral`? <0.25 wink>

-- 
Dave

anandabits:
I'm ambivalent as well.  Since we both feel that way, maybe we should just put it up for review as-is and you can work with the core team on making a final decision regarding the name we go with after the review completes.  How does that sound?

dabrahams:
Dang I hit the wrong button. Reopened.

anandabits:
@dabrahams you got it!

dabrahams:
OK, we’ve banished our ambivalence, and I’m sorry to say I’d like to invert the proposal.  We strongly believe that `ExpressibleAs*Literal`, without the `Syntax` namespace, is the right answer.  Would you mind doing the honors… again?

anandabits:
Lol, no problem.  Better to decide on a preferred name before the review starts! :)  

Awaiting further feedback...

DevAndArtist:
@anandabits If the `Syntax` namespace is the wrong way to go, I'd appreciate if you also could invert my suggestion from `Syntax.Literal.*Protocol` to `Literal.*Protocol`

anandabits:
@DevAndArtist so you want protocols with `*Protocol` naming scheme inside a `Literal` namespace?

DevAndArtist:
@anandabits that's correct. That was one of my very first suggestions in the evolution thread. `Literal.IntegerProtocol` etc.

lattner:
@dabrahams let me know when you think this is ready to merge and kick off.

anandabits:
@DevAndArtist your requested change has been made

dabrahams:
@lattner we’re good to go

lattner:
The proposal still has this in it, this doesn't seem right:

This namespace will initially be implemented using a case-less enum, but this detail may change in the future if submodules or namespaces are added to Swift. Swift does not currently allow protocols to be declared inside the scope of a type. In order to work around this limitation the protocols themselves will be declared using underscore-prefixed names internal to the standard library. Typealiases inside the Syntax enum will declare the names intended to be visible to user code.

lattner:
Should I drop that paragraph?

anandabits:
@lattner good catch.  Not sure how I missed that when we changed it.  Feel free to drop it, or let me know if you want me to do it.

lattner:
Thanks, I removed it in 4fc7d46

PR #385, opened 2016-06-28
Adding optional Array subscripts to 'commonly proposed' list

austinzheng:
@dabrahams


DougGregor:
Looks great, thanks!

brentdax:
Sorry to reply late, but I think it might be a good idea to say that adding *separate* Optional-returning subscripts has not been seriously proposed or evaluated yet, but is out of scope for the current phase of the Swift evolution process.

dabrahams:
@brentdax, if there's something you want to add, please submit a PR for the change so we can evaluate your exact proposed wording.

PR #389, opened 2016-06-29
Rewriting SE-0108 based on Doug Gregor's feedback

austinzheng:
@DougGregor @lattner 

Let me know if you'd like any last-minute corrections or amendments before this goes in!


DougGregor:
This looks great, thank you!

PR #393, opened 2016-06-29
Change copy based on Chris' feedback

austinzheng:
@lattner


marcusrossel:
I really like this proposal! I've had problem's with this several times, and never found it to be consistent.

lattner:
Thanks!

PR #395, opened 2016-06-30
Move SE-0067 to "Implemented"

ultramiraculous:
I think this was completed here? -> https://github.com/apple/swift/pull/2453


stephentyrone:
Substantially completed, yes.  There are a few features outstanding, which either need some Builtin support or depend on SE-0104 for implementation.

lattner:
@stephentyrone I know you're working on the last details of that, feel free to merge this when it makes sense to you.

lattner:
This has conflicts, but I'm going to apply the spirit of the patch.  The remaining few members can be treated as bugs to be fixed.

PR #404, opened 2016-07-01
Proposal: Union Type

frogcjn:
```swift
var stringOrURL: String | URL = "https://www.apple.com"
```
Please see https://github.com/frogcjn/swift-evolution/blob/master/proposals/xxxx-union-type.md

### DIFF
--- a/proposals/xxxx-union-type.md
+++ b/proposals/xxxx-union-type.md
@@ -163,24 +163,25 @@ There are serveral advance points:
 
 * It will be easy to compare with value of original type.
 ```swift
-    let union: A|B|C
+    let union: A|B|C = someA
     let otherA: A
 
     // If union is not type A, then return false;
     // If union is type A, then compare it with otherA!!😊
-    return union == otherA
+    print(union == otherA)
 
 ```
 instead of unwrap enum cases and compare.
 ```swift
-    let u: UnionOfABC
+    let u: UnionOfABC = UnionOfABC._A(someA)
     let otherA: A
 
-    if(let case ._A(value) = u) {
-        return value == otherA
+    if case let ._A(value) = u {
+        print(value == otherA)
     } else {
-        return false
+        print(false)
     }
+
 ```
 
 * Original types and union types can have a rational relationship between each other.
--- a/proposals/xxxx-union-type.md
+++ b/proposals/xxxx-union-type.md
@@ -16,7 +16,7 @@ var stringOrURL: String | URL = "https://www.apple.com"
 ```
 
 Swift-evolution thread: [Recently](http://news.gmane.org/gmane.comp.lang.swift.evolution/023056)
- [old](http://article.gmane.org/gmane.comp.lang.swift.evolution/7162/)
+ [old](http://article.gmane.org/gmane.comp.lang.swift.evolution/17193/)
 
 ## Motivation
 
--- a/proposals/xxxx-union-type.md
+++ b/proposals/xxxx-union-type.md
@@ -1,0 +1,193 @@
+# Feature name
+
+* Proposal: [SE-NNNN](NNNN-union-type.md)
+* Author: [Cao, Jiannan](https://github.com/frogcjn)
+* Status: **Awaiting review**
+* Review manager: TBD
+
+## Introduction
+
+Add union type grammar, represents the type maybe one of other types.
+
+```swift
+
+var stringOrURL: String | URL = "https://www.apple.com"
+
+```
+
+Swift-evolution thread: [Discussion thread topic for that proposal](http://news.gmane.org/gmane.comp.lang.swift.evolution)
+
+## Motivation
+
+There are many reasons to have this important feature, the obvious one is that developer will write less type declaration code in Swift, for example:
+
+For example, there is three class type A, B and C:
+
+```swift
+
+
+class A {
+    var commonProperty: String?
+    var propertyInA: String?
+}
+
+
+class B {
+    var commonProperty: String?
+    var propertyInB: String?
+}
+
+class C {
+    var commonProperty: String?
+    var propertyInC: String?
+}
+
+```
+
+Originaly, if we want to represent type of a varialbe maybe A, B or C, we should anounce a new type:
+
+```swift
+
+protocol Common {
+    var commonProperty: String? {get}
+}
+
+extension A : Common {}
+extension B : Common {}
+extension C : Common {}
+
+func input(value: Common) {
+    print(value.commonProperty)
+    switch value {
+    case let value as A:
+        // value is type A
+        print(value.propertyInA)
+    case let value as B:
+        // value is type B
+        print(value.propertyInB)
+    case let value as C:
+        // value is type C
+        print(value.propertyInC)
+    default:
+        // we don't want any Common type other then A, B, C, so we trigger a fatalError
+        fatalError("No other type than A, B, C")
+        // since we can't write this kind of requirement with protocol
+    }
+}
+```
+
+Or we can make an union type of A, B, C to limit the type.
+
+```swift
+
+enum UnionOfABC {
+    case _A(value: A)
+    case _B(value: B)
+    case _C(value: C)
+    var value: Common {
+        switch(self) {
+        case let _A(value): return value
+        case let _B(value): return value
+        case let _C(value): return value
+        }
+    }
+}
+
+func input(u: UnionOfABC) {
+    print(u.value.commonProperty)
+    switch u {
+    case let ._A(a):
+        print(a.propertyInA)
+    case let ._B(b):
+        print(b.propertyInB)
+    case let ._C(b):
+        print(b.propertyInC)
+    }
+}
+```
+
+## Proposed solution
+
+Now, if we using the union type feature, we can declare type conveniently,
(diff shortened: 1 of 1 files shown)


cezheng:
Seriously, are you a Java programmer here making a proposal to swift-evolution?

frogcjn:
@cezhang, Java has no union type like this.

cezheng:
I mean your sample code

```swift
let a = new A()
```

cezheng:
Well I personally don't like this proposal, since it is adding complexity to the language while not really simplifying the use cases you mentions.

frogcjn:
@czheng, I have some other people support this proposal.
I think union type is a rational grammar that every programming language should learn about. This feature should finally replace the generic enum solution for optional feature.

by the way.
```
let a = new A()
```
I wrote this kind of code in my Swift project. I would like to use type inference in Swift.
This is so swift way, Why Java? Since this topic is not about my proposal you can discuss this topic other place.

cezheng:
ok, I'm convinced that you have never really written any swift, and still writing a swift proposal
image

cezheng:
What other languages have is not the reason why swift should have the same syntax. Python also claims that switch is not necessary since if else can just do the same things, then why don't you propose to remove switch from swift?

I think for the same reason they remove c-style for and ++, -- from swift, we shouldn't add this to swift since it is adding unnecessary complexity to the language itself.

frogcjn:
@cezheng welcome to discuss on swift-evolution email list.

DevAndArtist:
@frogcjn 

1. You never read any of the swift-evolution rules or threads? Your proposal had no thread before PR.
2. Swift does not support infix operator on types at declaration. The first step in this direction will be [SE-0095](https://github.com/apple/swift-evolution/blob/master/proposals/0095-any-as-existential.md), but thats it for now. 
3. An `|` operator is not welcome from the core team point of view (at least for now).
4. The proposal to remove `.self` magic from types is deferred.
5. This proposal is way out of scope for Swift 3!

This proposal will probably be closed anytime soon, @lattner, @dabrahams?

frogcjn:
@DevAndArtist I have thread before PR.

DevAndArtist:
@frogcjn Actually no haven't, your PR is from 10:45 (my time) where the thread was opened at 11:06. But the point I had is that the community literally had no time to discuss the proposal. You'd know that this is out of scope for Swift 3 and not yet possible at all.

Here is some bikeshedding:

```swift
enum OneOf<T, U> {
	
	case lhs(T)
	case rhs(U)
	
	init(value: T) {
		
		self = .lhs(value)
	}
	
	init(value: U) {
		
		self = .rhs(value)
	}
}

func |<T, U>(_ lhs: T.Type, _ rhs: U.Type) -> OneOf<T, U>.Type {
	
	return OneOf<T, U>.self
}

let test1 = (String.self | Int.self).init(value: "test")
let test2 = (String.self | Int.self).init(value: 42)
```

frogcjn:
@DevAndArtist

I have thread before PR. You can search. I have discussed this porposal way back to February. I have read swift-evolution rules. You can search my thread.

frogcjn:
@DevAndArtist I have wrote details in proposal about why Enum generic is not enough.

frogcjn:
Please see http://article.gmane.org/gmane.comp.lang.swift.evolution/17193/

cezheng:
I just saw more people opposing it than supporting it. I think this is pretty much over.

frogcjn:
@cezheng, this is just a proposal, take it easy. I believe if this feature implemented in Swift will help developer to write their code and make the logic in compiler and developer's code clear.

DevAndArtist:
I'd rather propose in the future for some kind of a variadic generic (where `...T` does not mean the same as `T...`) and a variadic cased enum.

This would solve your problem with `default` on switches.

Bikeshedding:

```swift
enum OneOf<...T> {
	
	...case $#(T)

	init(value: ...T) {
		
		self = .$(value)
	}
}
```

frogcjn:
@DevAndArtist, I like this feature. I think any generic programming language should have this feature.
But I still think using generic to wrap value to implement the union or option feature is not a good solution,

cezheng:
It's you that should take it easy. I'm just showing my own opinion that this proposal is adding complexity to the language while increasing the potential for people to write bad code.

frogcjn:
@cezheng you can give examples and discussion on email list, why the union type is bad. Thanks.

DevAndArtist:
@frogcjn whats the problem with `Optional`? It's a neat feature we have in swift. :) 

The way you're trying to solve it (with the `|` operator) is to complex, because a true union type would merge the universe of same types like `String | String | Int` into `String | Int`. I believe no one would do that (`String | String`) anyway, that's why something like `enum OneOf<...T>` would solve the problem with less complexity. Thats why I personally think there is no need for a union type. ;)

DevAndArtist:
> @cezheng you can give examples and discussion on email list, why the union type is bad. Thanks.

It's not about the type being bad, its about the usage of a such complex type. As you showed it yourself, you'd use it for a simple task. So I ask you why is it necessary to add complexity to solve simple problems that might find a neat solution in the future with less complexity?

I kinda think that I've seen `...T` somewhere on the Swift repository.

frogcjn:
using generic to solve the either/optional/union problem, is not logical and complex.
I have wrote in my proposal.

There are several other reasons:

If you use generic to solve the either/union problem, developer will wrote Either declaration, and library like RxSwift and Alamofire has announced their own Result either type. And their either type have different design but sample ability. The Result either type should be unified by using union.

union type is flatten. generic wrap solution is not flatten.

For example:


typealias ABC  = A | B | C 

typealias ABCD = ABC | D

we just use an existed type ABC to construct ABCD

But how about generic wrap?

frogcjn:
If you construct ABCD with ABC using generic solution,  you'll get two layer, and when you do  `switch` you should write two layer of switch. And you should unwrap two times of that monster type.

DevAndArtist:
> For example:
>
> typealias ABC  = A | B | C
> 
> typealias ABCD = ABC | D
> 
> we just use an existed type ABC to construct ABCD
> 
> But how about generic wrap?

Bikeshedding:

```swift
// we generate the boundary with `A | B` or directly OneOf<A, B>
enum OneOf<...T> {

	...case $#(T)

	// Bikeshedding variadic enum casses:
	// we might need an index to set the value?
	init(index: Int, value: T) {
		
		self = .$index(value)
	}
}

/// Usage:
/// A | B | C == OneOf<A, B, C>
func |<T, U>(_: T.Type, _: U.Type) -> OneOf<T, U>.Type {
	
	// I believe the usage of `type` like this was prposed by Joe Groff
	// I also use the proposal to remove `.self` magic here
	return OneOf<T, U>
}

// Here is how to merge OneOf into a single dimension
func |<...T, U>(_: OneOf<...T>.Type, _: U.Type) -> OneOf<...T, U>.Type {
	
	// Copy and merge types into the new `OneOf` type
	return OneOf<...T, U>
}

func |<T, ...U>(_: T.Type, _: OneOf<...U>.Type) -> OneOf<T, ...U>.Type {
	
	// Copy and merge types into the new `OneOf` type
	return OneOf<T, ...U>
}

func |<...T, ...U>(_: OneOf<...T>.Type, _: OneOf<...U>.Type) -> OneOf<...T, ...U>.Type {
	
	// Copy and merge types into the new `OneOf` type
	return OneOf<...T, ...U>
}
```

Your example will become:

```swift
typealias ABC = A | B | C // or OneOf<A, B, C>
typealias ABCD = ABC | D // merging lhs OneOf with D to OneOf<A, B, C, D>
```

Mission accomplished.

frogcjn:
Union version more complex or generic version?
I think union version is more normal.

frogcjn:
I can't compile your code. I'm really not sure your code will pass the complex generic check 🙃

DevAndArtist:
I believe the complexity hides in merging the types into a single universe (correct me if I'm wrong), because for true union `String | Int | String` == `String | Int` == `Int | String`.

My code won't compile, I've never said it does. This is a bikeshedding example how your problem might be solved with less complexity by giving up merging the type universe.

True the **real** union type cannot be made of an `enum` not even a variadic enum will to the trick, because there is no way to know how huge the universe of `...T` will become when there is something like `A | A | B`. Logically you'd assume you will need only two cases but how would you know that in a type safe manner at compile time?

I'd rather invest my strength into variadic generics, where we also would allow something like this `<...T, ...U>` or `<...T, U>` and the creation of variadic enums, if we'd get variadic structs like the generic manifesto showed us.

Okay I'm done here discussing my vision. I'm out. ;)

DevAndArtist:
Just one simple thing to add:

* If `...T` equals 10 then `<...T, U>` = 10 + 1 = 11 types
* If `...U` equals 17 then `<T, ...U>` = 1 + 17  = 18 types
* If `...T` equals 20 and `...U`equals 22 then `<...T, ...U>` = 20 + 22 = 42 types

frogcjn:
what if OneOf<A, A>?

duplicate variable compile warning?

DougGregor:
Please use the swift-evolution mailing list for discussions.

lattner:
This is clearly out of scope for Swift 3, and has been discussed many times on the list.  Please wait until Swift 3 is done before tackling new feature areas.

PR #407, opened 2016-07-01
Import Objective-C `id` as Swift `Any`

jckarter:


lattner:
Wow, this proposal is really well done @jckarter.  I'm going to schedule it to start on tuesday.  Let me know if that won't work for you.

jckarter:
Works for me, thanks!

PR #410, opened 2016-07-05
[WIP] Revised SE-0107: UnsafeRawPointer API, initialization method names.

atrick:
Naming improvements suggested by Brent Royal-Gordon.

### DIFF
--- a/proposals/0107-unsaferawpointer.md
+++ b/proposals/0107-unsaferawpointer.md
@@ -229,7 +229,7 @@ The `load` and `storeRaw` operations are assymetric. `load` reads raw
 bytes but properly constructs a new value of type `T` with its own
 lifetime. Any copied references will be retained. In contrast,
 `storeRaw` only operates on a value's raw bytes, writing them into
-untyped memory. The in-memory copy will not be initialized and any
+untyped memory. The in-memory copy will not be constructed and any
 previously initialized value in this memory will not be deinitialized
 (it cannot be because its type is unknown). Consequently, `storeRaw`
 should only be performed on trivial types.
@@ -1455,7 +1455,10 @@ helper, no unsafe casting is necessary in the previous example:
 
 Most of the API was already presented above. For the sake of having it
 in one place, a list of the expected `UnsafeMutableRawPointer` members
-is shown below:
+is shown below.
+
+For full doc comments, see the [github revision]
+(https://github.com/atrick/swift/blob/80a1f2624ee4aae7d51253d1b2541826b01d4118/stdlib/public/core/UnsafeRawPointer.swift.gyb).
 
 ```swift
 struct UnsafeMutableRawPointer : Strideable, Hashable, _Pointer {
--- a/proposals/0107-unsaferawpointer.md
+++ b/proposals/0107-unsaferawpointer.md
@@ -1583,15 +1583,15 @@ extension UnsafePointer<Pointee> {
 }
 ```
 
-The following `UnsafeMutablePointer` members are renamed:
+The following `UnsafeMutablePointer` members are modified:
 
 ```swift
 extension UnsafeMutablePointer<Pointee> {
-  // Naming conventions.
+  // Naming conventions changed here:
   static func allocate(capacity: Int) -> UnsafeMutableRawPointer
   func deallocate(capacity: Int)
 
-  // Naming conventions.
+  // Naming conventions changed here:
   func assign(from source: UnsafePointer<Pointee>, forwardToCount: Int)
   func assign(from source: UnsafePointer<Pointee>, backwardToCount: Int)
   func moveInitialize(
@@ -1603,7 +1603,7 @@ extension UnsafeMutablePointer<Pointee> {
   func initialize<C : Collection>(from source: C)
   func moveAssign(from source: ${Self}, count: Int) {
 
-  // Now returns a raw pointer.
+  // This now returns a raw pointer.
   func deinitialize(count: Int = 1) -> UnsafeMutableRawPointer
 ```
 
--- a/proposals/0107-unsaferawpointer.md
+++ b/proposals/0107-unsaferawpointer.md
@@ -1477,34 +1477,36 @@ struct UnsafeMutableRawPointer : Strideable, Hashable, _Pointer {
   func bindMemory<T>(to: T.Type, capacity: Int) -> UnsafeMutablePointer<T>
   func assumingMemoryBound<T>(to: T.Type) -> UnsafeMutablePointer<T>
 
-  func initialize<T>(_: T.Type, with: T, count: Int = 1)
+  func initialize<T>(_: T.Type, to: T, count: Int = 1)
     -> UnsafeMutablePointer<T>
-  func initialize<T>(toContiguous: T.Type, atIndex: Int, with: T)
+  func initialize<T>(contiguous: T.Type, at: Int, to: T)
     -> UnsafeMutablePointer<T>
 
-  func initialize<T>(from: UnsafePointer<T>, count: Int)
+  func initialize<T>(from: UnsafePointer<T>, forwardToCount: Int)
     -> UnsafeMutablePointer<T>
-  func initializeBackward<T>(from: UnsafePointer<T>, count: Int)
+  func initialize<T>(from: UnsafePointer<T>, backwardToCount: Int)
     -> UnsafeMutablePointer<T>
 
   // The `move` APIs deinitialize the memory at `from`.
-  func moveInitialize<T>(from: UnsafePointer<T>, count: Int)
+  func moveInitialize<T>(from: UnsafePointer<T>, forwardToCount: Int)
     -> UnsafeMutablePointer<T>
-  func moveInitializeBackward<T>(from: UnsafePointer<T>, count: Int)
+  func moveInitialize<T>(from: UnsafePointer<T>, backwardToCount: Int)
     -> UnsafeMutablePointer<T>
 
-  func load<T>(_: T.Type) -> T
-  func load<T>(_: T.Type, atByteOffset: Int) -> T
-  func load<T>(fromContiguous: T.Type, atIndex: Int) -> T
+  func load<T>(as: T.Type) -> T
+  func load<T>(fromByteOffset: Int, as: T.Type) -> T
+  func load<T>(fromContiguous: T.Type, at: Int) -> T
 
   // storeRaw performs bytewise writes, but proper alignment for `T` is still
   // required.
   // T must be a trivial type.
-  func storeRaw<T>(_: T.Type, with: T)
-  func storeRaw<T>(toContiguous: T.Type, atIndex: Int, with: T)
-  func storeRaw<T>(contiguous: T.Type, from: UnsafeRawPointer, count: Int)
-  func storeRawBackward<T>(
-    contiguous: T.Type, from: UnsafeRawPointer, count: Int)
+  func storeRaw<T>(_: T, as: T.Type)
+  func storeRaw<T>(_: T, toContiguous: T.Type, at: Int)
+  func storeRaw<T>(_: T, toByteOffset: Int)
+  func storeRaw<T>(
+    contiguous: T.Type, from: UnsafeRawPointer, forwardToCount: Int)
+  func storeRaw<T>(
+    contiguous: T.Type, from: UnsafeRawPointer, backwardToCount: Int)
 
   func distance(to: UnsafeRawPointer) -> Int
   func advanced(by: Int) -> UnsafeRawPointer
@@ -1531,9 +1533,9 @@ struct UnsafeRawPointer : Strideable, Hashable, _Pointer {
   func bindMemory<T>(to: T.Type, capacity: Int) -> UnsafePointer<T>
   func assumingMemoryBound<T>(to: T.Type) -> UnsafePointer<T>
 
-  func load<T>(_: T.Type) -> T
-  func load<T>(_: T.Type, atByteOffset: Int) -> T
-  func load<T>(fromContiguous: T.Type, atIndex: Int) -> T
+  func load<T>(as: T.Type) -> T
+  func load<T>(fromByteOffset: Int, as: T.Type) -> T
+  func load<T>(fromContiguous: T.Type, at: Int) -> T
 
   func distance(to: UnsafeRawPointer) -> Int
   func advanced(by: Int) -> UnsafeRawPointer
@@ -1587,12 +1589,14 @@ extension UnsafeMutablePointer<Pointee> {
   func deallocate(capacity: Int)
 
   // Naming conventions.
-  func assign(from source: UnsafePointer<Pointee>, count: Int)
-  func assignBackward(from source: UnsafePointer<Pointee>, count: Int)
-  func moveInitialize(from source: UnsafeMutablePointer<Pointee>, count: Int)
-  func moveInitializeBackward(from source: UnsafeMutablePointer<Pointee>,
-                              count: Int)
-  func initialize(from source: UnsafeMutablePointer<Pointee>, count: Int)
+  func assign(from source: UnsafePointer<Pointee>, forwardToCount: Int)
+  func assign(from source: UnsafePointer<Pointee>, backwardToCount: Int)
+  func moveInitialize(
+    from source: UnsafeMutablePointer<Pointee>, forwardToCount: Int)
+  func moveInitialize(
+    from source: UnsafeMutablePointer<Pointee>, backwardToCount: Int)
+  func initialize(
+    from source: UnsafeMutablePointer<Pointee>, forwardToCount: Int)
   func initialize<C : Collection>(from source: C)
   func moveAssign(from source: ${Self}, count: Int) {
 


proposals/0107-unsaferawpointer.md
```diff
@@ -1477,34 +1477,36 @@ struct UnsafeMutableRawPointer : Strideable, Hashable, _Pointer {
   func bindMemory<T>(to: T.Type, capacity: Int) -> UnsafeMutablePointer<T>
   func assumingMemoryBound<T>(to: T.Type) -> UnsafeMutablePointer<T>
 
-  func initialize<T>(_: T.Type, with: T, count: Int = 1)
+  func initialize<T>(_: T.Type, to: T, count: Int = 1)
     -> UnsafeMutablePointer<T>
-  func initialize<T>(toContiguous: T.Type, atIndex: Int, with: T)
+  func initialize<T>(contiguous: T.Type, at: Int, to: T)
```
dabrahams: This one looks really easy to misuse.  Is the `at` parameter (BTW, you **need** parameter names distinct from the argument labels!) measured in bytes or in `T`'s?

dabrahams: Both of the above don’t look like they read well:

```swift
address.initialize(String.self, to: “hello")
address.initialize(contiguous: String.self, at: position, to: “hello”)
```
In the first one, it makes `String.self` read as the direct object, but we are not initializing the type; we’re initializing memory.

`contiguous` is very confusing in the latter call because it implies multiple elements, yet this initializes only a single element.  What’s it supposed to mean?

proposals/0107-unsaferawpointer.md
```diff
@@ -1477,34 +1477,36 @@ struct UnsafeMutableRawPointer : Strideable, Hashable, _Pointer {
...
-  func initializeBackward<T>(from: UnsafePointer<T>, count: Int)
+  func initialize<T>(from: UnsafePointer<T>, backwardToCount: Int)
     -> UnsafeMutablePointer<T>
 
   // The `move` APIs deinitialize the memory at `from`.
-  func moveInitialize<T>(from: UnsafePointer<T>, count: Int)
+  func moveInitialize<T>(from: UnsafePointer<T>, forwardToCount: Int)
     -> UnsafeMutablePointer<T>
-  func moveInitializeBackward<T>(from: UnsafePointer<T>, count: Int)
+  func moveInitialize<T>(from: UnsafePointer<T>, backwardToCount: Int)
```
dabrahams: Didn’t we agree to dump separate forward/backward APIs?

proposals/0107-unsaferawpointer.md
```diff
@@ -1477,34 +1477,36 @@ struct UnsafeMutableRawPointer : Strideable, Hashable, _Pointer {
...
     -> UnsafeMutablePointer<T>
 
-  func load<T>(_: T.Type) -> T
-  func load<T>(_: T.Type, atByteOffset: Int) -> T
-  func load<T>(fromContiguous: T.Type, atIndex: Int) -> T
+  func load<T>(as: T.Type) -> T
```
dabrahams: I don’t think we want a label here.  Isn’t it reasonable to say

```swift
address.load(String.self)
```
?

proposals/0107-unsaferawpointer.md
```diff
@@ -1477,34 +1477,36 @@ struct UnsafeMutableRawPointer : Strideable, Hashable, _Pointer {
...
+  func load<T>(fromByteOffset: Int, as: T.Type) -> T
```
dabrahams: I like the old one better.  The change from `at` to `from` is nice, though.

proposals/0107-unsaferawpointer.md
```diff
@@ -1477,34 +1477,36 @@ struct UnsafeMutableRawPointer : Strideable, Hashable, _Pointer {
...
+  func load<T>(fromContiguous: T.Type, at: Int) -> T
```
dabrahams: I don’t know what contiguous means in this context.  BTW, doc comments would help these APIs.  You shouldn’t really propose APIs without them.

proposals/0107-unsaferawpointer.md
```diff
@@ -1477,34 +1477,36 @@ struct UnsafeMutableRawPointer : Strideable, Hashable, _Pointer {
...
 
   // storeRaw performs bytewise writes, but proper alignment for `T` is still
   // required.
   // T must be a trivial type.
-  func storeRaw<T>(_: T.Type, with: T)
-  func storeRaw<T>(toContiguous: T.Type, atIndex: Int, with: T)
-  func storeRaw<T>(contiguous: T.Type, from: UnsafeRawPointer, count: Int)
-  func storeRawBackward<T>(
-    contiguous: T.Type, from: UnsafeRawPointer, count: Int)
+  func storeRaw<T>(_: T, as: T.Type)
```
dabrahams: Why are we putting “raw” in the name?  You’re already in a “raw” pointer.  Presumably all the operations are somehow raw.

proposals/0107-unsaferawpointer.md
```diff
@@ -1587,12 +1589,14 @@ extension UnsafeMutablePointer<Pointee> {
   func deallocate(capacity: Int)
 
   // Naming conventions.
```
dabrahams: What does that comment mean?

proposals/0107-unsaferawpointer.md
brentdax: I read it as "initialize contiguous `String` at `position` to `"hello"`"—that is, assume the pointer points to contiguous `String`s, and initialize the `String` at `position` to `"hello"`.

If you don't think that reads well, I wonder if we can combine both the `at` and `count` variants into one call:

```swift
func initialize<T>(_: T.Type, at: Int = 0, count: Int = 1, to value: T) -> UnsafeMutablePointer<T>

address.initialize(String.self, to: "hello")
address.initialize(String.self, at: 3, to: "hello")
address.initialize(String.self, count: 3, to: "hello")
address.initialize(String.self, at: 3, count: 3, to: "hello") 
```

If the bare type doesn't read well, it could have an `as` label.

proposals/0107-unsaferawpointer.md
brentdax: I'm not the one who chose this name, but I read it as a reminder that this API won't do reference counting or other operations we would normally associate with storing a value to memory.

proposals/0107-unsaferawpointer.md
atrick: I think the "as" label makes sense when working with a raw pointer.

I like this:

func initialize<T>(as: T.Type, at: Int = 0, count: Int = 1, to value: T) -> UnsafeMutablePointer<T>

address.initialize(as: String.self, to: "hello")
address.initialize(as: String.self, at: 3, to: "hello")
address.initialize(as: String.self, count: 3, to: "hello")
address.initialize(as: String.self, at: 3, count: 3, to: "hello")

proposals/0107-unsaferawpointer.md
atrick: Yes. Let's dump forward/backward APIs.

proposals/0107-unsaferawpointer.md
atrick: address.load(String.self) is readable, but the "as" label is a nice cue that this operation reads from raw memory. Memory could in theory hold a different type (that happens to by layout compatible with String :)

proposals/0107-unsaferawpointer.md
atrick: If we use the "as" label everywhere, which I like for raw pointers, then this is more readable:

address.load(fromByteOffset: offset, as: Int.self)

proposals/0107-unsaferawpointer.md
dabrahams: on Wed Jul 06 2016, Andrew Trick <notifications-AT-github.com> wrote:

> I like this:
>
> func initialize<T>(as: T.Type, at: Int = 0, count: Int = 1, to value: T) -> UnsafeMutablePointer<T>
>
> address.initialize(as: String.self, to: "hello")

Changing the order would be more idiomatic, since we already have 

         "hello" as String

  address.initialize(to: "hello", as: String.self)

> 
> address.initialize(as: String.self, at: 3, to: "hello")

That's non-grammatical, and I still don't know whether the `at:`
parameter is measured in bytes or in `MemoryLayout<String>.stride`s.
Which is it?  I can't even make a good recommendation until I know your
intentention.

> address.initialize(as: String.self, count: 3, to: "hello")

I would prefer `count:` to go last and be defaulted to 1

> address.initialize(as: String.self, at: 3, count: 3, to: "hello")

Are all these combinations really needed?  I'm just not sure it's worth
the API complexity.

dabrahams: One other point: saying `initializeMemory` would probably be clearer and more consistent.  We are not initializing the pointer.

proposals/0107-unsaferawpointer.md
dabrahams: Okay, I get your point… but if you really want to make that clear, you might want to go with

```swift
address.load(interpretingMemoryAs: String.self)
```

proposals/0107-unsaferawpointer.md
dabrahams: Agreed.

proposals/0107-unsaferawpointer.md
dabrahams: PING!

proposals/0107-unsaferawpointer.md
dabrahams: That’s already a precondition of the function.  If anything I would encode *that* in the name, e.g. `store(x, asTrivial: Int.self)`

proposals/0107-unsaferawpointer.md
atrick: I'll try to always include a link to the [full doc comments](https://github.com/atrick/swift/blob/80a1f2624ee4aae7d51253d1b2541826b01d4118/stdlib/public/core/UnsafeRawPointer.swift.gyb). This PR probably should have been done off the branch rather than the proposal itself.

  /// Load a single value of type `T` at the specified `index` from
  /// this memory as if it contains at least `index` + 1 contiguous
  /// values of type `T`.

The `load(fromContiguous:at:) API is a convenience to avoid typing fromByteOffset: strideof(T.self) * index.

The original request was for a load(fromArrayOf: T.Type, at: Int) API. But I don't want to use the term "array" so I've been using "contiguous"

proposals/0107-unsaferawpointer.md
atrick: Brent is correct.

TL;DR: Storing into raw memory is significantly more raw than loading from raw memory. `storeRaw` is basically our exposed `memcpy` API.

From the proposal:

The `load` and `storeRaw` operations are assymetric. `load` reads raw
bytes but properly constructs a new value of type `T` with its own
lifetime. Any copied references will be retained. In contrast,
`storeRaw` only operates on a value's raw bytes, writing them into
untyped memory. The in-memory copy will not be constructed and any
previously initialized value in this memory will not be deinitialized
(it cannot be because its type is unknown). Consequently, `storeRaw`
should only be performed on trivial types.

Assigning memory to a nontrivial type is done by binding the type:

```swift
rawPtr.bindMemory(to: PreviousType.self, capacity: 1).deinitialize(count: 1)
rawPtr.initialize(NewType.self, with: NewType())
```

proposals/0107-unsaferawpointer.md
atrick: This is a list of methods that only changed with respect to their naming conventions. Next time I'll send a PR on the branch that defines UnsafeRawPointer for more thorough review.

proposals/0107-unsaferawpointer.md
atrick: ...and I think Dave's point is that if the method is only valid on trivial types, the user doesn't really need to worry about all that.

I'm fine with `store(x, asTrivial: Int.self)`

I just don't want anyone to think that `load` & `store` are symmetric.

proposals/0107-unsaferawpointer.md
dabrahams: I’m not convinced this API is a good idea.  Is there really a strong enough use-case for it?  

My theory: if you are dealing with actual contiguous `T`’s you should be using typed access.  Otherwise, your `T` is unlikely to line up on a multiple of `MemoryLayout<T>.stride` anyhow.  What’s left over?  Are the uses important enough to complicate the API for?

atrick: I'm working on a new PR, but meanwhile here's an update:

I'll remove load(fromContiguous:at:) and storeRaw(_:toContiguous:at:) because they read poorly, they are additive, and we don't have compelling use cases yet.

Now, for single-value reads and writes we have:

func load<T>(as: T.Type) -> T
func load<T>(fromByteOffset: Int, as: T.Type) -> T

func store<T>(_: T, asTrivial: T.Type)
func store<T>(_: T, toByteOffset: Int, asTrivial: T.Type)

It's necessary to provide these APIs because they are the only way to
access the bytes of untyped memory. They don't really satisfy common
anticipated use cases. I think the "ByteOffset" convenience variants
are reasonable if though they are additive because they make it even
more clear that the function is accessing raw bytes. For manual layout, I
would slightly prefer that people use the "ByteOffset" APIs than bytewise pointer arithmetic.

atrick: Sorry for copy-pasting quickly. This is the proposed API for single-value reads and writes of raw bytes:

```swift
func load<T>(as: T.Type) -> T
func load<T>(fromByteOffset: Int, as: T.Type) -> T

func store<T>(_: T, asTrivial: T.Type)
func store<T>(_: T, toByteOffset: Int, asTrivial: T.Type)
```

proposals/0107-unsaferawpointer.md
atrick: I would like to distinguish initialization via raw pointer from initialization via typed pointer primarily because raw pointer initialization binds memory to a type.

Option 1:
--------

```swift
// n copies of a single value
rawPtr.initializeMemory(as: Int.self, count: n, to: x)
// read n copies out of memory at `ptrToInt`
rawPtr.initializeMemory(as: Int.self, from: ptrToInt, count: n)
```

Option 2:
--------

```swift
// n copies of a single value
rawPtr.initializeAndBind(as: Int.self, count: n, to: x)
// read n copies out of memory at `ptrToInt`
rawPtr.initializeAndBind(as: Int.self, from: ptrToInt, count: n)
```

proposals/0107-unsaferawpointer.md
```diff
@@ -1477,34 +1477,36 @@ struct UnsafeMutableRawPointer : Strideable, Hashable, _Pointer {
...
-  func storeRaw<T>(contiguous: T.Type, from: UnsafeRawPointer, count: Int)
-  func storeRawBackward<T>(
-    contiguous: T.Type, from: UnsafeRawPointer, count: Int)
+  func storeRaw<T>(_: T, as: T.Type)
+  func storeRaw<T>(_: T, toContiguous: T.Type, at: Int)
+  func storeRaw<T>(_: T, toByteOffset: Int)
+  func storeRaw<T>(
+    contiguous: T.Type, from: UnsafeRawPointer, forwardToCount: Int)
+  func storeRaw<T>(
+    contiguous: T.Type, from: UnsafeRawPointer, backwardToCount: Int)
```
atrick: I realized I have not gotten enough feedback on this method:
 `fun storeRaw<T>(contiguous: T.Type, from: UnsafeRawPointer,
forwardToCount: Int)`

This is the `memcpy` API. It is very similar to bulk initialization:

`ptrToInt = rawPtr.initializeMemory(as: Int.self, from: srcAddress, count: n)`

But in this case, the memory does not need to be uninitialized, no
values are constructed, the memory is not bound, and no typed pointer
is returned.

For single value stores we have:
`rawPtr.store(x, asTrivial: Int.self)`

For bulk stores I think we should simply have:

`rawPtr.store(bytes: strideof(Int.self) * n, from: srcAddress)`

Proposed:
`func store<T>(bytes: Int, from: UnsafeRawPointer)`

proposals/0107-unsaferawpointer.md
dabrahams: on Thu Jul 07 2016, Andrew Trick <notifications-AT-github.com> wrote:

> For manual layout, I would slightly prefer that people use the
> "ByteOffset" APIs than bytewise pointer arithmetic.

  let x = p.load(fromByteOffset: 20, as: Int.self)

  let x = (p + 20).load(as: Int.self)

Just looking at these, I don't find the first one to be any clearer than
the second.  Do you?

-- 
Dave

proposals/0107-unsaferawpointer.md
dabrahams: on Thu Jul 07 2016, Andrew Trick <notifications-AT-github.com> wrote:

>>      -> UnsafeMutablePointer<T>
>> -  func initialize<T>(toContiguous: T.Type, atIndex: Int, with: T)
>> +  func initialize<T>(contiguous: T.Type, at: Int, to: T)
>
> I would like to distinguish initialization via raw pointer from
> initialization via typed pointer primarily because raw pointer
> initialization binds memory to a type.

They can both be viewed as doing that, can't they, since binding to a
type is idempotent?

>
> Option 1:
> --------
>
> ```swift
> // n copies of a single value
> rawPtr.initializeMemory(as: Int.self, count: n, to: x)
> // read n copies out of memory at `ptrToInt`
> rawPtr.initializeMemory(as: Int.self, from: ptrToInt, count: n)
> ```
>
> Option 2:
> --------
>
> ```swift
> // n copies of a single value
> rawPtr.initializeAndBind(as: Int.self, count: n, to: x)
> // read n copies out of memory at `ptrToInt`
> rawPtr.initializeAndBind(as: Int.self, from: ptrToInt, count: n)
> ```

Option 1 is much better, IM (subjective) O.

-- 
Dave

proposals/0107-unsaferawpointer.md
atrick: @jckarter requested the load-from-byte-offset feature as a significant improvement over `(p + offset).load(as:)`. I think Joe's main argument is that if `offset` is a complex expression, the `fromByteOffset` form will be much more readable. I agree.

I'm not particularly interested in very seldom used, purely additive APIs, but in this case I do think it the `byteOffset` form, for both load and store, makes it more clear that offset is in bytes and that the address computation is untyped. Otherwise, the user needs to infer that all from the pointer type.

proposals/0107-unsaferawpointer.md
atrick: Yes, binding is idempotent!
There is an important difference between `rawPtr.initializeMemory(as: T.self, to: T())` and `ptrToT.initialize(to: Pointee())` because the raw initialize can be called to `rebind` to a different type. The later has a precondition that the memory is already bound to `Pointee`.

That difference is somewhat signaled by the `as` label. However, I think it's also good to give the `raw` version a different name to clearly distinguish them. It then reads nicely as "initialize memory as `T`" because changing the memory's bound type is as important to this operation as changing the initialized state.

I do *not* want to rename all of the `UnsafePointer<T>` APIs with the `Memory` suffix, mainly because it would be excessive, cumbersome, and wouldn't solve any real usability problem:
- initialize
- moveInitialize
- assign
- moveAssign

I think "memory" is the obviously implied target of this API family.

proposals/0107-unsaferawpointer.md
dabrahams: on Fri Jul 08 2016, Andrew Trick <notifications-AT-github.com> wrote:

>>      -> UnsafeMutablePointer<T>
>>  
>> -  func load<T>(_: T.Type) -> T
>> -  func load<T>(_: T.Type, atByteOffset: Int) -> T
>> -  func load<T>(fromContiguous: T.Type, atIndex: Int) -> T
>> +  func load<T>(as: T.Type) -> T
>> +  func load<T>(fromByteOffset: Int, as: T.Type) -> T
>> +  func load<T>(fromContiguous: T.Type, at: Int) -> T
>
> @jckarter requested the load-from-byte-offset feature as a significant
> improvement over `(p + offset).load(as:)`. I think Joe's main argument
> is that if `offset` is a complex expression, the `fromByteOffset` form
> will be much more readable. I agree.

We have normal ways of dealing with that situation:

    let byteOffset = a complex expression
    (p + byteOffset).load(as: Int.self)

There's no need to design special provisions into this API for doing
what every Swift programmer knows how to do.

> I'm not particularly interested in very seldom used, purely additive
> APIs, but in this case I do think it the `byteOffset` form, for both
> load and store, makes it more clear that offset is in bytes and that
> the address computation is untyped. Otherwise, the user needs to infer
> that all from the pointer type.

Then are we planning to omit arithmetic on UnsafeRawPointer altogether?
If not, why not?

proposals/0107-unsaferawpointer.md
dabrahams: Memory is the target of the initialize family but not the others.  The others are operating on initialized `T`s.

Once you initialize memory to a particular type, it seems thoroughly obvious that the memory is bound to that type.  If any word is dispensable here, I’d say it’s “bind”.

PR #414, opened 2016-07-07
Move "ErrorProtocol" to right place in SE-0006

onevcat:
The change for `ErrorProtocol` should be placed in correct section under "Strip Type suffix" instead of "renaming to "iterator""


DougGregor:
Hmmm. SE-0112, if accepted, would turn ErrorProtocol into Error... but this is still an improvement. Thanks!

PR #418, opened 2016-07-08
[SE-0104] Protocol-oriented integers (revision 2)

moiseev:
Addressed [recommendations from the Core Team](https://lists.swift.org/pipermail/swift-evolution-announce/2016-June/000206.html).


DevAndArtist:
@moiseev one question: why do we use `rhs` as the attribute name in 'pseudo' infix functions?

```swift
func adding(_ rhs: Self) -> Self
func subtracting(_ rhs: Self) -> Self
func multiplied(by rhs: Self) -> Self
func divided(by rhs: Self) -> Self
// etc.
```

Wouldn't it be better to use `value` instead?

```swift
func adding(_ value: Self) -> Self
func subtracting(_ value: Self) -> Self
func multiplied(by value: Self) -> Self
func divided(by value: Self) -> Self
// etc.
```

moiseev:
@DevAndArtist no particular reason. Anyways, I've just renamed `rhs` to `other` where appropriate.

PR #419, opened 2016-07-09
Readability SE-0005 of the code

nohirap:
I was easy to see the code of the swift of the SE-0005.


CodaFi:
Two things:
1) If you're going to use explicit code sheds then please outdent the code.
2) When that is done please squash your commits.

Otherwise LGTM.

PR #420, opened 2016-07-10
Revise SE-0107: rename UnsafePointer & UnsafeRawPointer methods.

atrick:
I've revised this proposal based on last week's feedback. A few of the
additive APIs are removed and a number of UnsafePointer and
UnsafeRawPointer methods are renamed.

The examples in the text still need to be updated.

The full UnsafeRawPointer API with doc comments is here:
https://github.com/atrick/swift/blob/22e3a2885e4236888ec447a7148acf633d8544f5/stdlib/public/core/UnsafeRawPointer.swift.gyb

The proposed UnsafePointer and UnsafeRawPointer changes are on this branch:
https://github.com/atrick/swift/commits/rawptr

### DIFF
--- a/proposals/0107-unsaferawpointer.md
+++ b/proposals/0107-unsaferawpointer.md
@@ -1459,7 +1459,7 @@ in one place, a list of the expected `UnsafeMutableRawPointer` members
 is shown below.
 
 For full doc comments, see the [github revision]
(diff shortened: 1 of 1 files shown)


lattner:
@atrick feel free to merge changes to your proposals when you feel that is the right thing to do.

PR #422, opened 2016-07-12
Add a proposal to eliminate value-to-optional coercion in operator arguments.

rudkx:


lattner:
@rudkx I sent some specific feedback to swift-evolution.  Please revise the proposal however you'd like and lemme know when it is ready to merge.

rudkx:
@lattner I have most of your feedback incorporated, and am taking a quick look at a few more projects before I push a new version.

lattner:
Please email me when you're ready to kick this off.

rudkx:
@lattner There's been a bit of a snag with adding the overloads for equality and identity since they allow things like this to type check now:
  let i = 1
  if i == nil {}

I'm looking into what the options are for continuing to diagnose this while allowing mixed-type comparisons to proceed.

PR #424, opened 2016-07-12
Extend Optional Chaining to Function, Initializer, and Subscript Parameters

liam923:
This is a proposal to extend optional chaining to parameters.

### DIFF
--- a/proposals/NNNN-extend-optional-chaining-to-function-initializer-and-subscript-parameters.md
+++ b/proposals/NNNN-extend-optional-chaining-to-function-initializer-and-subscript-parameters.md
@@ -1,0 +1,141 @@
+# Extend Optional Chaining to Function, Initializer, and Subscript Parameters
+
+* Proposal: [SE-NNNN](NNNN-filename.md)
+* Author: [Liam Stevenson](https://github.com/liam923)
+* Status: **Awaiting review**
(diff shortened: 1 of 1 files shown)


marcusrossel:
I really like the idea, as it basically shifts the responsibility of handling `nil`-values from the function author to the function user. This might reduce the number of functions that take optional parameters, as it isn't more convenient than using the suggested syntax, which should in turn result in more functions with non-optional return values. It reminds me of the `rethrows` keyword, which basically makes a function throwing is you pass a throwing closure. And your syntax makes the return type of a function optional if you pass it an optional. This seems very consistent with Swift as a whole.

liam923:
@marcusrossel I hadn't thought about its similarity to `rethrows`, but now that you mention it, I agree that it's similar. I'm glad that you agree with me that it conforms with Swift's design. I think that the syntax is consistent with optional chaining and optionals as a hole.

lattner:
This is an additive proposal that is out of scope for Swift 3.  It also has not had substantial public discussion on the mailing list.  Please wait until after Swift 3 is done to bring this back up.

liam923:
@lattner When will that be? After Swift 3's formal release?

lattner:
Sometime in august.

PR #426, opened 2016-07-12
Propose “Reinstate Mixed-`Optional` Equality Operators”

jtbandes:
[Rendered Markdown](https://github.com/jtbandes/swift-evolution/blob/59dbf349ff18d2e4c29a87d1a34b1e214b717778/proposals/NNNN-reinstate-mixed-optional-equality-operators.md)

### DIFF
--- a/proposals/NNNN-reinstate-mixed-optional-equality-operators.md
+++ b/proposals/NNNN-reinstate-mixed-optional-equality-operators.md
@@ -1,0 +1,135 @@
+# Reinstate Mixed-`Optional` Equality Operators
+
+* Proposal: [SE-NNNN](NNNN-reinstate-mixed-optional-equality-operators.md)
+* Author: [Jacob Bandes-Storch](https://github.com/jtbandes)
+* Status: **Awaiting review**
+* Review manager: TBD
+
+## Introduction
+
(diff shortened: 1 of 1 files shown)


lattner:
I think that this should be incorporated into @rudkx's proposal, since we wouldn't want to accept it without doing this.

jtbandes:
I think some folks, @rudkx included, are skeptical about whether this is worthwhile, so I'm not sure it should really be in the same proposal. But, @rudkx, if you agree then feel free to incorporate any of this in your proposal and we can abandon this one.

rudkx:
@jtbandes @lattner No, at first I thought this was something we could address separately due to the very small number of changes required in most projects, but I think the dictionary example has me convinced that we might as well make this completely painless.

I'll go ahead and add you as an author if you don't mind. Your feedback has been very useful.

lattner:
Yes, the dictionary example is what I'm concerned about.  It is better (IMO) to preserve existing behavior for that.  No need to list me as a coauthor.

rudkx:
Actually I mean @jtbandes, but yes your feedback on the proposal itself has been very helpful as well. :)

jtbandes:
Sounds good to me. I'm glad all these are coming together for Swift 3 :)

PR #431, opened 2016-07-14
Revise SE-0107 UnsafeRawPointer: remove default alignedTo argument.

atrick:


lattner:
Thanks @atrick and congrats!

PR #434, opened 2016-07-15
'(U)Int.init(ObjectIdentifier)' should have a 'bitPattern:' label

aschwaighofer:
`Int.init(ObjectIdentifier)` and `UInt.init(ObjectIdentifier)` should have a `bitPattern:` label to be consistent with other similar APIs.

Initial draft.


gribozavr:
I think this is good to go!  :shipit:

PR #436, opened 2016-07-18
Update 0117-non-public-subclassable-by-default.md

agnosticdev:
Under the motivation header, change interesting to inheritable so that the reader had a better understanding of the sentence meaning.


lattner:
John is planning to revise the proposal more substantially.  Thanks

PR #437, opened 2016-07-18
Cleaning up stdlib Pointer and Buffer Routines proposal.

charlieMonroe:


moiseev:
@charlieMonroe, thanks for preparing the proposal!
Just a heads-up, the following JIRAs are covered by it:

- https://bugs.swift.org/browse/SR-1937
- https://bugs.swift.org/browse/SR-1955
- https://bugs.swift.org/browse/SR-1956

bob-wilson:
I didn't see the earlier discussion and have been investigating SR-1937 and SR-1956 (the withUnsafe[Mutable]Pointer issues).

With regard to SR-1937, besides adding the "to:" argument label, I'd like to suggest that we remove the 2- and 3-argument versions: withUnsafe[Mutable]Pointers. The naming is already different than the 1-argument functions (plural vs. singular) and the argument label doesn't work so well with multiple arguments. These extra functions are not a complete solution (if you need more than 3 arguments) and they're not used much -- there is only one real use in the stdlib, with one more use in a validation-test. We can replace them by composing nested uses of the 1-argument functions in the few places where they are needed.

For SR-1956, I discussed Jordan's objection with Dmitri and Dave and we're leaning toward keeping the inout argument. But, if the argument has to be inout anyway, then there's not much point in having separate functions for UnsafePointer vs. UnsafeMutablePointer. If you have an UnsafeMutablePointer, there is an implicit conversion to use it in place of an UnsafePointer. That means we can have a single "withUnsafePointer" function that under the hood uses an UnsafeMutablePointer. The "withUnsafeMutablePointer" function can then be removed.

I have prototyped these changes in the stdlib and they seem to work fine.

charlieMonroe:
@moiseev - yes, they are all linked by the proposal, but since it's changing the public API, it should go through evolution first - at least that's what I was told :)

@bob-wilson -

There has been a generic evolution thread covering all these issues since they were on a longer list of issues and these (marked as easy) were thrown together (http://thread.gmane.org/gmane.comp.lang.swift.evolution/23093). There hasn't been much discussion, but given the deadline for approving any changes, I decided to put the proposal together anyway.

Should I update the proposal suggesting removal of the plural `withUnsafe[Mutable]Pointers` methods in favor of nesting several single-pointer ones? Personally, it kind of makes sense to me, if their uses cases are limited and can be easily replaced by nesting...

I've read Jordan's objections, but is there any reason, why not use `withUnsafeMutablePointer` instead, if you need `inout`? Sorry if I'm missing something... I agree that if `inout` is kept in place, there's no much point in keeping both.

moiseev:
Oops... I missed the links to the JIRAs in the proposal. My bad. And yes, it definitely should go through the evolution. I did not mean otherwise.

bob-wilson:
If we change withUnsafePointer to not take an inout argument, then the argument will be passed by value and the pointer will reference a copy. At least for globals, Jordan says there is an expectation that withUnsafePointer will always get the same address, which would not be true if there was a copy. Changing the behavior of withUnsafePointer in this way would be disruptive. I can see the convenience of having an API like withUnsafePointer that does not require an inout argument, but it would be better to introduce a new API with a different name for that purpose. That is not something we need to do now for Swift 3.

charlieMonroe:
How about the current usage of withUnsafePointer is changed to withUnsafeMutablePointer during the migration to Swift 3? This would leave the current behavior for people disiring it, but in case of objects, inout doesn't make sense since it will still be a pointer to the heap, but you need to have the variable as var.

On the other hand, you are right that simply eliminating all the other functions aside from withUnsafePointer and having withUnsafePointer to pass mutable pointer into the closure is a solution that would simplify the API and keep the current behavior.

eeckstein:
The "`ManagedProtoBuffer` should be removed" part LGTM.

PR #438, opened 2016-07-18
Proposal: harmonize access modifiers

xwu:
This is a follow-on in the wake of Adrian's withdrawal of SE-0119 from consideration. It seems to have garnered a modicum of consensus on the list.

### DIFF
--- a/proposals/XXXX-harmonize-access-modifiers.md
+++ b/proposals/XXXX-harmonize-access-modifiers.md
@@ -63,6 +63,10 @@ This implies that public API commitments will need to be annotated as `public` a
 
 *  One alternative, still open for consideration, is to eliminate #4 and disallow explicit access modifiers on extensions. As an advantage, this would clarify the mental model that extensions are not their own entities, as they cannot be referred to by name and have no runtime representation. As a disadvantage, extensions cease to be an access modifier grouping construct, which some users really like.
 
+## Acknowledgments
+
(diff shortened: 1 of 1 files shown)


lattner:
This has a modicum of consensus, but not strong consensus.  This doesn't seem like it is likely to be successful, so I don't think it meets the bar to merge.  @DougGregor what do you think?

DougGregor:
I agree with @lattner ; the support on the list is lukewarm at best. I don't think we have enough consensus to move forward.

masters3d:
@lattner I am curious about proposals like [105](https://github.com/apple/swift-evolution/blob/master/proposals/0105-remove-where-from-forin-loops.md) and [119] (https://github.com/apple/swift-evolution/blob/master/proposals/0119-extensions-access-modifiers.md) which had strong negative feedback during the socializing state. 
Should proposals with strong but negative feedback be merged? I know we are at the end of swift 3 and time is of the essence but I wanted to get clarification so maybe I can help update the docs.

DougGregor:
We try to assess consensus from earlier discussions, and if it doesn't look like something that will pass review, we won't bring it forward because doing so takes up reviewer bandwidth.

lattner:
@masters3d both of those proposals were rejected.  That is precisely what we're aiming to avoid: we don't want to expend community and core team effort on a proposal does not have consensus, because it is likely to be rejected.

PR #449, opened 2016-07-20
Create nnnn-variadics-as-attribute.md

Haravikk:
Reviewed pull-request for Variadics as Attribute.


lattner:
In the discussion on the list, several people on the main mailing list are strongly against this, and I see little support.  I also see little chance that this would be accepted by the core team.  If you're serious about pursuing this, please continue on-list discussion and try to obtain more consensus.

PR #456, opened 2016-07-22
Replace String.init(repeating repeatedValue: Character, count: Int) and String.init(repeating repeatedValue: UnicodeScalar, count: Int) by String.init(repeating repeatedValue: String, count: Int)

swiftix:
This proposal suggest replacing String initializers taking Character or UnicodeScalar as a repeating value by a more general initializer that takes a String as a repeating value. This is done to avoid the ambiguities in the current String API, which can be only resolved by explicit casting. It is also proposed to remove one of the String.append APIs to match these changes.

It is addressing https://bugs.swift.org/browse/SR-1964 (rdar://problem/17028332).

The idea has been discussed on the [following swift-evolution thread](http://thread.gmane.org/gmane.comp.lang.swift.evolution/24626) already.


gribozavr:
LGTM.

PR #464, opened 2016-07-25
Formalized ordering

pyrtsa:


lattner:
The discussion on this has not converged on -evolution.  I think this topic has missed the window for swift 3.

PR #467, opened 2016-07-25
Add proposals/xxxx-fix-confusion-about-consuming-algorithms-on-sequence.md

gribozavr:


lattner:
The risk/reward tradeoff of taking this is too low.  Lets discuss it for Swift 4 sometime next week.

PR #468, opened 2016-07-25
[Revised SE-0126] [Ready for review] Refactor metatypes

DevAndArtist:
@lattner We're completely refactored the idea and believe that we found a good solution for the current metatypes mess. Please merge and kick off review.

Sorry for the wrong proposal that was attached in the last PR. That was an accident.


lattner:
I'm sorry @DevAndArtist but it is too late to take this.  This has major impact that we don't have time to design and *implement* in time for Wednesday.

PR #482, opened 2016-08-01
Gmane to Pipermail link changes

kkdevenda:
I have updated gmane links with pipermail for proposals [SE-0017] to [SE-0029] and [Se-0040] to [SE-0044].


Danappelxx:
You missed some (I only checked two, but neither SE-0030 or SE-0033 were fixed).

kkdevenda:
@Danappelxx For now I updated [SE-0017] to [SE-0029] and [Se-0040] to [SE-0044] only. I will update remaining ([SE-0030] to [SE-0039]) very soon.

PR #488, opened 2016-08-04
Update status for SE-0104

jessesquires:
I *think* this is completed for Swift 3, but I'm not sure.

Based on: https://github.com/apple/swift/pull/3479 and https://github.com/apple/swift/pull/3502.

cc @moiseev


lattner:
This didn't make it.

PR #494, opened 2016-08-08
[SE-0028] Update swift-version in proposals.xml

benrimmington:
[SE-0028](https://github.com/apple/swift-evolution/blob/master/proposals/0028-modernizing-debug-identifiers.md) was implemented in **Swift 2.2** according to the [CHANGELOG.md](https://github.com/apple/swift/blob/master/CHANGELOG.md#swift-22) and the apple/swift#1218 pull request.


jtbandes:
Thanks!

PR #507, opened 2016-08-30
Update UnsafeRawPointer proposal

natecook1000:
Updates the examples to use the new `MemoryLayout` APIs and fixes a
few typos.

### DIFF
--- a/proposals/0107-unsaferawpointer.md
+++ b/proposals/0107-unsaferawpointer.md
@@ -30,11 +30,11 @@ circumvent the language's natural type safety. Consider the following
 example of *type punning* using the `UnsafePointer` type:
 
 ```swift
-  let ptrT: UnsafeMutablePointer<T> = ...
-  // Store T at this address.
-  ptrT[0] = T()
-  // Load U at this address
-  let u = UnsafePointer<U>(ptrT)[0]
+let ptrT: UnsafeMutablePointer<T> = ...
+// Store T at this address.
+ptrT[0] = T()
+// Load U at this address
+let u = UnsafePointer<U>(ptrT)[0]
 ```
 
 This code violates assumptions made by the compiler and falls into the
@@ -62,7 +62,6 @@ func takesUIntPtr(_ p: UnsafeMutablePointer<UInt>) -> UInt {
 func takesIntPtr(q: UnsafeMutablePointer<Int>) -> UInt {
   return takesUIntPtr(UnsafeMutablePointer(q))
 }
-
 ```
 
 Furthermore, no API currently exists for accessing raw, untyped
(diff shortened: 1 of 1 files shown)


natecook1000:
@atrick Can you review?

proposals/0107-unsaferawpointer.md
```diff
@@ -672,8 +670,8 @@ pointer cast is explicit:
 ```swift
 // --- new version ---
 func testInitAorB() {
-  let p = UnsafeMutableRawPointer.allocate(bytes: sizeof(Int.self),
-                                           alignedTo: alignof(Int.self))
+  let p = UnsafeMutableRawPointer.allocate(bytes: MemoryLayout<Int>.stride,
+                                           alignedTo: MemoryLayout<Int>.alignment)
 
   initRawAorB(p, isA: true)
   printA(p.assumingMemoryBound(to: A.self))
```
natecook1000: @atrick Calling `initRawAorB(p, isA: true)` and then `initRawAorB(p, isA: true)` in this function violates the stated preconditions for `initializeMemory(as:to:)`, since the second call initializes memory for `B` that is already initialized for `A`. Is this permitted because `A` is a trivial type and `A` and `B` are layout compatible, or is deinitialization required here?

proposals/0107-unsaferawpointer.md
```diff
@@ -1460,15 +1463,15 @@ even encourages interoperability between C APIs and a String's UTF8
...
-      .cast(to: UnsafePointer<CChar>)
-    let fd = mkstemps(cStrBuf, suffixlen)
-    let fileName = String(cString: cStrBuf)
-    ...
-  }
+var utf8 = template.nulTerminatedUTF8
+let (fd, fileName) = utf8.withUnsafeMutableBufferPointer {
+  (utf8) -> (CInt, String) in
+  let cStrBuf = UnsafeRawPointer(utf8.baseAddress!)
+    .assumingMemoryBound(to: UnsafePointer<CChar>)
```
natecook1000: @atrick I figured `cast(to:)` was an earlier name for `assumingMemoryBound(to:)`—is that correct?

proposals/0107-unsaferawpointer.md
atrick: It's ok to reinitialize trivial types. It doesn't even matter here that they are layout compatible (as long as the allocated memory is large enough and properly aligned). I think there was initially some concern about exposing the term "trivial type" in these doc comments. But that cat is out of the bag.

I'll fix the doc comments:
https://github.com/apple/swift/pull/4560

proposals/0107-unsaferawpointer.md
atrick: Yes, that's fine.

atrick:
Awesome. Looks great!

araykopotay:
a5afale

PR #523, opened 2016-09-11
Update SE-0138: UnsafeRawBufferPointer after week #1 of community rev…

atrick:
…iew.

The review is extended to September 14.


lattner:
Please make sure that an extension of the review period gets announced, so people are aware that there was a change and that the period has been extended.  Thanks!

lattner:
Oh... you already did this, never mind! :-)

PR #524, opened 2016-09-12
[gardening] fix inconsistent proposal statuses

jessesquires:
Update statuses for SE-0018, SE-0030, SE-0079, SE-0100 to be consistent with the [Proposal Status](http://apple.github.io/swift-evolution/) page.

These should all have a status of "deferred".


jessesquires:
cc @jtbandes for reviewing this PR

PR #528, opened 2016-09-15
Forbid weak keyword for computed properties

s1ddok:


jtbandes:
This seems like more of a bug than a language change that'd require review. @DougGregor?

lattner:
This seems like something we should warn about, not emit an error for.  As such, it doesn't need an evolution proposal, and doesn't break source compatibility.

s1ddok:
In my opinion it will make Swift error-safe if it will not be available at all. As it makes room for some occasional mistakes. Especially in that custom get/set pair area. One might simply put `weak` keyword in the wrong place. 

Emitting error for such things will make Swift more fool-proof and that is what we all love, aren't we?

fbartho:
Hold-on, the weak attribute on a public property could be a signal to the outside world. Just because it has getters/setters doesn't mean that the property isn't semantically weak internally. 

The more complex example is useful case & should not be banned -- assuming of course the API implementor has the private member variable correctly tagged with `weak`

s1ddok:
Yup, but that is the problem. You have to mark internal stored property as weak, while it has no effect on public one. Therefore it playa the documentation function, while not actually promising anything. One can simply forget to put weak onto internal property or even worse can think that parent will handle it all.

lattner:
From an implementation perspective, weak is effectively an implementation detail of a stored property, so I can see where @s1ddok is coming from.  However, I completely disagree with the point about error-safety vs weak.  That point has nothing to do with the choice of error vs warning.  You could apply the same logic to every warning we produce today, arguing that they should be errors.  That is simply not the basis we use for the decision between error or warning.

codinggirl:
As @fbartho said, ```weak``` just used for a public interface, we should use it in private vars in a correctly way.

Maybe we should place a warning when set a ```weak```-marked computed property to non-weak instance var in custom setter.

s1ddok:
I dont see why would language like Swift contain `keyword` that has no effect at all in certain cases.

As @lattner admits, that's it only affects stored properties just like I thought it does, I think we should semantically separate stored properties from computed ones. As of property with custom get/set pair plays a role of `getProperty()`/`setProperty()` pair of methods, just in a more convenient way.

You can put weakness of the property in the documentation, why would one want to use a keyword for that reason? 

I think implementing static code analyzer, that will check for all cases, where `newValue` is being assigned to strong reference property from `weak set` will be hard or even near to impossible since cases may be very complex. 

I don't really think I'm in position to argue with any of you guys, but I think **every** keyword must be doing at least something. While `weak` keyword claims something, that it can't guarantee.

fbartho:
@s1ddok I think my main objection to removing the use in the second example is that it serves a very useful communication purpose for consumers of an API.

If you look at the `Generated Interface`/Closed Source Binary you can't possibly know about the private member variable. The private member variable is an internal implementation detail that would be very difficult to enforce in a sane way for all use cases.

Other examples:
- mark a computed property as weak, when the internal implementation actually keeps it as a strong reference for a period of time, and then releases it.
- mark a computed property as weak to hint to the user that the value may be nil-ed at any time depending on some other object's lifecycle. (In this case, no actual weak storage is necessary, but semantically to the API consumer, it feels like a weak object, that they need to preserve)
…

s1ddok:
I totally understand what cases can be used for implementing custom `get/set` pair and how they can be semantically weak. 

It's just in my experience, such cases are usually covered by documentation, not obsolete keywords, especially the very complex ones that you mentioned. 

You still have to read the interface to know if certain property behaves weak or not. But when you read the docs, the API provider **guarantees* that, while when you see a keyword, you think it is guaranteed by the language, while it is not. 

I think we are speaking about same thing but from different angles. My point is just that keywords should not be a part of documentation.

fbartho:
We are indeed speaking about the same thing.

I however, can't emphasize enough how much I disagree with your statement:
>My point is just that keywords should not be a part of documentation.
I'm definitely on the opposite side of things there.

:P Cheers!

proposals/NNNN-forbid-weak-keyword-for-computed-properties.md
```diff
@@ -0,0 +1,72 @@
...
+        set {
+            // some **necessary** logic here, that can't be putted in willSet
+            _f = newValue
+            // some **necessary** logic here, that can't be putted in didSet
+        }
+    }
+}
+```
+
+We all love Swift's `didSet` and `willSet` code containers, they are really useful in avoiding excess code and adding certain behaviour to stored properties. But sometimes we **really** want to implement our own getter/setter. For example if we need to handle both will- and did- set cases, but need to remember some state in between them. As you can see in the example above, our computed (but not a get-only) property is marked by `weak` keyword and has an optional type on it, but the things is that it is mapped to strong private property, which makes the whole thing obsolette.
```
DougGregor: We really don't need this much editorializing here (the "we all love..." bits are really distracting for the read). The important point is the last sentence: the property is marked `weak` but actually puts the value into a strong private, stored property, so `weak` is misleading here. That's all you need to say.

proposals/NNNN-forbid-weak-keyword-for-computed-properties.md
```diff
@@ -0,0 +1,72 @@
...
+
+The problem is not only that keyword has no affect and brings obsolette restrictions (forcing optional type), but also leads to ambigouty and brings a lot of confusion while debugging and reading code.
```
DougGregor: s/affect/effect, typo "obsolette".

CodaFi:
@lattner Would a patch [like this](https://github.com/CodaFi/swift/commit/c3b55a0661091dd9190648a3cf4fe0067b97c9b2) (plus tests) be acceptable to submit now?

DougGregor:

s1ddok:
@DougGregor will you notify community of the possibility to reopen PRs?

DougGregor:
We'll send out a note when we're re-opening the process for more proposals. It'll be a few months before we get there.

PR #529, opened 2016-09-16
proposal 0022 typo

zhugejunwei:
change "lets" to "let"


jtbandes:
Oops, I was too quick to accept this. The original is correct. I've reverted it.

PR #530, opened 2016-09-17
Change shadow copy to shallow copy

wongzigii:
It should be shallow copy instead of shadow copy.
https://developer.apple.com/library/content/documentation/General/Conceptual/DevPedia-CocoaCore/ObjectCopying.html


jtbandes:
This isn't correct; this proposal has nothing to do with NSCopying. Shadow copy is like a regular copy of any struct type, but this one is implicit, hence "shadow".

PR #532, opened 2016-09-23
SE-0140 and SE-0110 have been implemented

amartini51:


DougGregor:
Are you going to update index.xml as well?

amartini51:
Fixed index.xml in 1b8345c, thanks.

amartini51:
Looks like our tracking bugs are wrong then:  [Jira](https://bugs.swift.org/browse/SR-2008) and [Radar](rdar://problem/27383557) say this landed.  Good thing this got called to our attention!

benrimmington:
@DougGregor apple/swift@f42158b12e2072ac258817edd455de3747a3ce38 is apple/swift#3922, which only reverts apple/swift#3895 (the first attempt of SE-0110).

apple/swift#4102 (the second attempt) has not been reverted. It isn't in the `swift-3.0-branch`, so the current status should be **Implemented (Swift 3.1)**.

Would it be better to wait until Swift 4, either by reverting apple/swift#4102, or by using the `-swift-version` flag of apple/swift#4826?

PR #535, opened 2016-09-27
SE-0138: UnsafeRawBufferPointer, minor revision.

atrick:
This removes the unnecessary hack of passing a closure argument as
'inout'. The solution is to declare both subscript setters as
'nonmutating'.

This hack was inherited from the UnsafeBufferPointer interface and that
detail was overlooked during proposal reviews. It came up during code
review with DaveA.


lattner:
Feel free to merge this whenever @atrick

PR #537, opened 2016-10-03
Typo correccted in the proposal.

kandelvijaya:
There was a hanging incomplete sentence which i think its better to remove.


jtbandes:
`its` was correct; `it's` is incorrect.

lattner:
Thanks!

PR #542, opened 2016-10-07
mark SE-0139 as implemented in Swift 3.0.1

kostiakoval:
mark SE-0139 as implemented in Swift 3.0.1


jtbandes:
It looks like the commits for this went onto `swift-3.0-branch` a couple weeks ago. Does that mean they're in 3.0.1?

kostiakoval:
Swift 3.0 was release with Xcode 8, and this feature wasn't included there. 
So I think it should go the the next release, which is 3.0.1. Also its mentioned in Xcode [release notes](http://adcdownload.apple.com/Documentation/Xcode_8.1_beta_2_SDK_Release_Notes/Release_Notes_for_Xcode_8.1_beta_2.pdf)

if you try this code snipper from SE-0139 (in Xcode 8, swift 3.0) it will crash 
```swift
let i = 17
let plist = ["seventeen": i]
// OK
try! JSONSerialization.data(withJSONObject: plist)

let j: UInt8 = 38
let brokenPlist = ["thirty-eight": j]
// Will throw because `j` didn't bridge to a JSON type
try! JSONSerialization.data(withJSONObject: brokenPlist)
```

PR #543, opened 2016-10-07
Update 0139-bridge-nsnumber-and-nsvalue.md

kostiakoval:
fix code snipper to latest Swift syntax, without NS prefix


lattner:
thanks!

PR #545, opened 2016-10-13
Proposal for removing ManagedBufferPointer

eeckstein:


Interfere:
Hi. Is there are any additions if I need `allocWithTailElems_2`, `allocWithTailElems_3` routines, for my own containers?

eeckstein:
@Interfere You cannot use allocWithTailElems_<n> outside the swift standard library, because they are Builtins.
As a replacement for ManagedBufferPointer you can use ManagedBuffer.

Interfere:
@eeckstein Yeah, I know that. I think it's not fair if stdlib of the language uses a bunch of hacks not available to other developers. ManagedBuffer uses allocWithTailElems inside, by the way.

The question is: is there any alternatives for other allocWithTailElems_{n} routines, like ManagedBuffer2, ManagedBuffer3 etc.? Could we provide them? It's pretty unpleasant to force developers to use pointer math instead.

Sorry, I think we'd better move that discussion to mailing list.
Thanks.

lattner:
Is this still interesting to consider?

DougGregor:
Core team: there hasn't been movement on this in a long time. Please re-open the discussion if you're still interested.

PR #546, opened 2016-10-14
Proposal of private protocol conformances

ChrisRutkowski:


lattner:
This is an additive proposal that is out of scope for Swift 4 stage 1.  Please see the main swift evolution page (https://github.com/apple/swift-evolution) for more details of what is currently in scope.  Thanks.

PR #549, opened 2016-10-27
Update status of SE-0138

jessesquires:
As specified on the [status page](http://apple.github.io/swift-evolution/)


jtbandes:
:thumbsup:

PR #552, opened 2016-11-05
Ease restrictions on protocol nesting

karwa:


brentdax:
Not an authority, just a guy, but a couple things:

1. We should probably specify that the standard library will include typedefs for the old names of the types you rename. I'm not sure if they should be version-limited or not.

2. Personally, I'm not comfortable with nesting concrete types inside protocols unless they are mere defaults for associated types and can be replaced in a conforming type. Everything else in a `protocol … {}` declaration specifies an abstract requirement; I'm not convinced it makes sense to break that pattern. Instead, I think that concrete nested types should be put in protocol extensions:

```swift
protocol FloatingPoint {
    var sign: Sign { get }
}
extension FloatingPoint {
    enum Sign {
        case plus
        case minus
    }
}

struct Float : FloatingPoint {
    var sign: FloatingPoint.Sign { /* return the sign */ }
}
```

karwa:
Sure - even though there was a mailing list discussion, the lists haven't been all that active recently so I still welcome input. I'll stick a [WIP] tag on it so everybody can give it another look over.

The idea is that these are **not** associated types; they are types which only make sense in the context of the `FloatingPoint` protocol. Everything which conforms to `FloatingPoint` must be able to give a `FloatingPointClassification` which is one of a few specific values defined by the protocol. They can't define their own classification enum and return that instead; that *would* be an associated type. Also, there is nothing that says a nested type can't be the default for an associated type, if that makes sense for your model:

```
protocol TextStream {
     protocol Transformer { 
        static func transform(_ char: Character) -> Character

        struct None : Transformer {
             static func transform (_ char: Character) -> Character { return char }
        }
     }
     
     associatedtype Processor : Transformer = Transformer.None
}
```

We all want this feature, so I want to make the proposal as non-prescriptive as possible and hopefully make it non-controversial by doing that. But I'm starting to think that by allowing function bodies in protocol definitions at all - even inside of nested types - will be controversial.

lattner:
This is an additive change, we'll consider it in Swift 4 stage 2.

karwa:
@lattner what about standard library changes?

slavapestov:
> Personally, I'm not comfortable with nesting concrete types inside protocols unless they are mere defaults for associated types and can be replaced in a conforming type. Everything else in a protocol … {} declaration specifies an abstract requirement

This is not quite true, @brentdax - typealiases in protocols are not requirements.

proposals/xxxx-ease-protocol-nesting.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Proposed solution
+
+There are two important restrictions to this proposal:
+- Nested protocols may not capture generic type parameters from their contexts
+- Nested types (including protocols) may not capture associated types from their contexts
+
+These restrictions are due to currently-limited support for existential types. There are many interesting ideas to overcome both, but after discussions on the mailing lists, we should be able to handle most of the common cases with these limitations and it keeps things reasonable to implement.
+
+The first part is to allow protocols to be nested inside of structural types (for example, in the delegate pattern):
```
slavapestov: "structural type" is not the right term here. Call it a "nominal type"

proposals/xxxx-ease-protocol-nesting.md
```diff
@@ -0,0 +1,126 @@
...
+        case plus
+        case minus
+    }
+    
+    var sign: Sign { get }
+}
+
+struct Float : FloatingPoint {
+
+    var sign: FloatingPoint.Sign { /* return the sign */ }
```
slavapestov: Elsewhere can I also write 'Float.Sign' -- ie, will name lookup find nested types in protocols that the base type conforms to?

karwa: That isn't the case for typealiases today (I don't know whether or not it _should_; personally I'd really like it - it helps with complex closure types). The name lookups for actual types in protocols should be the same as our current rules for typealiases.

karwa:
Also, it turns out we've considered function bodies inside protocol definitions before, as a 'syntactic enhancement', so I'm feeling less like we need to implement a no-bodies rule for nested types.

https://github.com/apple/swift/blob/master/docs/GenericsManifesto.md#default-implementations-in-protocols-

lattner:
> @lattner what about standard library changes?

Ah ok, if that's the motivation, then I'd make that more clear in the title of the proposal.  I'd like a standard library person to chime in about whether they think it is important.  @airspeedswift or @dabrahams ?

dabrahams:
@lattner I don't know what standard library changes we're considering. If the question is, is the ability to do this kind of nesting important for the standard library, the answer is yes.

DevAndArtist:
> I'm not convinced it makes sense to break that pattern. **Instead, I think that concrete nested types should be put in protocol extensions.**

Exactly what I was saying on the mailing list!

Otherwise how am I supposed to deal with different access modifiers, because their declaration inside the protocol is disallowed.

```swift
public protocol A {
    
   struct B {
      var c: A { ... }
   }
}
```

Are `B` and `B().c` also public?

Vs.

```swift
public protocol A {}

extension A {
   /* any access modifier I want */ struct B {
      /* now it's internal by default */ var c: A { ... }
   }
}
```

karwa:
@DevAndArtist well, the point is that these types are _part_ of the protocol. For example, the FloatingPointClassification enum is as much a part of FloatingPoint as any of its functional members. It is not something which can be overridden, and so putting it in an extension feels "wronger" (IMO) than putting functional code in a protocol definition.

As for access control, I don't see an issue with the current language rules. Things inside the protocol definition are necessary for conformance; access control is therefore meaningless at that top-level, otherwise you may have an incomplete picture of the required types. To illustrate:

```swift
protocol A {
    // at top-level of definition, no access control - everything is necessary to implement conformance
    enum KindOfThing { case thingOne; case thingTwo }
    var kind : KindOfThing { get }
}
```

But once we get inside of more complex types (such as structs or classes), is it *not* necessarily true that every internal member is necessary for conformance. Therefore access control is meaningful:

```swift
public protocol A {
    func process(data: B)

    // A.B is top-level in the protocol, therefore public
    struct B {
        // This is a bog-standard struct body, therefore `c` is implicitly internal.
        var c : String
        public var d : Int
    }
}
```

Note that the non-public member `A.B.c` means that `A.B` does not have a public default initialiser -- that's fine; it doesn't necessarily have to be publicly-constructible. The protocol `A` only defines that types need to read data from a `A.B` (`process(data: B)`). Public conformers will only be able to see the member `A.B.d` in that case. In order to make A.B constructible by others, we would need to make `.c` publicly-visible or add an initialiser (in the struct definition body itself, or in an extension).

If it proves too onerous to explicitly mark-up the public members of nested structs, we could consider making them implicitly match the visibility of the protocol.

The thing about protocol extensions is that they aren't necessary for conformance. They're just bags of extra functionality built on top of the actual requirements. Because of that, they can have any access modifier, and the same would apply to any types they defined to help them:

```swift
extension Integer {
    fileprivate enum Sign { case positive; case negative }
    fileprivate var sign : Sign { return self < 0 ? .negative : .positive }
}
assert(42.sign == .positive)
assert(-36.sign == .negative)

// Some other file...
let sign : Integer.Sign       // Error: `sign` is marked `fileprivate`
```

karwa:
@DougGregor could I ask for your opinion on the generics part of this?

DevAndArtist:
@karwa haven't read the new version yet, but a first glance I noticed that the indent of the very last code samples is wrong. ;)

proposals/xxxx-ease-protocol-nesting.md
```diff
@@ -0,0 +1,292 @@
...
+
+## Motivation
+
+Nesting types inside other types allows us to scope their usage and provide a cleaner interface. Protocols are an important part of Swift, and many popular patterns (for example, the delegate pattern) define protocols which are intended to be used in the context of other types. It would be nice to apply type-nesting here: `MyClass.Delegate` reads better than `MyClassDelegate`, and literally brings structure to large frameworks.
+
+Similarly, we have examples in the standard library where supporting types are defined with the intention that they be used in the context of some protocol - `FloatingPointClassification`, `FloatingPointSign`, and `FloatingPointRoundingRule` are enums which are used by various members of the `FloatingPoint` protocol. It would also be nice to apply type-nesting here, with the types belonging to the protocol itself - e.g. `FloatingPoint.Sign`.
+
+## Proposed solution
+
+The first part is to allow protocols to be nested inside of structural types (for example, in the delegate pattern):
```
slavapestov: nominal types, not structural types

karwa:
@lattner I think this is ready to go

lattner:
@DougGregor would be a better review manager for this.

karwa:
I know I said it was ready, but I was working on something unrelated when I realised that this "generic workaround for capturing an associated type" is actually something we do in the standard library all the time. We have loads of views on various types of collection which all take that protocol as a single generic parameter - essentially capturing the type.

That means they're good candidates for nesting. So this change includes loads more changes to the standard library now. @dabrahams or @airspeedswift might also be interested.

karwa:
I haven't seen a proposal for the standard library after conditional conformances, but I imagine most of that collapsing should happen as part of that change, e.g:

```swift
LazySequence<T>
LazyCollection<T>
```

will become (I assume):

```swift
LazySequence<T>
extension LazySequence: Collection where T: Collection
```

I believe it would make sense to nest the resulting type in `Sequence`. It wouldn't be inherited, but `Sequence` could choose to expose a typealias (which is inherited by current language rules):

```swift
extension Sequence {
     struct Lazy<S: Sequence> { /* ... */ }
     typealias Lazy = Lazy<Self> // "captures" Self, therefore a good candidate for inheriting
}

let _: Array<Int>.Lazy = [1, 2, 3].lazy    // type is really: Sequence.Lazy<Array<Int>>
```

which would be much more convenient than the current spelling: `LazySequence<Array<Int>>` (actually we could add these convenient typealiases even without protocol nesting, but the fact that the typealias would live inside a protocol makes a more compelling case for the implementation to live there, too)

karwa:
@dabrahams As for `Collection.Empty` and `Collection.One` - they are interesting, almost the protocol equivalent of a singleton (an implementation which always exists). They are definitely a different conceptual thing to `FloatingPoint.Sign` or `Collection.Slice`, but do they really _not belong_ inside the protocol?

I also considered moving those types to `AnyCollection<T>` - as `AnyCollection<T>.Empty` and `AnyCollection<T>.One`, but I think the AnyCollection wrapper is really a temporary crutch for limited existential support, so it may not be as good a home as the actual Collection protocol.

karwa:
@slavapestov I can't reply to your comments directly so will have do batch them here:

> With the current implementation of name lookup, qualified and unqualified lookup will find the nominal type as a member of the conforming type. In Swift 3 there was a bug where unqualified name lookup didn't find a typealias, but this is fixed now.

I think the nicest rule would be that, if the nested type is used by a protocol requirement, it *does* actually get imported in to the namespace of the conforming type (the compiler would generate a hidden typealias). That way, you would actually get something called `Float.Sign` and `Double.Sign`, could refer to them as just `Sign` from extensions, etc.

For everything else (i.e. nested types which are not used by requirements and do not capture associated types), they likely have nothing to do with the conforming type. It makes sense then, that everybody should have to spell it as `RandomAccessCollection.Concurrent<Array<Int>>` and not just `.Concurrent`, regardless of whether or not the current context conforms to `RandomAccessCollection`.

We leave the door open for capturing types (when they are possible) to be spelled differently, so you could maybe just write `let myVar: Concurrent` inside an extension of `Array<Int>` in order to get a `RandomAccessCollection.Concurrent where Collection == Array<Int>`.

> At the very least, it is a separate proposal :)

I'm worried that if we leave it, there may be ABI decisions we take now which make it unpalatable later. Even if this part of the proposal can't be implemented right away, it'd be nice to keep it; especially since you can have generic NSObject subclasses, and it's reasonable to want to have nested delegate protocols inside of them.

karwa:
@DougGregor what do you think about the generics change (using unparameterised generic types). I think the example in the 'Detailed Design' part is quite a nice example of how existentials could solve the parameterised protocol issue.

DevAndArtist:
@karwa Please move further discussion and any new draft update to the mailing list. Most feedback should come from the mailing list and not happen in PR comments.

karwa:
@DevAndArtist that will happen during the review

DevAndArtist:
@karwa It's only for the good. Changes you make here are not communicated with the rest of the community. From my point of view, the the proposal during the review would reflect mostly your personal standpoint.

Some quote from older emails of mine:
> *How long does the proposal needs to be in a Pitch state before we can submit a PR?*
> @lattner: It needs to be in pitch state long enough to gain **consensus with the community**.

dabrahams:
Just a note of encouragement: I keep finding cases where I'd want to use this feature in the standard library.  It can't come soon enough!

dabrahams:
One thing I don't see in here is any discussion of where qualification might be unneeded.  For example

```swift
protocol Collection {
  protocol Segmentation { }
}

struct Array : Collection {
  struct Segments : Segmentation { } // LEGAL?
}

protocol BidirectionalCollection : Collection {
  struct DefaultSegmentation<B: BidirectionalCollection> : Segmentation { // LEGAL?
  }
}
```

It would be great to cover that stuff explicitly.

karwa:
@dabrahams Yes, those should both be legal. It's the same as the `Scrollable.Delegate` example in the proposal. I didn't go in to unqualified lookup in more detail because I couldn't find much to describe what is _currently_ legal or not.

But so long as it means there is no externally-visible `Array.Segmentation` protocol, and that plain `Segmentation` only makes sense inside a source context which conforms to `Collection`, I think that's fine.

DougGregor:
The core team discussed this PR for Swift 4 stage 2. It's additive to the language and ABI (which would put it out of scope per the goals of stage 2, https://github.com/apple/swift-evolution/blob/master/README.md), except that it could be source-breaking in two ways:

* If the standard library were to adopt the feature, it would become part of the standard library ABI
* It would change the import of Objective-C Delegate/DataSource protocols to make them nested.

In both cases, if we were to make this change at some later point, it would be fairly easy to provide a migration path via deprecated typealiases and (possibly) ABI-symbol renaming tricks. Given that, the scope of the proposal, and the compressed time scale for Swift 4, we are not going to consider this proposal for Swift 4.

kballard:
Can we reconsider this proposal for Swift 5?

djtech42:
any updates on this?

basememara:
Please make protocols more powerful 😢

And the reverse would also contribute to a protocol-oriented world.. being able to nest protocols within concrete types.

ZsoltMolnarrr:
Hey, this feature would be super helpful. Please consider allocating resources for this.

anyongc:
hhjjjn n

PR #553, opened 2016-11-08
Refactor Metatypes

DevAndArtist:
cc @Anton3 @brentdax


Anton3:
Excuse me, but 11 days and no response? Could someone please schedule a review?

DevAndArtist:
cc @lattner

Coeur:
In the proposed solution, can you describe the cases where `T` is a _tuple_ or a _func_?

DevAndArtist:
@Coeur I don't understand. It's already there, search for:

> Structural types follow the subtype/supertype relationships of their constituent types.

Coeur:
@DevAndArtist oh good. I was wondering what would be the type of a tuple depending on both a struct and a class, like `(a: String, b: NSString)`, but now I understand it would be the same construct as of now.

DevAndArtist:
We're currently considering about a small but a meaningful change, stay tuned.

nullpixel1:
Why has this PR not been reviewed yet?

DevAndArtist:
@nullpixel1 Here is a quote from the mailing list:

> AFAICT, the Refactoring Metatypes proposal is mostly independent. I view it as something that is important to look at in phase 2 because it is source-breaking, and if we’re going to do it, it should happen in Swift 4 rather than Swift 5.
>
> Doug

Currently I'm waiting for some feedback about using `Metatype` and `AnyMetatype` instead to get rid of the ambiguity of the custom `Type` type. **Metatype** describes exactly what that type is. Feel free to comment on the mailing list. ;)

nullpixel1:
Oh, for some reason I wasn't in the mailing list. I just resubscribed

DevAndArtist:
Feel free to merge and review the proposal. ;)

DougGregor:
The core team discussed this proposal and decided that we should not bring it forward for Swift 4. The core team noted that it *could* fit into Swift 4 stage 2 based on the "Source-breaking changes" criteria, but we felt that Swift 4 is not a good time to address metatypes. Rather, we should revise metatypes (and potentially break source compatibility) at the point where the new system can provide demonstrable benefits in terms of new functionality, e.g., significantly improved reflection capabilities. Actually tackling reflection is too large for Swift 4, and we do not want to introduce the "pain" of breaking source without a demonstrable "gain" in terms of expressivity or features.

DevAndArtist:
cc @brentdax @Anton3 

@DougGregor Thank you guys for considering the proposal. It's a pity that we'll stuck with `.Protocol` using protocols in generic context for now.

PR #556, opened 2016-11-10
Marking SE-0145 as rejected.

abertelrud:
It will be revised based on community feedback and then be brought back for review.


jtbandes:
Wouldn't "withdrawn" be more appropriate if a revision is planned?

PR #560, opened 2016-11-16
"private" access control visible in extensions

uroskrkic:


lattner:
This should go through the "pitch" phase on swift-evolution.

uroskrkic:
Hi Chris, thanks for your reply. I've just followed the instructions from "process.md" file, and it is never said anything about "pitch" phase. What does it mean? And how can I (or who is responsible) to put this topic into the "pitch" phase? Thank you! Best regards, Uros.

lattner:
"socialize the idea" under https://github.com/apple/swift-evolution/blob/master/process.md

PR #567, opened 2016-12-02
Proposal to move UnsafeMutablePointer.initialize(from:) to  UnsafeMutableBufferPointer

airspeedswift:
Taking silence as assent on the original evolution pitch :) so submitting a PR.

### DIFF
--- a/proposals/0147-move-unsafe-initialize-from.md
+++ b/proposals/0147-move-unsafe-initialize-from.md
@@ -5,20 +5,13 @@
 * Review Manager: TBD
 * Status: **Awaiting review**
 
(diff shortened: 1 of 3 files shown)


proposals/0147-move-unsafe-initialize-from.md
```diff
@@ -0,0 +1,163 @@
+# Feature name
```
BasThomas: You forgot the `Feature name` here.

airspeedswift: Oops thanks! Fixed.

PR #568, opened 2016-12-04
Increasing objective-c and swift interoperability

wh1pch81n:
This proposal aims to increase objective-c's ability to read enum strings written in swift to increase interoperability.


DevAndArtist:
@wh1pch81n please add `.md` extension ;)

DougGregor:
This proposal is additive---it has no impact on source compatibility or ABI stability, so we can consider again in Swift 4 stage 2. Closing the PR for now; please re-open when we open up Swift 4 stage 2.

wh1pch81n:
Are we currently on Swift 4 stage 1? How can we find out?

DougGregor:
We're in stage 1, and I didn't message that well at all. I've updated the README over at

  https://github.com/apple/swift-evolution

so that's more clear. Someone from the core team will send out a message (and update the README) when we open up "stage 2".

wh1pch81n:
Thanks for the speedy response!

What bugs me is that the stage 1 info is in the middle. When we briefly go
to the repo that stuff is usually cut off and must be explicitly opened.

I suggest adding something to the very top of the readme file to make it
easier to know what stage we are in.

Something like this maybe

https://img.shields.io/badge/Swift%204%20stage%201-Open%20to%20requests-brightgreen.svg

I see some open sourced projects using this to quickly communicate the
latest version. We can think of the stages as a kind of version.

What do you think? Should I maybe make a separate pill request to add this
image?

DougGregor:
A PR that adds that---with a link to the discussion of what Swift 4 stage 1 actually is, if possible---would be welcome!

PR #570, opened 2016-12-08
Change spellings to American English variants in SE-0147

jpsim:
Swift Evolution tends to stick with American spellings


airspeedswift:
Oops, right, thanks JP

PR #571, opened 2016-12-14
remove reference to outdated flag in SPM

masters3d:
https://bugs.swift.org/browse/SR-3407
"The default behaviour for swift build is to build without tests so the flag is not needed"


lattner:
@ddunbar ?

ddunbar:
The proposals in swift-evolution are part of the historical record, not the current documentation. I don't think we want to try and keep them up-to-date with how the command behaves today versus what it did when the proposal was authored.

PR #572, opened 2016-12-15
Delete Swift 3.0 default attributes from SE-0045

mono0926:
Delete Swift 3.0 default attributes from SE-0045

- `@warn_unused_result`
- `@noescape`


DougGregor:
Looks like a nice cleanup that helps readability for future readers, thanks!

PR #574, opened 2016-12-20
[Status] Disable Jekyll for GitHub Pages

jtbandes:
GitHub Pages apparently changed so that `index.xml` no longer takes precedence over `README.md` as a directory index. Explicitly excluding the README seems to restore the proper behavior.

(Noticed this issue after the site was rebuilt when I committed https://github.com/apple/swift-evolution/commit/ec38c58500cf6189215ee472b3ee87fe9483890e.)


jtbandes:
Taking a step back, I see why this problem occurred. GitHub recently made changes to Pages:  https://github.com/blog/2289-publishing-with-github-pages-now-as-easy-as-1-2-3  A better solution seems to be adding a `.nojekyll` file.

PR #577, opened 2017-01-05
Proposal to limit `@objc` inference.

DougGregor:


jckarter:
Nothing personal, Doug, but this should probably wait till phase 2.

lattner:
Why?  This is very impactful on source semantics.

DougGregor:
There are a number of such PRs, where they impact source code/semantics. It feels wrong to privilege this one when we've maintained focus on ABI- and standard-library-impacting features in stage 1.

PR #582, opened 2017-01-11
Mark SE-0145 and SE-0082 as implemented

BasThomas:


jtbandes:
Thanks @BasThomas. Please keep them in numerical order. While you're at it, could you update 0082?

@ddunbar is it correct that these are implemented for 3.1 and not 4?

aciidb0mb3r:
Yes these are implemented for 3.1

BasThomas:
Oops, fixed!

BasThomas:
Updated 0082 as well.

PR #585, opened 2017-01-14
Proposal: Enum Case Blocks

timshadel:
Add an optional syntax to declare all code related to a single enum case in one spot. 

- [X] Check prior proposals
- [X] Consider the goals of the upcoming Swift release
- [X] Socialize the idea

**Socializing Results**

* 59 emails exchanged over 2 weeks
* Swift-evolution threads: [Consolidate Code for Each Case in Enum](https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20170102/thread.html#29966), [second week discussion](https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20170109/thread.html#30041)
* I'm grateful for the comments from Tony Allevato (@allevato), Rien (@swiftrien), Robert Widmann, Derrick Ho, Anthon Zhilin, Daniel Leping, Karim Nassar, David Sweeris, Sean Heber, Charles Srstka, Guillaume Lessard, David Waite, and Jay Abbott.

### DIFF
--- a/proposals/NNNN-enum-case-blocks.md
+++ b/proposals/NNNN-enum-case-blocks.md
@@ -1,9 +1,9 @@
 # Enum Case Blocks
 
-* Proposal: SE-XXXX
+* Proposal: SE-NNNN
 * Authors: [Tim Shadel](https://github.com/timshadel)
 * Review Manager: TBD
-* Status: **TBD**
+* Status: **Awaiting review**
 
 ## Motivation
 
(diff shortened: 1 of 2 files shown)


DougGregor:
The core team discussed this proposal and determined that it doesn't fit in the scope of Swift 4: it is purely additive (and could be introduced at any time), is not source-breaking (so it doesn't need to be front-loaded), and does not affect ABI or the standard library.

PR #586, opened 2017-01-18
Complete SE-0080

ultramiraculous:


jtbandes:
Just tested the 1/17 snapshot from https://swift.org/download/#snapshots; it looks like Float->Int is still not there, although the changes appear in https://github.com/apple/swift/commits/swift-3.1-branch/stdlib/public/core/FloatingPointTypes.swift.gyb. Perhaps the next snapshot will have it.

PR #591, opened 2017-01-20
Default generic arguments

srdanrasic:
This proposal introduces default generic arguments.

Discussion thread: https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20170123/030645.html


DougGregor:
The core team discussed this proposal and determined that it does not meet the goals for inclusion in Swift 4: it's purely additive, has no ABI impact, is not source-breaking, is a nontrivial feature for a relatively small audience, and has no current standard library impact---and, if it did, it's something we could stage in later without breaking the library ABI.

PR #593, opened 2017-01-23
Normalize Enum Case Representation

dduan:
JIRA: https://bugs.swift.org/browse/SR-3542
Discussion: https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20170116/030477.html


dduan:
Anything I can improve here before merging? @jckarter

PR #597, opened 2017-02-07
Submit "Package Manager Tools Version" proposal for evolution review

rballard:


krilnon:
Would you mind updating the value of the `Status` field to be \*\***bold**\*\* on this and #596? It'll fit in with the style of the rest of the proposals better that way.

abertelrud:
I'll make any adjustments to make it match the template.

PR #598, opened 2017-02-08
Protocol-oriented integers (again)

moiseev:


proposals/0000-even-more-improved-integers.md
```diff
@@ -0,0 +1,1107 @@
...
+Major changes to the APIs introduced by this proposal (as compared to
+[SE-0104][se104]) are listed in a [dedicated section](#whats-new-since-se-0104).
+
+## Motivation
+
+Swift's integer protocols don't currently provide a suitable basis for generic
+programming. See [this blog post](http://blog.krzyzanowskim.com/2015/03/01/swift_madness_of_generic_integer/)
+for an example of an attempt to implement a generic algorithm over integers.
+
+The way the `Number` protocol is defined, it does not generalize to floating
```
benrimmington: I think this should be `Arithmetic` instead of `Number`, because it refers to the existing Swift 3 protocol.

moiseev: Good catch. I believe it's a series of renames here that lead to this. Should have been `IntegerArithmetic`, as this is the one that actually exists in Swift 3. Thanks!

proposals/0000-even-more-improved-integers.md
```diff
@@ -0,0 +1,1107 @@
...
+  /// created from a value of `21.0`, while the attempt to initialize the
+  /// constant `y` from `21.5` fails:
+  ///
+  ///     let x = Int(exactly: 21.0)
+  ///     // x == Optional(21)
+  ///     let y = Int(exactly: 21.5)
+  ///     // y == nil
+  ///
+  /// - Parameter source: A floating-point value to convert to an integer.
+  init?<T : FloatingPoint>(exactly source: T)
```
benrimmington: Can this initializer be moved to the `Number` protocol? Or will it be duplicated in the `BinaryFloatingPoint` or `FloatingPoint` protocols?

moiseev: I found your email (that was never replied to, sorry) with the same question. Will reply in the evolution mailing list.

proposals/0000-even-more-improved-integers.md
```diff
@@ -0,0 +1,1130 @@
...
+  and `SignedNumber`.
+
+* `minimumSignedRepresentationBitWidth` property was removed.
+
+* `word(at:)` methods was replaced by `var words: Words` where `Words` is a new
+  associated type.
+
+  This will help avoid quadratic complexity in algorithms iterating over all words, and also allow standard library to provide a default conformance to `Hashable`.
+
+* `trailingZeros` property was added to the `BinaryInteger` protocol.
```
xwu: Are we now renaming this `trailingZeroBits` (ditto below for `leadingZeros`)?

moiseev: Good catch. I changed the prototype, but not the proposal. 🤦‍♂️

PR #599, opened 2017-02-13
[Status] Use Apple system font when available

jtbandes:


krilnon:
Alright; this looks good too. Thank you!

PR #602, opened 2017-02-15
[Status] Add styling for <code> elements

jtbandes:
⚠️ Do not merge yet!

@krilnon This is a work in progress — I faked the `<code>` elements by parsing words that contain camel case or the chars `.(:)` out of titles. (If `` ` ``s in titles were retained, we could simply parse those instead. If you want to move forward with this, I can change the regex to match backtick pairs, and we can merge it, then update the JSON file at your leisure.)

The point of this initial PR is to allow us to iterate on the design. Right now it's mostly copied from my original status page. Here's how it looks:

http://jtbandes.github.io/swift-evolution/

screenshot

### DIFF
--- a/index.css
+++ b/index.css
@@ -59,6 +59,19 @@ ul {
     padding: 0;
 }
 
+code {
+    font-family: "SFMono-Regular", Menlo, Consolas, monospace;
+    font-size: 90%;
+    padding: 0.1em 0.3em;
+    color: #666;
+    background-color: #f7f7f7;
+    border-radius: 3px;
+    word-break: break-all;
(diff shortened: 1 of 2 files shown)


krilnon:
Hey @jtbandes,

We considered the treatment of code voice in headings during our design process and decided against it, which is why the backquotes aren't present in the raw proposal titles. It tends to be more difficult to parse code voice in headings as a reader, and you might have noticed that we don't use it in headings in our developer documentation.

Proposal titles also go out unformatted in things like the review announcement emails, so a good title will be readable without special formatting.

PR #608, opened 2017-02-17
Stage-2 Proposal: Single expression optional `return`

DevAndArtist:


DougGregor:
The core team discussed this PR and determined that it is out-of-scope based on the goals of Swift 4 stage 2, which specify that we are not accepting proposals that are purely syntactic sugar in the language:

  https://github.com/apple/swift-evolution/blob/master/README.md

DevAndArtist:
@DougGregor looking forward into the future, when do you think would be the best time to bring this proposal back? The focus of Swift 5 will probably be ABI *again*, and it won't accept such proposal either.

DougGregor:
I don't know what to tell you. At some point, Reflection should be an explicit goal for Swift. I don't know if that's Swift 5 or not... it's plausible that it would fit later in the Swift 5 schedule if ABI goes well. Languages can only evolve so fast and in so may directions.

PR #615, opened 2017-02-20
Fix typos

abargh:


abargh:
Sorry... should have finished reading before submitting the last one. ;)

jtbandes:
Maybe we should write a proposal to introduce the `Equitable` protocol 😂

abargh:
😀

PR #617, opened 2017-02-21
Added some more details into *Proposed solution: Compile-time checking*

TorinKwok:
According to the [discussion](https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20170213/032328.html) in *swift-evolution* list, compiler needs an exact rule to make decision on whether a property is subjecting to `NSCopying` semantic.

*from Brent Royal-Gordon <<brent@architechies.com>> via swift-evolution:*

> However, occasionally a constructor might be non-copying on purpose, so we'll need a way to disable the warning. A few possibilities:
>
> * Use `.self` on the value
> * Parenthesize the value
> * Parenthesize the variable

Brent presented an important consideration and three potential solutions in the mail, but at the same time I chose one beyond those, which has been contained in this pull-request.


DougGregor:
Because we ended up going with the "compiler magic" approach, I don't think it's worth applying this PR to expand the discussion of the road not taken. If you disagree, feel free to rebase and re-open.

PR #624, opened 2017-02-28
Proposal to limit `@objc` inference.

DougGregor:


BasThomas:
I think you've been using an old template here, as the Rationale does not have its own header anymore?

PR #629, opened 2017-03-04
Add Scope-Bounded Capabilities proposal

anandabits:


anandabits:
@BasThomas the zip contains a playground that the proposal links to.

proposals/NNNN-scope-bounded-capabilities.md
```diff
@@ -0,0 +1,348 @@
...
+
+With one exception, this proposal will not cause any Swift 3 source to fail to compile but will produce different behavior in four cases.  In all cases a mechanical migration is possible.
+
+### `fileprivate`
+
+This proposal deprecates `fileprivate`.  This keyword should be supported for at least one release with a deprecation warning and eventually removed.
+
+### `private`
+
+This proposal reverts the meaning of `private` to file scope.  This will not break any code but many users will want to convert existing uses of `private` to `scoped` during migration to Swift 4.
```
JustinJiaDev:
```
class Foo {
    private bar() { ... }
}

extension Foo {
    private bar() { ... }
}
```

Will break.

anandabits:
Thanks for pointing that out @JustinJiaDev.  I have updated the proposal to clarify this.

CodaFi:
With @JustinJiaDev's comments, I think you should go back and consider the full source-breakage that will result from this proposal.  Now that it is not additive, the stakes for its inclusion go up significantly.

anandabits:
@CodaFi I copied the latest template which includes a "Source compatibility" section.  Isn't that the same as "Impact on Existing Code"?

All drafts of this proposal have explicitly acknowledged the source breakage that will occur, most of which is a change in semantics rather than fail-to-compile.  All breakages will not be affected in Swift 3.1 compatibility mode and are easy to mechanically migrate.  

I also believe that all of the things this proposal breaks are things that we *want* to break eventually regardless of what the final state looks like.  Swift 4 is the right time to have this discussion.  If it were merely an additive change Swift 4 would not be the right time to have this discussion.

If you believe there are cases of source breakage which are not explicitly covered please at least point in the right direction.

anandabits:
Out of scope for Swift 4.

PR #630, opened 2017-03-05
Reformatted quotes in 0115

ctxppc:
The use of code blocks for quotes was confusing, especially given that the quotes were so long in this proposal, so I applied block-quote formatting. I also pushed the quoted code into those block-quotes to make it clear that they're still part of the quote and moved an "author's note" after a code example that directly followed the text explaining it.


DougGregor:
Yeah, this looks clearer to me. Thanks!

PR #633, opened 2017-03-08
Increasing objective-c and swift interoperability

wh1pch81n:
Reopening a previous Pull request #568 
This proposal describes allowing enums to be made available to objective-c as global NSStrings.


DougGregor:
The core team discussed this proposal and determined that it does not fit the goals of Swift 4 stage 2. It is purely additive, in that it does not affect existing parts of the language (it adds new capabilities), and doesn't fit into the "themes" of the Swift 4 release carved out for purely additive changes.

PR #635, opened 2017-03-09
SE-0155 Revision 2

dduan:
This is the revised version of SE-0155


DevAndArtist:
Is that a typo?

> ... therefore the expected syntax **in** invalid.

Should it not be
>... therefore the expected syntax is invalid.

dduan:
@DevAndArtist thanks for the catch!

dduan:
I'm closing this temporarily to incorporate some (emerging) results from SE discussions.

dduan:
This is ready for another round of review. Unless you found anything that still need improvements. @rjmccall @jckarter

jckarter:
LGTM. @rjmccall, want to start another round of review?

PR #652, opened 2017-03-21
[Proposal] extend API for managing tail-allocated buffers

Interfere:
This proposal is an attempt to extend API for creating and managing tail-allocated buffers.

### Draft Proposal Thread
[https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20170313/033933.html](https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20170313/033933.html)

Thanks


DougGregor:
The core team discussed this proposal and felt that it was far too early to commit to any particular API in this area. There are known problems with the implementation of these built-ins, so the gap between "trick we can use in the standard library and validate" and "general feature we can expose as API" is too large to consider this now.

PR #654, opened 2017-03-23
Example typo fix

voidref:
I am fairly certain the example meant to adhere `C` to the `P` protocol as well as inheriting from `NSObject`.


dduan:
would you mind updating the `respondToSelector` syntax as well? It should be `responds(to: "bar")`

voidref:
no problem!

voidref:
This still generates a warning, but I think we can live with it, as an example.

dduan:
thanks @voidref !

PR #655, opened 2017-03-23
Proposal for explicit toll-free conversions

CodaFi:
Proposal text for explicit toll-free conversions.


DougGregor:
The core team discussed this proposal today and determined that it's not in scope for Swift 4. Eliminating these implicit conversions should only be considered in the context of a larger discussion about CF bridging, and is too large to tackle at the moment.

(Sorry I solicited a proposal that the core team considered out of scope)

PR #662, opened 2017-03-29
Initial draft of String proposal #1

airspeedswift:


benrimmington:
The original [SE-0134](https://github.com/apple/swift-evolution/blob/aea8b836d21051076663c5692ec1d09bb3222527/proposals/0134-rename-string-properties.md) proposal recommended that `nulTerminatedUTF8` should be renamed to `nullTerminatedUTF8`.

> I found a lot of nul terminated strings in the text. Is that a typo and shouldn't it be null terminated strings?
> 
> Nevermind, it's the ASCII name for the null character.

proposals/0161-StringRevision1.md
```diff
@@ -0,0 +1,342 @@
+# String Revision: Collection Conformance, C Interop, Transcoding
+
+* Proposal: [SE-0161](0161-StringRevision1.md)
```
benrimmington: `0161` is already taken (by the *Smart KeyPaths* proposal). All existing filenames are lowercased.

```diff
-* Proposal: [SE-0161](0161-StringRevision1.md)
+* Proposal: [SE-NNNN](NNNN-string-revision-1.md)
```

PR #664, opened 2017-03-31
Add all algorithm to Sequence

airspeedswift:


siavashalipour:
👍

DougGregor:
The core team talked about this, and while it is small... there is still much bike shedding to be done to get to agreement on the right name for this algorithm, and we still have a number of higher-priority proposals in flight. So, we should revisit this post-Swift 4.

PR #667, opened 2017-04-03
Add Package Manager Custom Targets Layout proposal

aciidb0mb3r:


benrimmington:
<https://github.com/apple/swift-evolution/blob/master/proposals/0162-package-manager-custom-target-layouts.md>

PR #671, opened 2017-04-05
Key path literal syntax

michael-lehew:
Update proposal to reflect new ideas regarding short-hand key-path references.

### DIFF
--- a/proposals/0161-key-paths.md
+++ b/proposals/0161-key-paths.md
@@ -28,7 +28,7 @@ Making indirect references to a properties' concrete types also lets us expose m
 We would also like to support being able to use _Key Paths_ to access into collections and other subscriptable types, which is not currently possible.
 
 ## Proposed solution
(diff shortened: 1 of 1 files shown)


michael-lehew:
That's fair -- the short version of \ vs # is that the core team wants to save # for future use. It's heritage with macros in C derived languages is appealing, and even as # has been used thus far in swift there is a slight 'perform verb X' aspect to it (#keyPath -> does something returns a string, #available -> does something returns a bool);   #Person.foo.blah could maybe construed that way... except the 'does something'  is a lot more like 'does nothing'.   \ on the other hand has existing precedent in swift of meaning 'behave differently for a moment' (as in string interpolation).  I can tighten that prose up and update the PR if it makes sense.

anandabits:
That's a fair explanation of the rationale.  It makes perfect sense to me.  If this goes through maybe we can get a companion through as well for unbound methods.  Thanks for your work on this proposal!  I'm really excited to see this happen.  :)

michael-lehew:
Thanks Matt, I think that helps considerably.

anandabits:
You're welcome!  This is shaping up to be a really exciting proposal!  Thank you.

PR #675, opened 2017-04-08
Add remove(where:) to the Standard Library

airspeedswift:


hartbit:
I dearly missed this yesterday. Any plans on getting it through review?

Coeur:
@airspeedswift

* Preference would go for symmetry naming with `filter`: after all, it is not named `filterAll`. So a simple `remove` will do.

* Introducing a mutating implementation for the removal job to be optimal means we could also introduce a mutating variant of `filter` with optimized results.

For the two reasons above, [following conventions](https://swift.org/documentation/api-design-guidelines/#name-according-to-side-effects), I would go with:
```
// new in swift 4
public mutating func remove(_ isExcluded: (Element) throws -> Bool) rethrows
// new in swift 4
public mutating func filter(_ isIncluded: (Element) throws -> Bool) rethrows
// new in swift 4
public func removed(_ isExcluded: (Element) throws -> Bool) rethrows -> [Element]
// renamed from swift 3 `filter`
public func filtered(_ isIncluded: (Element) throws -> Bool) rethrows -> [Element]
```

tkremenek:
We discussed this in the Core Team meeting and this needs an implementation to be considered for review.

proposals/NNNN-RemoveWhere.md
```diff
@@ -0,0 +1,134 @@
+# Adding in-place `remove(where:)` to the Standard Library
+
+* Proposal: [SE-NNNN](NNNN-filename.md)
+* Authors: [Ben Cohen](https://github.com/airspeedswift)
+* Review Manager: TBD
+* Status: **Awaiting review**
```
dabrahams: This should include a link to the implementation PR I think.

PR #678, opened 2017-04-13
Ternary comparison proposal

airspeedswift:


proposals/NNNN-ComparisonReform.md
```diff
@@ -0,0 +1,531 @@
...
+  case orderedSame = 0
+  case orderedDescending = 1
+}
+```
+
+
+### `Comparable`
+
+Comparable will be changed to have a new ternary comparison method: `compare(_
+other: Self) -> ComparisonResult`. `x.compare(y)` specifies where to place x
```
JadenGeller: What is the motivation for an instance method over a static method? I realize that `NSObject` sets a precedent here, but I'd like it to be explained before we copy that design.

I personally feel that both the left-hand side and the right-hand side of a comparison should be on equal grounding. An instance method privileges the left-hand argument with dynamic dispatch. Maybe this is desirable behavior for classes? I'm not convinced.

I would prefer the following design of `static`/`class` function as well as an additional top-level function:
```swift
protocol Comparable {
  static func compare(_ lhs: Self, _ rhs: Self) -> ComparisonResult
}

fun compare<T: Comparable>(_ lhs: T, _ rhs: T) -> ComparisonResult {
  return T.compare(lhs, rhs)
}
```

This design removes the asymmetry since `compare` is no longer an instance method.
```swift
switch compare(a, b) { ... }
```

Notably, this design results in static dispatch by default when comparing two values since `T` will be unified to the closest ancestor class of the arguments. If dynamic dispatch is desired (potentially for certain OOP designs), the implementor can delegate to an appropriate subclassable method within `compare`.

proposals/NNNN-ComparisonReform.md
```diff
@@ -0,0 +1,531 @@
...
+that's being done, `<` will be more efficient in many cases (if only because
+it's easier for the optimizer).
+
+For backwards compatibility reasons, `compare` will have a default
+implementation defined in terms of `<`, but to enable only using `compare`,
+`<` and `==` will also have default implementations in terms of `compare`.
+
+The compiler will verify that either `compare`, or `<` and `==`, are provided
+by every type that claims to conform to `Comparable`. This will be done in some
+unspecified way unavailable outside the standard library (it can be made
```
JadenGeller: I'm slightly uncomfortable giving magic behavior to standard library protocols, but I agree that this is the behavior we want…

proposals/NNNN-ComparisonReform.md
```diff
@@ -0,0 +1,531 @@
...
+
+func printEqual<T: FloatingPoint>(_ x: T, _ y: T) {
+  print(x == y)
+}
+
+printEqual(0, 0)           // true (integers use `<T: Equatable>` overload)
+printEqual(nan, nan)       // false (floats use `<T: FloatingPoint>` overload)
+```
+
+As a result of this change, hashing of floats must be updated to make all NaNs
```
JadenGeller: This seems potentially problematic.

I have a radical question: Should `FloatingPoint` types even be `Hashable` and `Equatable`? It seems to me that using `hashValue` or `==` on a `Float` is almost always the wrong thing to do.

Radical design, if only just to be refuted:
```swift
protocol Orderable {
  func <(lhs: Self, rhs: Self) -> Bool
}

protocol Equatable {
  func ==(lhs: Self, rhs: Self) -> Bool
}

protocol Comparable: Orderable, Equatable {
  // Requires total ordering
  func compare(lhs: Self, rhs: Self) -> ComparisonResult
}
extension Comparable {
  func ==(lhs: Self, rhs: Self) -> Bool {
    return compare(lhs, rhs) == .same
  }
    
  func <(lhs: Self, rhs: Self) -> Bool {
    return compare(lhs, rhs) == .ascending
  }
}
```
Here, it is not necessary that an `Orderable` type (like `Float`) have total ordering semantics. Instead, `Comparable` adds this requirement as well as `Equatable` inheritance. `Float` could provide functions like `Float.compare(a, b, semantics: .level2)` for when comparisons are necessary, but they would not be part of the protocol.

I realize this is all probably too source-breaking to even consider, but I thought I'd ask since I'm wondering what your thoughts are on this.

proposals/NNNN-ComparisonReform.md
```diff
@@ -0,0 +1,531 @@
...
+This must be implemented before ABI stability is declared.
+
+## Effect on API resilience
+
+N/A
+
+## Alternatives Considered
+
+
+### Spaceship
```
JadenGeller: I'd support adding this operator even if it simply delegated to `compare`.

proposals/NNNN-ComparisonReform.md
```diff
@@ -0,0 +1,531 @@
...
+just lead to the problems Swift has today.
+
+### Different Names For `compare` and `ComparisonResult`
+
+A few different variants for `ComparisonResult` and its variants were considered:
+
+* Dropping the `ordered` part of `ComparisonResult`'s cases e.g. `.ascending`
+* Naming of `ComparisonResult` as `SortOrder`
+* `enum Ordering { case less, equal, greater }` ([as used by Rust](https://doc.rust-lang.org/std/cmp/enum.Ordering.html))
+* Case values of `inOrder`, `same`, `outOfOrder`
```
JadenGeller: I agree `inOrder` is a bad name since it doesn't specify *which* order (i.e. ascending, descending)

proposals/NNNN-ComparisonReform.md
```diff
@@ -0,0 +1,531 @@
...
+compatibility consequences of introducing naming changes to an existing type
+seems of insufficient benefit.
+
+The method `compare(_:)` does not fully comport with the API naming guidelines.
+However, it is firmly established with current usage in Objective-C APIs, will
+be fairly rarely seen/used (users will usually prefer `<`, `==` etc), and
+alternatives considered, for example `compared(to:)`, were not a significant
+improvement.
+
+### Add Overloads for `(T, T) -> ComparisonResult`
```
JadenGeller: I would strongly support replacing `(T, T) -> Bool` with `(T, T) -> ComparisonResult` variants—not simply supplementing. Here's why:

1) The `areInIncreasingOrder` variant is confusing at the call-site since the definition must be consulted to determine which order the comparison expects.
2) The `areInIncreasingOrder` implementation is very dirty when an `==` result is desired. This is especially bad if we expect other authors to mirror this API:
```swift
if !areInIncreasingOrder(a, b) && !areInIncreasingOrder(b, a) {
  // equal!
}
```
Not only is this unintuitive, but it is also less performant in many cases.

proposals/NNNN-ComparisonReform.md
```diff
@@ -0,0 +1,531 @@
...
+
+## Future Work
+
+This section covers some topics which were briefly considered, but were
+identified as reasonable and possible to defer to future releases. Specifically
+they should be backwards compatible to introduce even after ABI stability. Two
+paths that are worth exploring:
+
+
+### Ergonomic Generalized Comparison for Keyed Containers
```
JadenGeller: I'm hoping some sort of "anonymous protocol conformance" feature could solve this.

proposals/NNNN-ComparisonReform.md
```diff
@@ -0,0 +1,531 @@
...
+
+
+### ComparisonResult Conveniences
+
+There are a few conveniences we could consider providing to make
+ComparisonResult more ergonomic to manipulate. Such as:
+
+```swift
+// A way to combine orderings
+func ComparisonResult.breakingTiesWith(_ order: () -> ComparisonResult) -> ComparisonResult
```
JadenGeller: Oooh, I like this! (though we need to bikeshed the naming…)

proposals/NNNN-ComparisonReform.md
Coeur: Also, static syntax let you potentially compare Optionals where the left argument would be nil. Or at least it makes it easier to mimic the static pattern when you want to implement support for comparison with `nil`.

proposals/NNNN-ComparisonReform.md
```diff
@@ -0,0 +1,531 @@
...
+  case orderedAscending = -1
+  case orderedSame = 0
+  case orderedDescending = 1
+}
+```
+
+
+### `Comparable`
+
+Comparable will be changed to have a new ternary comparison method: `compare(_
```
Coeur: I'm fine with _ternary comparison_. But if you look for an alternative, maybe _3-sided comparison_, which can also be worded as _triangle comparison_.

airspeedswift:
Closing for now, may be revisited before Swift 5.

benrimmington:
I don't know if this is relevant to Swift, but the C++ `operator <=>` proposal uses a different return type for each comparison category (`strong_ordering`, `weak_ordering`, `partial_ordering`, `strong_equality`, `weak_equality`).

<http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0515r1.pdf>

PR #679, opened 2017-04-13
Fix GitHub capitalization

getaaron:


CodaFi:
While you're fixing this nit, please fix SE-0105 and SE-0041 too.

PR #681, opened 2017-04-14
Flexible Access Control Scoping

erica:
This proposal introduces flexible access control scoping to Swift. It offers an underlying additive architecture that mitigates Swift scoped access issues. For example, a Swift coder should not have to mark a symbol `internal` so it can be accessed in just one other file by a targeted consumer, delegate, or data source. The proposal retains Swift 3 keywords and expands scoping to match the level of customization found in other parts of the Swift programming language. 

This proposal follows on to [SE-0169](https://github.com/apple/swift-evolution/blob/master/proposals/0169-improve-interaction-between-private-declarations-and-extensions.md). SE-0169 proposed a minimal workaround to improve the less than ideal "access control situation we have today, while leaving open this direction for future discussion." 

cc: @lattner @tkremenek @DougGregor @hartbit  @jeffreybergier

### DIFF
--- a/proposals/zzzz-scoping.md
+++ b/proposals/zzzz-scoping.md
@@ -1,0 +1,292 @@
+# Flexible Access Control Scoping
+
+* Proposal: TBD <!-- [SE-NNNN](NNNN-filename.md) -->
+* Authors: [Erica Sadun](https://github.com/erica), [Jeffrey Bergier](https://github.com/jeffreybergier)
+* Review Manager: TBD
+* Status: TBD <!-- **Awaiting review** -->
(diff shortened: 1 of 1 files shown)


DougGregor:
SE-0169 has been [accepted](https://github.com/apple/swift-evolution/blob/master/proposals/0169-improve-interaction-between-private-declarations-and-extensions.md), and the core team isn't going to consider further revisions to access control in Swift 4. Specific to this proposal, the core team felt that adding more levels of access control would not be a good tradeoff in complexity vs. expressiveness.

PR #682, opened 2017-04-14
Update 0163-string-revision-1.md

airspeedswift:
PR with two additional methods discussed during the review.

### DIFF
--- a/proposals/0163-string-revision-1.md
+++ b/proposals/0163-string-revision-1.md
@@ -86,8 +86,8 @@ a `const char *`).
 
 The C string interop methods will be updated to those described
 [here](https://github.com/apple/swift/blob/master/docs/StringManifesto.md#c-string-interop):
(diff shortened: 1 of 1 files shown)


proposals/0163-string-revision-1.md
```diff
@@ -145,6 +149,14 @@ C string interop will be consolidated on the following methods:
 
 ```swift
 extension String {
+  /// Constructs a `String` having the same contents as `codeUnits`.
+  ///
+  /// - Parameter codeUnits: a collection of code units in
+  ///   the given `encoding`.
+  /// - Parameter encoding: describes the encoding in which the code units
+  ///   should be interpreted.
+  init<C: Collection, Encoding: UnicodeEncoding>(codeUnits: C, encoding: Encoding)
```
LuizZak: Missing a constraint over `C`?

airspeedswift: Thanks for spotting this.

PR #684, opened 2017-04-16
Nested extensions

tinoheth:
Too little, too late... but afaics, the attached design received less opposition than all other proposals that try to change the Swift access rules (well, that's probably because it was largely ignored).

If nothing else, this PR will teach me something about how Swift evolution really works, so I've to apologise in advance for all formal mistakes my contribution might have.

I would like to give thanks to several commenters on swift-evolution for their encouragement and constructive criticism - I'm just leaving out specific names in the proposal because I fear that would look like a cheesy attempt to raise their support.
After all, being mentioned by me isn't such a big honour - especially as it's not that unlikely that all my scribble is discarded quietly ;-)

Happy easter!
Tino

### DIFF
--- a/zzzz-nesting.md
+++ b/zzzz-nesting.md
@@ -71,6 +71,6 @@ None known
 
 ## Alternatives considered
 
-SE-0169, SE-0159, renaming `fileprivate` and/or `private`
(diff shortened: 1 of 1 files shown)


DougGregor:
Per the core team's [acceptance of SE-0169](https://lists.swift.org/pipermail/swift-evolution-announce/2017-April/000357.html), we're not going to be considering new proposals in this area for Swift 4.

PR #685, opened 2017-04-17
Updated description for 0168-multi-line-string-literals.md

johnno1962:
@jckarter here is the updated description for this proposal revised to provide more detail and as a result of discussion on the swift-evolution thread. This should be the bassis of any second review phase if required. There is a complete prototype implementation submitted as PR https://github.com/apple/swift/pull/8813 if there is a concern finding the time to implement it. The changes were not that great in the finish.


johnno1962:
Restarting pull request with fresh fork due to mistaken merge.

PR #689, opened 2017-04-18
proposals/0161: close one unclosed markdown `...

floatingatoll:
This sentence was missing both the closing `` *and* the period terminating the sentence, so it is not certain that the sentence was meant to end with `KeyPath`. I provide this PR assuming it was meant to, but the authors should verify their intentions before accepting my submission.


parkera:
Thanks!

PR #691, opened 2017-04-20
Update per swift-evolution and Swift core team feedback

parkera:


michael-lehew:
This matches my understanding of the decision. LGTM.

PR #696, opened 2017-04-26
Update 0173-swap-indices.md

jdberry:
There seems to be some confusion on the speling of Manifesto ;)


airspeedswift:
Wow, not one but two terrible typos of the same word. Go me.

PR #697, opened 2017-04-27
Update 0104-improved-integers.md

damirstuhec:
Fixes wrong variable name in `static func +=(_ lhs: inout Self, rhs: Self)` example.


CodaFi:
Thanks!

⛵️

PR #710, opened 2017-05-10
Fix sample code in SE-0172

ole:
The sample code given in the Motivation section of SE-0172 does not compile. It calls the wrong method (`index(where:)` instead of `index(of:)`) and the unwrapping of the optional return value is missing.


ole:
@airspeedswift I think this one is for you.

airspeedswift:
oops, thank you!

PR #712, opened 2017-05-10
[Proposal] Allow `#if` to guard switch case clauses

rintaro:
https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20170508/036568.html


rintaro:
Closing.
We don't need full evolution process for this feature.

PR #719, opened 2017-05-23
Update 0166-swift-archival-serialization.md

salmanjamil:
Fixed a grammer mistake. Replaced `This is can be limiting in Swift ...` with `This can be limiting in Swift ...`


itaiferber:
Good catch — thanks!

PR #722, opened 2017-06-07
Create nnnn-always-flatten-the-single-element-tuple.md

SusanDoggie:
new proposal


tkremenek:
The entire direction here needs to be revisited on swift-evolution, as large parts of SE-0110 were reverted from Swift 4.  We haven't made a decision as a community yet on what we should do.  I'm closing this PR, and let's bring the discussion back to swift-evolution.

PR #726, opened 2017-06-21
SE-0110 Status Deferred

Coeur:


proposals/0110-distingish-single-tuple-arg.md
```diff
@@ -1,10 +1,10 @@
...
 
 * Proposal: [SE-0110](0110-distingish-single-tuple-arg.md)
-* Authors: Vladimir S., Austin Zheng
-* Review Manager: [Chris Lattner](http://github.com/lattner)
-* Status: **Implemented (Swift 4)**
-* Decision Notes: [Rationale](https://lists.swift.org/pipermail/swift-evolution-announce/2016-July/000215.html)
+* Authors: Vladimir S., [Austin Zheng](https://github.com/austinzheng)
+* Review Manager: [Chris Lattner](https://github.com/lattner), [Doug Gregor](https://github.com/DougGregor)
+* Status: **Implemented (Swift 4 beta)**, **Reverted (Swift 4)**
+* Decision Notes: [Implementation Rationale](https://lists.swift.org/pipermail/swift-evolution-announce/2016-July/000215.html), [Reversion Rationale](https://lists.swift.org/pipermail/swift-evolution-announce/2017-June/000386.html)
```
benrimmington: Only a *portion* of SE-0110 was rolled back in apple/swift#10027 and apple/swift#10414, so you might want to change to:

```diff
-* Status: **Implemented (Swift 4 beta)**, **Reverted (Swift 4)**
-* Decision Notes: [Implementation Rationale](https://lists.swift.org/pipermail/swift-evolution-announce/2016-July/000215.html), [Reversion Rationale](https://lists.swift.org/pipermail/swift-evolution-announce/2017-June/000386.html)
+* Status: **Deferred**
+* Decision Notes: [Rationale](https://lists.swift.org/pipermail/swift-evolution-announce/2016-July/000215.html), [Additional Commentary](https://lists.swift.org/pipermail/swift-evolution-announce/2017-June/000386.html)
```

A status from [process.md](https://github.com/apple/swift-evolution/blob/master/process.md#proposal-states) is required, so that the proposal appears on the [status page](https://apple.github.io/swift-evolution/). I don't know if the status page can handle more than one review manager.

Coeur: if only one review manager possible, I suggest we keep the active one (Doug Gregor)

krilnon: > I don't know if the status page can handle more than one review manager.

It can't. Do you think it should?

benrimmington: > > I don't know if the status page can handle more than one review manager.
> 
> It can't. Do you think it should?

In this case, I'd keep the original review manager only.

A deferred or rejected proposal will usually have a separate follow-up proposal (with a different review manager).

Would a **Partially Implemented** status be useful (for [SE-0110](https://github.com/apple/swift-evolution/blob/master/proposals/0110-distingish-single-tuple-arg.md), [SE-0148](https://github.com/apple/swift-evolution/blob/master/proposals/0148-generic-subscripts.md) and [SE-0161](https://github.com/apple/swift-evolution/blob/master/proposals/0161-key-paths.md))?

benrimmington:
Thanks, @Coeur.

PR #727, opened 2017-06-26
Pull request: Allow @objc in all property declarations

CharlesJS:
This is a fairly modest pitch, but it would make working with KVO and Interface Builder in Swift rather easier.

### DIFF
--- a/proposals/xxxx-Allow-objc-on-all-property-declarations.md
+++ b/proposals/xxxx-Allow-objc-on-all-property-declarations.md
@@ -1,0 +1,204 @@
+# Introduction
+
+This pitch proposes to allow the @objc keyword on all property declarations, even ones whose type cannot be represented in Objective-C.
+
+# Motivation
+
+You’re thinking, “But that’s crazy. Why would you ever want to do that?” But hear me out:
+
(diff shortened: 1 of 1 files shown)


lattner:
A point of process, but pitches should be emailed to swift-evolution.

CharlesJS:
It was e-mailed:

https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20170605/037269.html

Unfortunately, it was not controversial enough to generate any replies, but since it's a very modest change that would not affect any existing code, I thought I would send it by the core team to see what they thought.

rjmccall:
The Core Team talked about this.  While we understand the value of being able to traffic an arbitrary value through Objective-C, we're concerned about undermining the predictability of the `@objc` annotation.  Programmers who expect a type to be exported in some straightforward, idiomatic way will be surprised to find that the value was instead exported as an opaque value, and under this proposal they will often not realize the problem until runtime testing, especially given the general laxity of `id` type-checking in Objective-C.

For example, there was a conversation on swift-evolution recently in which someone proposed that Int? be exported as a nullable NSNumber*.

This would also limit our ability to find new ways to idiomatically export Swift types because doing so would no longer be additive: an existing declaration being exported this way would undergo a silent runtime semantic change.

CharlesJS:
While I understand these concerns, do they not already apply to the existing bridging mechanism already in place? Does a user not stand to be equally surprised when, for instance, a Swift type is put into an [AnyHashable : Any] dictionary which is then bridged to an opaque type in an NSDictionary? Would adding a new idiomatic bridge for a type not cause a silent runtime semantic change in that case?

rjmccall:
The possibility of being surprised by an opaque value is far lower when the type is already opaque on the Swift side.

PR #737, opened 2017-08-01
Mark some proposals as implemented

natecook1000:
These 6 are all implemented—there are a few others with at least partial/mostly-done implementations, but I left those alone for now.


benrimmington:
These should be **Implemented (Swift 4)** — instead of just **Implemented** — for the [status page](https://apple.github.io/swift-evolution/#?version=4).

lattner:
@DougGregor ?

DougGregor:
Yeah, it should be **Implemented (Swift 4)**. Thanks for doing this, @natecook1000

natecook1000:
Thanks @benrimmington — missed that distinction. (Also, there's only five implemented proposals here, SE-0174 just has its bug attached.)

DougGregor:
Thanks!

PR #742, opened 2017-08-10
Add a Pull Request link to the proposal template

benrimmington:


benrimmington:
I'll close this PR since @DougGregor is already planning to [update the proposal template with a link to the implementation](https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20170807/038827.html).

PR #743, opened 2017-08-10
Update shield image to include start and end dates

benrimmington:


krilnon:
What motivates the change from green to orange?

benrimmington:
> What motivates the change from green to orange?

@krilnon The badges on [swift/README.md](https://github.com/apple/swift/blob/master/README.md) use `brightgreen` for build-passing and `red` for build-failing.

The badge for Swift 4 stage 2 used `brightgreen` with "Open to requests", even after the deadline for new proposals. I wanted the Swift 5 badge to remain accurate after March 1, 2018. So instead of `brightgreen` or `red`, I tried using the same `#EF5138` color as the [Swift.org logo](https://swift.org/assets/images/swift.svg) but the end result was a slightly different color: Swift 5: August 8, 2017 ... March 1, 2018

So then I chose `orange` from the list of named colors on <https://shields.io/>, but maybe `#EF5138` with the linear gradient is close enough? Or perhaps a `lightgrey` for Swift 5: August 8, 2017 ... March 1, 2018 would be better?

tkremenek:
I'd prefer green until March 1.  We can then change it afterwards to wording more appropriate for what we are willing to take at that time.  The reality is that we will likely take changes after March 1 if they are needed refinements to support the goal of the release.  Also, just specifying the date range feels misleading.  The release does not end on March 1; we're just raising the bar on proposals, but nowhere in the new badge does it say "open for requests", it just specifies a date range that is subject to interpretation.

Honestly, I think the current badge strikes a reasonable balance.  The color choice of orange doesn't have any semantic implication, using lightgray is more subtle and makes the badge easy to miss, just specifying a date range doesn't convey much information (and is misleading), and saying something like "open for requests until March 1" would also be misleading as the story is nuanced on when we will accept proposals (per the text later in the README.md).

tkremenek:
Thanks for working on this @benrimmington

PR #750, opened 2017-09-14
SE-184a (Unsafe[Mutable][Raw][Buffer]Pointer overhaul, part 1)

kelvin13:


atrick:
This PR is ready to merge. It only includes the changes that were reviewed by the core team. Everything that was not yet accepted by the core team was stripped out and will go back to evolution as a new proposal.

PR #755, opened 2017-10-28
[status page] Add 4.1 to the language versions

ikesyo:
SE-0075 and SE-0186 are already implemented for Swift 4.1.


krilnon:
Yep, this makes sense to add. Thanks!

PR #758, opened 2017-11-03
Add "SE-0187 Introduce Sequence.filterMap(_:)"

moiseev:


proposals/0187-introduce-filtermap.md
```diff
@@ -0,0 +1,127 @@
...
+where Swift 4 introduced a `String` conformance to the `Collection` protocol.
+The following code used to compile (due to the `flatMap` overload in question).
+
+~~~swift
+func getNames(people: [Person]) -> [String] {
+  return people.flatMap { $0.name }
+}
+~~~
+
+But it no longer does, because now there is a better overload that does not
```
airspeedswift: In this case, I think it does still compile because the type context forces the other overload to kick in. The point about it breaking when done over two statements stands, tho.

moiseev: Well, this is from the universe where we did not add [these overloads](https://github.com/apple/swift/blob/master/stdlib/public/core/StringRangeReplaceableCollection.swift.gyb#L401).

rjmccall:
Please ignore me, I somehow was looking in the wrong directory.

PR #768, opened 2017-11-27
DynamicMemberLookup proposal

lattner:
This proposal introduces a new `DynamicMemberLookupProtocol` type to the standard
library.  Types that conform to it provide "dot" syntax for arbitrary names which are resolved
at runtime.

It is useful for dynamic language interoperability, but also for things like JSON libraries.  Extensive motivation and discussion are in the proposal, an implementation [is ready](https://github.com/apple/swift/pull/13076), and the proposal has been through multiple rounds of the pitch phase.

-Chris


lattner:
I plan to significantly revise this (again) so it isn't worth considering this PR.

PR #771, opened 2017-11-28
S/E Proposal for “Raw” string literals

johnno1962:
Hi Apple,

This is a straightforward proposal to disable escape processing using the \ character by prefixing a string literal by “r" i.e. r”\a \string \with \invalid \escapes". This could be a useful addition when using regular expression libraries or representing windows file paths or swift source as text in programs. This would apply to single and multi-line string literals.

John


amosavian:
Why not use `'` delimiter?

johnno1962:
Thanks @DougGregor.

PR #773, opened 2017-12-08
[Proposal] Add last(where:), lastIndex(where:), and lastIndex(of:) methods

natecook1000:


amosavian:
It's awesome. However I prefer a `options` argument rather than new methods. Something like:

```swift
array.index(of: 10, options: .backward)
```

fnoyanisi:
I think the fact the proposal does not break current ABI / API and provides great flexibility, implementing new methods sounds like a better idea to me than modifying or adding functionality to existing ones.

amosavian:
Neither API nor ABI are not stabilized yet. It's better to discuss about which is more preferable rather than which one won't break api.

I've said I prefer less functions with arguments (with default value of course!) rather than a few dozen of similar functions which does same thing in different manner.

natecook1000:
@amosavian Thanks for that feedback! I added your suggestion to the "alternatives considered" section, so people should see it when the proposal eventually goes under review.

proposals/0000-add-last-methods.md
```diff
@@ -0,0 +1,118 @@
...
+
+- An [alternative approach](https://github.com/apple/swift-evolution/pull/773#issuecomment-351148673) is to add a defaulted `options` parameter to the existing searching methods, like so:
+
+    ```swift
+    let a = [20, 30, 10, 40, 20, 30, 10, 40, 20]
+    a.index(of: 10)                       // 2
+    a.index(of: 10, options: .backwards)  // 6
+    ```
+
+  This alternative approach matches the convention used in Foundation, such as  `NSString.range(of:options:)` and `NSArray.indexOfObject(options:passingTest:)`, but in the opinion of this author, separate functions better suit the structure of the Swift standard library.
```
airspeedswift: Parameterizing the direction is also a compile-time and run-time pessimisation.

rjmccall:
This is now SE-0204.

PR #777, opened 2018-01-04
SE-0192 Non-exhaustive enums: add `unknown` and rename to '@frozen'

jrose-apple:
Revision after the review period (okay, technically it ends tonight). I've incorporated the two main points of review feedback: picking a name, and preserving exhaustivity checking.


NachoSoto:
Love it ❤️ thanks for incorporating these changes!
Two questions (I don't think they're addressed in the doc but let me know if I missed it):
- What happens if a library maintainer adds a new case to a `@frozen` `enum`?
- What happens if a library maintainer changes an `enum` from `@frozen` to non-frozen?

PR #787, opened 2018-02-11
[Proposal] introduce Dictionary.compactMapValues(_:)

d-date:


proposals/0000-introduce-compact-map-values.md
```diff
@@ -0,0 +1,107 @@
...
+        return try self.reduce(into: [Key: T](), { (result, x) in
+            if let value = try transform(x.value) {
+                result[x.key] = value
+            }
+        })
+    }
+}
+```
+
+Comparing with as below, the implementation above is much faster than below by `Dictionary.reduce` use `inout` result  from Swift 4.
```
lorentey: I don't see how `reduce(into:)` would be faster; however, I don't think we need to show an alternative implementation -- this (and the code below) should be removed.

proposals/0000-introduce-compact-map-values.md
```diff
@@ -0,0 +1,107 @@
...
+            }
+        }
+        return transformed
+    }
+}
+```
+
+## Source compatibility
+
+This change is purely additive so has no ABI stability consequences.
```
lorentey: It has no effect on *source compatibility*, not ABI stability.  (But we do need a separate "Effect on ABI Stability" section, too.)

proposals/0000-introduce-compact-map-values.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Effect on API resilience
+
+This change is purely additive so has no API resilience consequences.
+
+## Alternatives considered
+
+We can implement as you see this proposal and add custom extension, but it's as boiler-plate for you in spite of having many usecases and highly useful.
```
lorentey: Rewrite:

> We can simply omit this method from the standard library -- however, we already have `mapValues` and `filter`, and it seems reasonable to fill the API hole left between them with a standard extension.

d-date:
@lorentey Fixed it! Thank you for your kindness😄

d-date:
Implemented https://github.com/apple/swift/pull/15017

proposals/0000-introduce-compact-map-values.md
```diff
@@ -0,0 +1,102 @@
...
+// r5 == ["a": 1, "b": 2]
+```
+
+## Detailed design
+
+Add the following to `Dictionary`:
+
+```swift
+extension Dictionary {
+    public func compactMapValues<T``>(_ transform: (Value) throws -> T?) rethrows -> [Key: T] {
```
lorentey: Typo: backticks after type parameter `T` in func signature.

d-date: @lorentey  Thank you! Fixed it!

DougGregor:
Core team discussed this; it's ready for review pending a pull request against the Swift repository that contains the implementation + tests.

d-date:
Please tell me what we need to implement more.

d-date:
@lorentey Thank you for your suggestion. Could you check my changes?

airspeedswift:
Scheduling for review.

PR #790, opened 2018-02-19
Allow inout as argument label

iattempt:
The [proposal-0031](https://github.com/apple/swift-evolution/blob/master/proposals/0031-adjusting-inout-declarations.md) modified `inout` keyword.

### DIFF
--- a/proposals/0001-keywords-as-argument-labels.md
+++ b/proposals/0001-keywords-as-argument-labels.md
@@ -34,7 +34,7 @@ escaping to use those APIs. For example:
 
 ## Proposed solution
 
-Allow the use of all keywords except `inout`, `var`, and `let` as argument labels. This affects the grammar in three places:
+Allow the use of all keywords except `var`, and `let` as argument labels. This affects the grammar in three places:
 
 * Call expressions, such as the examples above. Here, we have no grammatic ambiguities, because "<keyword> \`:\`" does not appear in any grammar production within a parenthesized expression list. This is, by far, the most important case.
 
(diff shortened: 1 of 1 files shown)


benrimmington:
Hello, @iattempt, and thanks for your pull request.

Firstly, in my opinion, I don't think "historical" proposals should be updated like this. They aren't intended to be the current documentation, as far as I know.

Secondly, your changes in this pull request are incorrect or misleading. Using `inout` (or `let` or `var`) as an argument label still requires backticks. The related proposals ([SE-0003](https://github.com/apple/swift-evolution/blob/master/proposals/0003-remove-var-parameters.md), [SE-0031](https://github.com/apple/swift-evolution/blob/master/proposals/0031-adjusting-inout-declarations.md), [SE-0053](https://github.com/apple/swift-evolution/blob/master/proposals/0053-remove-let-from-function-parameters.md)) were implemented in Swift 3, so it might be too early to remove this restriction.

/cc @DougGregor (author of [SE-0001](https://github.com/apple/swift-evolution/blob/master/proposals/0001-keywords-as-argument-labels.md)).

DougGregor:
Given that this is SE-0001, I don't think it makes sense to just update it in-place. We do occasionally make changes to existing proposals when implementation experience requires it, but this has been in Swift since it went open-source.

So, I think this change should be a new proposal, which should start with a pitch over on forums.swift.org. I also suggest that it shouldn't be limited to `inout`: `var` and `let` should get the same treatment (or we shouldn't make the change at all).

PR #795, opened 2018-03-01
[gardening] Correct the Swift version for which SE-0195 and SE-0196 are implemented

xwu:
Swift 4.2 is a thing! Surprise!


lattner:
Thx!

PR #796, opened 2018-03-03
Rewrote index.js using ES6 standards

arguiot:
Hello, I rewrote the entire `index.js` using ES6 / ES7 standards.

This reduces the file size and improves readability for no performance loss.

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,20 +12,21 @@
 
 'use strict'
 
+
 /** Holds the primary data used on this page: metadata about Swift Evolution proposals. */
-var proposals
+let proposals;
 
 /**
  * To be updated when proposals are confirmed to have been implemented
  * in a new language version.
  */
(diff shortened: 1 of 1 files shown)


krilnon:
Hi @arguiot, I don't think there's a tangible benefit to swapping to ES6 right now. Users of the status page by-and-large don't read the source code, and the size of `index.js` is considerably outweighed by the JSON proposals payload.

If we introduced a transpilation step to this page, which I don't recommend we do, we could minify the JS as well as transpiling it, which would be a more effective way to reduce the size of the source files if that were a pressing issue. (For context, I originally wrote this incarnation of the status page in ES6 and backported it by hand to ES5 to support more browsers.)

Also, you should check out the [contributing.md](https://github.com/apple/swift-evolution/blob/master/CONTRIBUTING.md) document for this repo. In particular, adopt the style rule about not using line-ending semicolons.

We can reevaluate this if and when the status page evolves and browser adoption progresses a bit more.

PR #797, opened 2018-03-03
[Proposal] Windows support

RUSshy:


CodaFi:
Language-level support for additional platforms does not require an evolution proposal.  Windows support has been championed by a number of contributors for a while now, and the best way to make it happen is to join in their efforts to the best of your ability.

airspeedswift:
Hi @RUSshy – this appears to be more of a feature request than an evolution proposal, which isn't the goal of this repository. As @CodaFi notes, there is an active effort to make Swift available for Windows which would greatly appreciate contributions.

PR #799, opened 2018-03-09
Add containsOnly to Sequence

airspeedswift:
Revision of a previous proposal, with a rename to `containsOnly`


BasThomas:
Would adding `containsNone` be out of scope for this proposal?

airspeedswift:
Maybe, though the case is weaker since it's simply `!contains`.

PR #801, opened 2018-03-12
[Proposal] Allow default parameter overrides

jetmax25:


DougGregor:
This has been languishing because the proposed semantics would require a significant change to the Swift ABI, which has been in the process of being locked down, and there was no implementation that demonstrated how well this change would work out in practice. The ABI change comes from the proposed overriding behavior: we would effectively have to add the new entries to a class vtable or a protocol witness table for each default argument, whereas the current implementation of default arguments is caller-side and has no ABI considerations. I'm going to close this out; we can discuss these issues further on the forums to see if there's a way forward that considers the implementation, ABI impact, and semantic impact of this change.

PR #802, opened 2018-03-13
Fix typo in swift-3_0.md

RenanGreca:
Fixed a typo (soure -> source).


krilnon:
Thanks for noticing and fixing this!

PR #804, opened 2018-03-15
added short function name for document.querySelector

uxitten:


uxitten:
if change q to $ is better?

PR #813, opened 2018-03-22
Ownership Annotations

CodaFi:
This proposal adds the `borrow` type annotation.  With it, Swift users will be able to opt in to value-type and ARC optimizations safely, easily, and reliably.  To complement `borrow`, we are also proposing the `consume` type annotation and the `consuming` member function attribute that signal a parameter will be passed with the existing parameter-passing convention.

[Discussion starts here](https://forums.swift.org/t/ownership-annotations/11276).


tkremenek:
The Core Team discussed that the right time to discuss this language addition is when we get to designing move semantics.  We will revisit this topic then.  Thank you for the proposal.

PR #814, opened 2018-03-22
Exclude Function for Array Object

AnanthaKrish:


Marcocanc:
I think there are barely any cases in which you'd need to subtract an array from another. In most cases the order is not important, but we have Set's [subtracting](https://developer.apple.com/documentation/swift/set/1783322-subtracting) for that. If the order is important, you could just use `filter`.

airspeedswift:
Hi @AnanthaKrish – has this been pitched on the [evolution pitch forum](https://forums.swift.org/c/evolution/pitches)? If not, please be sure to open up a pitch thread there before raising a pull request. Sorry if it has and I've missed it!

Note that there are a couple of things that would need to be addressed before this proposal moves forward:

 - it should not be specific to `Array`, it should probably be generalized to a protocol in the collection hierarchy
 - the implementation as shown is quadratic complexity – that is, the time taken increases by n squared by the length of the array. There is normally a better way to solve this problem, by relying on hashing or sorting of the input to ensure something more like linear complexity. But this solution may not be generalizable to something that belongs in the standard library.

PR #818, opened 2018-03-26
Proposal to add offset(of:) to MemoryLayout.

jckarter:


lattner:
Very nice, +1 from me.

PR #825, opened 2018-04-06
Set implementation version of SE-0143 to Swift 4.1

broomburgo:
I simply set the status of SE-0143 to **Implemented (Swift 4.1)**

### DIFF
--- a/proposals/0143-conditional-conformances.md
+++ b/proposals/0143-conditional-conformances.md
@@ -3,7 +3,7 @@
 * Proposal: [SE-0143](0143-conditional-conformances.md)
 * Author: [Doug Gregor](https://github.com/DougGregor)
 * Review Manager: [Joe Groff](https://github.com/jckarter)
-* Status: **Implemented (Swift 4.2)**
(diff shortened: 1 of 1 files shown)


linqingmo:
No, Swift4.1 Swift runtime does not yet support dynamically querying conditional conformance. So Swift4.2 is correct.

linqingmo:
```swift
protocol P {
    func test()
}

extension Array: P where Element: P {
    func test() {
        forEach { $0.test() }
    }
}

class T: P {
    func test() {
        print("Support dynamically querying conditional conformance.")
    }
}

let array = [T(), T()]
array.test() // Swift 4.1 works

func testDynamically(_ a: Any) {
    if let a = a as? P {
        a.test() // Swift 4.2
    } else {
        print("Dose not support dynamically querying conditional conformance.") // Swift4.1
    }
}

testDynamically(array)
```

broomburgo:
Thanks a lot for the clarification, this is huge, I shouldn't have missed it.

linqingmo:
You’re welcome.

PR #829, opened 2018-04-11
Proposal: Collection search, remove & replace

lancep:


proposals/0209-collection-search.md
```diff
@@ -0,0 +1,66 @@
+# Collection search
```
tkremenek: The name of the proposal doesn't cover its actual scope, which includes replacing.  Maybe "Collection search & replace"?

lancep: Might as well throw "remove in there too "Collection search, remove & replace"

thanks!

airspeedswift:
Given the main use case is on `String`, and that there are potentially more efficient implementations specific to `String`, this may need more work to decide if it should be a customization point or similar. It could also be combined with #846. Returning to forums for more discussion.

PR #832, opened 2018-04-13
Update SE-0110's status

BasThomas:
It was accepted in https://lists.swift.org/pipermail/swift-evolution-announce/2016-July/000215.html

### DIFF
--- a/proposals/0110-distingish-single-tuple-arg.md
+++ b/proposals/0110-distingish-single-tuple-arg.md
@@ -3,7 +3,7 @@
 * Proposal: [SE-0110](0110-distingish-single-tuple-arg.md)
 * Authors: Vladimir S., [Austin Zheng](https://github.com/austinzheng)
 * Review Manager: [Chris Lattner](https://github.com/lattner)
-* Status: **Accepted**
+* Status: **Implemented (Swift 4)**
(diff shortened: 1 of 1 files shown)


benrimmington:
@BasThomas, see the [comments in PR 726](https://github.com/apple/swift-evolution/pull/726#discussion_r123172100).

Part of SE-0110 was reverted/deferred, due to a [usability regression](https://lists.swift.org/pipermail/swift-evolution-announce/2017-June/000386.html).

Which of the [proposal states](https://github.com/apple/swift-evolution/blob/master/process.md#proposal-states) would be the most accurate?

BasThomas:
I think "accepted with revisions" might make sense. Otherwise a "partially implemented" status (like you suggested) might be a good alternative. What do you think?

masters3d:
I think deferred is the best status given the supplied link. https://lists.swift.org/pipermail/swift-evolution-announce/2017-June/000386.html

BasThomas:
Ping @DougGregor — what would be the appropriate status?

DougGregor:
This proposal is mostly-but-not-completely implemented. I prefer to leave it in the “accepted” state until we are sure it is done, which we hope will be within the next few months

PR #839, opened 2018-04-22
Update SE-0202 (Random Unification) with the accepted revisions

Azoy:


tkremenek:
I updated the announcement title on the forums.

airspeedswift:
Thanks for the update, @Azoy. This looks good to me.

airspeedswift:
@benrimmington I am going to introduce a follow-up thread on the forums re the name.

PR #844, opened 2018-04-30
[SE-0206] Incorporate acceptance notes; set status to Implemented

lorentey:
- Remove `Hasher.init(seed:)`
- Remove most `Hasher.combine` overloads
- Change `Hasher.finalize` to be `__consuming`
- Clarify that `Hasher` is order-sensitive
- Clarify `hashValue`'s deprecated status (only custom implementations are deprecated)
- Clarify that SE-0206 is *not* going to be source-breaking; `hashValue` impls will compile with a warning in 4.2 mode.
- Update links to implementation
- Move to Implemented status

### DIFF
--- a/proposals/0206-hashable-enhancements.md
+++ b/proposals/0206-hashable-enhancements.md
@@ -3,12 +3,12 @@
 * Proposal: [SE-0206](0206-hashable-enhancements.md)
 * Authors: [Karoy Lorentey](https://github.com/lorentey), [Vincent Esche](https://github.com/regexident)
 * Review Manager: [Joe Groff](https://github.com/jckarter)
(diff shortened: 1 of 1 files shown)


lorentey:
@jckarter This incorporates the changes in the core team's acceptance notes, and also clarifies that while `hashValue` is deprecated as a protocol requirement, this proposal is not source breaking.

The old *Source Compatibility* section implied that users will need to go through a forced migration step, which is not the case. I replaced it with text that describes the actual intended behavior.

apple/swift#16009 and apple/swift#16073 implemented the proposal. (Except for the deprecation warning when `hash(into:)` needs to be derived from a manual `hashValue`.)

PR #848, opened 2018-05-09
[SE-0185] Synthesis in extensions is necessary for conditional conformances.

huonw:
Also, the bugs with Codable in extensions now seem to be fixed, so
maintaining consistency with that (plus, those bugs would've likely
affected Equatable and Hashable) is no longer necessary.

This interaction wasn't fully considered in the original proposal.


DougGregor:
The core team has approved this amendment. Merging.

PR #859, opened 2018-06-11
Fix a Swift 4.2 Filtering Issue in SE-0197 proposal

PaulSolt:
This proposal doesn't appear when you filter by Swift 4.2: https://apple.github.io/swift-evolution/#?version=4.2

I believe the status needs to change to "Implemented (Swift 4.2)"


DougGregor:
Thanks!

PR #866, opened 2018-06-23
Update status of SE-0143

MultiColourPixel:
Corrected the status to indicate that his has been implemented in Swift 4.1 as per https://swift.org/blog/swift-4-1-released/


tkremenek:
Technically it is *fully* implemented in 4.2.  Support for dynamic casting with conditional conformances landed in 4.2.  So while the bulk of the feature arrived in 4.1 (as reported in the blog post) the rest of it landed in 4.2.

PR #872, opened 2018-07-09
Create IfOptionalUnWrappingOperator(>>) proposal

ShenghaiWang:


Kuniwak:
I think it is enough to use [`Optional.map(_:)`](https://developer.apple.com/documentation/swift/optional/1539476-map). Or I want to know the advantage of `>>` if it is not enough.

rafaelGuerreiro:
This approach has some problems that I can think of:
1. It uses closures, which take some time to capture any accessors. I'm not sure this is true for nonescaping closures, probably not.
2. It doesn't offer the ability to do something if the variable is nil. This is one of the greatest things of this construct that many people forget about.
3. It would be very lengthy to do that with more than one variable. Whereas, the way it is, we can simply separate them by comma.

4. (Not sure) would it affect the optimization?

5. (Personal) I don't see `>>` as a sign to unwrap, but to move something into the rhs.

ShenghaiWang:
I think it's more clear semantically and it's in align with Haskell syntax. I will close it since you guys don't like it :)

rafaelGuerreiro:
Maybe you should pitch this idea in the Swift evolution forum: https://forums.swift.org/c/evolution

Kuniwak:
FYI, the operator `<^>` that is defined by [typelift/swiftz](https://github.com/typelift/swiftz) has the same meaning of your `>>`. I feel that typelift/swift is a good choice if you like using operators such as `>>` instead of `Optional.map(_:)`.

PR #877, opened 2018-07-15
Support 'less than' operator in compilation conditions

danielmartin:
Implementation: https://github.com/apple/swift/pull/17960


tkremenek:
This is scheduled for review on August 13.

hartbit:
@danielmartin The proposal doesn't make it clear if `<=` is supported. I can't see it be very useful, but it would be consistent with `>=`.

tkremenek:
@danielmartin I am kicking off this review.

PR #881, opened 2018-07-20
[SE-0220] Fix "proposal contains no content" error

benrimmington:


benrimmington:
The parser for the status page has found an error in this proposal.

<https://data.swift.org/swift-evolution/proposals>

It might be caused by the `count(where:)` title, which is entirely in backticks.

This PR expands the title, to follow the [SE-0032][] and [SE-0207][] proposal titles.

[SE-0032]: <https://github.com/apple/swift-evolution/blob/master/proposals/0032-sequencetype-find.md>

[SE-0207]: <https://github.com/apple/swift-evolution/blob/master/proposals/0207-containsOnly.md>

khanlou:
My vote is to do both. Feel free to change my title, but it seems like the parser should be able to handle this?

benrimmington:
Thanks, @krilnon. I'll close this PR now.

PR #893, opened 2018-08-17
[SE-0225] Use permalinks in examples

benrimmington:


benrimmington:
@robmaceachern I've converted the `/blob/master/` links to permalinks.

PR #895, opened 2018-08-21
Proposal for identity key paths

jckarter:
https://forums.swift.org/t/some-small-keypath-extensions-identity-and-tuple-components/13729

### DIFF
--- a/proposals/ZZZZ-identity-keypath.md
+++ b/proposals/ZZZZ-identity-keypath.md
@@ -1,0 +1,112 @@
+# Identity key path
+
+* Proposal: [SE-ZZZZ](ZZZZ-identity-keypath.md)
+* Authors: [Joe Groff](https://github.com/jckarter)
+* Review Manager: TBD
+* Status: **Awaiting review**
+* Implementation: [apple/swift#18804](https://github.com/apple/swift/pull/18804)
+
+## Introduction
+
+Add the ability to reference the identity key path, which refers to the entire
+input value it is applied to.
+
+Swift-evolution thread: [Some small keypath extensions: identity and tuple components](https://forums.swift.org/t/some-small-keypath-extensions-identity-and-tuple-components/13729)
+
+## Motivation
+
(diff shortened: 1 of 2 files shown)


proposals/ZZZZ-identity-keypath.md
```diff
@@ -0,0 +1,112 @@
...
+print(x.self) // prints 2
+```
+
+By analogy, we could spell the identity key path `\.self`, since it notionally
+refers to this `self` member:
+
+```swift
+let id = \Int.self
+
+x[keyPath: id] 3
```
tinoheth: That line need a "=", doesn't it?

jckarter: Thanks!

proposals/ZZZZ-identity-keypath.md
```diff
@@ -0,0 +1,112 @@
...
+
+By analogy, we could spell the identity key path `\.self`, since it notionally
+refers to this `self` member:
+
+```swift
+let id = \Int.self
+
+x[keyPath: id] = 3
+print(x[keyPath: id]) // prints 3
+```
```
michael-lehew: It might be nice, to include in addition to this example, one that shows it being used on the ViewController example you motivated the proposal with.

proposals/ZZZZ-identity-keypath.md
```diff
@@ -0,0 +1,112 @@
...
+The Swift standard library required some modifications to correctly handle
+identity key paths.
+
+## Alternatives considered
+
+The biggest design question here is the syntax. Some other alternatives
+include:
+
+- The special syntax `\.`, a key path with "no components".
+- A static property on `KeyPath` and/or `WritableKeyPath`.
```
michael-lehew: Where do we vote? :)   I like `\.self`

jckarter: That'd be something to discuss during the formal proposal review.

PR #897, opened 2018-08-22
Amend SE-0224 to include #if compiler as well.

tkremenek:
This is based on feedback in the [review](https://forums.swift.org/t/se-0224-support-less-than-operator-in-compilation-conditions/15213/2).

### DIFF
--- a/proposals/0224-ifswift-lessthan-operator.md
+++ b/proposals/0224-ifswift-lessthan-operator.md
@@ -30,6 +30,10 @@ code:
 #if !swift(>=4.2)
 // This will only be executed if the Swift version is less than 4.2.
 #endif
+
+#if !compiler(>=4.2)
+// This will only be executed if the Swift compiler version is less than 4.2.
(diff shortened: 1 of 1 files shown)


tkremenek:
@danielmartin This amends SE-0224 to include `if compiler`, which your implementation supports.  The plan is to accept SE-0224 with this revision (which is already reflected in the implementation).  Does this look good to you?

benrimmington:
@tkremenek Will the `#if !compiler(>=4.2)` and `#if compiler(<4.2)` conditions always be false?

I've just noticed that the `#if compiler` tests (added by @hartbit for SE-0212) were accidentally removed in apple/swift#17689

danielmartin:
@tkremenek Looks good to me! I added some basic `#if compiler` tests to my implementation to prevent this feature to ever regress.

tkremenek:
@danielmartin That's great.  Thanks!

PR #901, opened 2018-09-03
Add an `adjacentPairs` algorithm to Sequence

mpangburn:


benrimmington:
@mpangburn Can this proposal be closed?

Swift Algorithms now has a similar API: apple/swift-algorithms#20

timvermeulen:
`slidingWindows` from Swift Algorithms requires `Collection`, so it's not entirely the same thing.

tkremenek:
@mpangburn: It seems that this proposal hasn't seen much progress lately. As mentioned above, we may be able to roll this into the larger "Algorithms" package.  Would you be open to including this in the Algorithms package?

mpangburn:
> @mpangburn: It seems that this proposal hasn't seen much progress lately. As mentioned above, we may be able to roll this into the larger "Algorithms" package.  Would you be open to including this in the Algorithms package?

I'll be moving forward with contributing this code to the Algorithms package; this can now be closed out. Thanks for the nudge!

PR #908, opened 2018-09-08
Additional SE-0228 changes requested by @milseman

brentdax:
Makes an example more explicit, and further clarifies that the examples in “Potential uses” are just examples and this isn’t an Apple-only feature.

I can't formally tag @milseman for a review, but let's give him a chance to comment before we merge this.


milseman:
@airspeedswift @DougGregor could you merge this prior to launching the review?

DougGregor:
Sure!

PR #909, opened 2018-09-18
SE-0228 default implementation amendment

brentdax:
Adds a default `init(stringLiteral:)`, plus an unavailable version which is used when the default `init(stringInterpolation:)` for `DefaultStringInterpolation` is available.

This amendment was suggested by @jrose-apple, who found it strange that you would need to implement `init(stringLiteral:)` when `init(stringInterpolation:)` provides a superset of its functionality. @milseman and I agree that it's a good idea.

Not sure if we should merge this now or wait for the core team meeting.

cc @DougGregor


DougGregor:
We're right at the end of the scheduled review period, so I'll bring this up to the core team rather than merging speculatively. Thanks!

PR #914, opened 2018-09-26
SR-8649: Amendment to [SE-0167] to include Range types

dlbuckley:
## Overview
This proposal amendment adds the standard library `Range` types to the `Swift Encoders` proposal.

Range is a very useful type to have conform to `Codable`. A good usage example is a range being sent to/from a client/server to convey a range of time using `Date`, or a safe operating temperature range using `Measurement<UnitTemperature>`.

Due to the range types not being defined in Swift Foundation, I've created a new section specifically for them residing in the Standard Library.

### Forum Discussion
https://forums.swift.org/t/range-conform-to-codable/15552
### Bug report
https://bugs.swift.org/browse/SR-8649
### Pull Request
https://github.com/apple/swift/pull/19532

PR #915, opened 2018-09-27
Codable Range proposal

airspeedswift:
Full proposal version of #914

### DIFF
--- a/proposals/0000-codable-range.md
+++ b/proposals/0000-codable-range.md
@@ -14,6 +14,10 @@ conformance.
 
 Swift-evolution thread: [Range conform to Codable](https://forums.swift.org/t/range-conform-to-codable/15552)
 
+## Motivation
+
+`Range` is a very useful type to have conform to `Codable`. A good usage example is a range being sent to/from a client/server to convey a range of time using `Date`, or a safe operating temperature range using `Measurement<UnitTemperature>`.
+
 ## Proposed solution
 
 The following Standard Library range types will gain `Codable` conformance
--- a/proposals/0000-codable-range.md
+++ b/proposals/0000-codable-range.md
@@ -1,0 +1,35 @@
+# Add Codable conformance to Range types
+
+* Proposal: [SE-NNNN](NNNN-codable-range.md)
+* Authors: [Dale Buckley](https://github.com/dlbuckley)
+* Review Manager: TBD
+* Implementation: [apple/swift#19532](https://github.com/apple/swift/pull/19532)
+* Status: **Awaiting review**
+
+## Introduction
+
+[SE-0167] introduced `Codable` conformance for some types in the standard
+library, but not the `Range` family of types. This proposal adds that
+conformance.
+
+Swift-evolution thread: [Range conform to Codable](https://forums.swift.org/t/range-conform-to-codable/15552)
+
+## Proposed solution
+
+The following Standard Library range types will gain `Codable` conformance
+when their `Bound` is also `Codable`:
+
+ * `Range`
+ * `ClosedRange`
+ * `PartialRangeFrom`
+ * `PartialRangeThrough`
+ * `PartialRangeUpTo`
+
+## Effect on ABI stability, resilience, and source stability
+
+This is a purely additive change, and so has no impact.
+
+## Alternatives considered
+
+None
+
--- a/proposals/0000-codable-range.md
+++ b/proposals/0000-codable-range.md
@@ -35,5 +35,70 @@ This is a purely additive change, and so has no impact.
 
 ## Alternatives considered
 
-None
+One area of concern mentioned during the discussion is that of
+potential data corruption. Consider this: if an application implements its own
+`Codable` conformance for `Range` or `ClosedRange` (which is, by the way, not
+recommended, as both the protocol and the type are defined in the standard
+library), there might be data permanently stored somewhere (in the database, in
+a JSON or PLIST file, etc.) which was serialized using that conformance. Unless
+the serialization format is exaclty the same as the one used in the standard
+library, that data will be considered invalid.
 
+When this proposal is implemented, any `Codable` conformance to the `Range` type
+outside standard library will result in a compiler error as duplicate
+conformance. At which point, we suggest the following course of action to avoid
+data loss.
+
+1. Define your own type wrapping `Range`.
+
+    ```swift
+    public struct MyRangeWrapper<Bound>
+    where Bound: Comparable {
+    public var range: Range<Bound>
+    }
+    ```
+
+2. Port `Codable` conformance from `Range` to this new type.
+
+    ```swift
+    extension MyRangeWrapper {
+      enum CodingKeys: CodingKey {
+        case lowerBound
+        case upperBound
+      }
+    }
+
+    extension MyRangeWrapper: Decodable
+    where Bound: Decodable {
+      public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        let lowerBound = try container.decode(Bound.self, forKey: .lowerBound)
+        let upperBound = try container.decode(Bound.self, forKey: .upperBound)
+        self.range = lowerBound ..< upperBound
+      }
+    }
+
+    ```
+
+    Note that unless you wish to keep using this serialization format in the
+    future, and are OK with using what's provided by the standard library, only
+    the `Decodable` conformance is needed for data migration purposes.
+
+3. Support both old and new formats when deserializing your data that contains
+   ranges.
+
+   ```swift
+    extension JSONDecoder {
+      func decodeRange<Bound>(
+        _ type: Range<Bound>.Type, from data: Data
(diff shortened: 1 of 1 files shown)


proposals/0000-codable-range.md
```diff
@@ -0,0 +1,39 @@
...
+
+* Proposal: [SE-NNNN](NNNN-codable-range.md)
+* Authors: [Dale Buckley](https://github.com/dlbuckley)
+* Review Manager: TBD
+* Implementation: [apple/swift#19532](https://github.com/apple/swift/pull/19532)
+* Status: **Awaiting review**
+
+## Introduction
+
+[SE-0167] introduced `Codable` conformance for some types in the standard
```
benrimmington: `[SE-0167]` should be `[SE-0167](0167-swift-encoders.md)`

dlbuckley:
@airspeedswift is there anything else that I can do to help push this further along?

benrimmington:
@dlbuckley The proposal is described as a "purely additive change", so my guess would be Swift 5.1, but that decision is nothing to do with me.

@airspeedswift [Conditional conformance collisions](https://forums.swift.org/t/conditional-conformance-collisions/18023) might be a problem.

dlbuckley:
@airspeedswift other people having written their own implementations of this is indeed an issue when updating to a newer version of the language with this included.

We actually had an issue recently where we were encoding values of `Measurement` type, and while the encoded data was correct, it was a pain to query in mongo. What we ended up doing was wrapping up all uses of `Measurement` in the Encode/Decode methods where it was used. This allowed us to use our own method of encoding and decoding `Measurement` types.

While that example was a work around, I do wonder if this highlights an extension point in `Encoder/Decoder` that allows for registering custom encode/decode methods for a specific type which then ignores the default `Codable` conformance of the type. Maybe this is something that @itaiferber can chime in with here?

As for the suggestion of first trying to decode the type with an unkeyed unarchiver and then trying again with the keyed unarchiver; I think there are a number of risks with that. Firstly, how do you know which keys that are being used? In my first version I used keys that helped to distinguish between all of the different Range types, but I know other extensions have just kept to the names of the properties (`upperBound` and `lowerBound`). The biggest problem I see is that if the data is read in from the keyed format, and then saved out as the unkeyed format, it could break whatever is relying on that data.

I think we are now past the point where this can be added in without causing a number of headaches. We just need to decide if causing the headaches is worth it, or it's best to leave Swift without the 'batteries-included' implementation of the Range types.

parkera:
We're still in early days of Swift, so I think it would be a mistake to just give up on doing this permanently.

With respect to the extension point on `Encoder` for custom types, it was always our view that the way to do this in general was by making a wrapper struct for your type. I think it's still pretty important to the design of `Codable` to preserve encapsulation as much as possible. We made exceptions for a few types in various encoders like `JSONEncoder` because of the prevailing conventions of data encoded in those formats (e.g., `URL` as a single `String`, in that case, or `Date` stored as ISO8601).

itaiferber:
As far as existing extensions go, I agree with @parkera. It's never been safe to extend others' types with `Codable`, and will never be safe: the only right thing to do is to create wrapper structs around others' types which you control (@dlbuckley mentions with `Measurement`). This is something I've mentioned a lot in discussion, but it's not something we can really force folks to do. (We could technically prevent `extension Foo : Encodable` and `extension Foo : Decodable` if `Foo` is defined outside of the current module, but not without a ton of breakage, and kludge in the compiler.)

So I think we need to consider to what extent we are worried about extensions like this. No matter what we do, adopting `Codable` on the `Range` types is going to be a source-compatibility issue for others. I personally thing that the right thing for them to do at that point is to resolve the source-breakage by going and wrapping their existing `Codable` ranges with the wrapper types that maintain their format. Or, if they really care to, they can keep their overriding implementations of `init(from:)` and `encode(to:)` and those keep working for them within their modules. (I would say this is brittle and pretty wrong, but it is at least easy.)

As for allowing the defeat of `Codable` implementations in general as an extension point on `Encoder`/`Decoder` — we discussed this during the initial design but decided pretty strongly against it; ignoring the design constraints of now forcing this capability on all `Encoder`s and `Decoder`s, there's a fundamental encouragement to breaking encapsulation that that facility offers. We break this already for some types with `JSONEncoder` in highly targeted cases, but even that feels wrong in a sense. (The philosophically-consistent solution might be to instead offer `Codable` wrappers for these types, like `ISO8601CodableDateWrapper` or something.)

In any case, I don't think we can start guessing at others' representations for `Range` and the like — as flexible as `Codable` is, we can't guess at keys used in keyed containers, or how to decode something like `[12, 57, 96, 22]`.

airspeedswift:
> We're still in early days of Swift, so I think it would be a mistake to just give up on doing this permanently.

I agree, and there's also a significant benefit to doing it in 5.0 to avoid having to gate on OS version to use it if we add it in the future.

I think a best efforts approach would be reasonable here. It's actually perfectly acceptable IMO to break code that had conformed std lib types to `Codable` manually. It's not OK to conform a library's types to that same library's protocols for exactly this reason. It's just nice of us not to if possible, depending on the combination of circumstances.

moiseev:
@dlbuckley 
> As for the suggestion of first trying to decode the type with an unkeyed unarchiver and then trying again with the keyed unarchiver; I think there are a number of risks with that. Firstly, how do you know which keys that are being used?

Since we're talking about the range, where the lower bound should be less than or equal than the upper bound, we don't really care about the exact values of the keys. It should be possible to just obtain all the keys, try them, and then assign the smaller value to lower bound. I'm not too comfortable with `Codable` to know for sure if it's at all possible to obtain all the values from a keyed container without specifying the keys, though.

dlbuckley:
> Since we're talking about the range, where the lower bound should be less than or equal than the upper bound, we don't really care about the exact values of the keys. It should be possible to just obtain all the keys, try them, and then assign the smaller value to lower bound. I'm not too comfortable with Codable to know for sure if it's at all possible to obtain all the values from a keyed container without specifying the keys, though.

Unfortunately it's not possible to just obtain all of the keys like that. Due to the nature of Codable you need to specify the keys you expect when getting the keyed container. Using the `allKeys` property on a keyed container only returns all of the keys you give it, not the other way around.

I've updated the PR with a change that was requested by @moiseev last week which hopefully mitigates his concerns around decode error handling.

@itaiferber thanks for chiming in. I totally agree with you and @parkera regarding the reasoning around protecting encapsulation here. I was simply thinking out loud on if there was a way around it to be nice to people who have already gone down the avenue of their own implementations.

itaiferber:
> Unfortunately it's not possible to just obtain all of the keys like that. Due to the nature of Codable you need to specify the keys you expect when getting the keyed container. Using the `allKeys` property on a keyed container only returns all of the keys you give it, not the other way around.

@dlbuckley Just FYI, that's not strictly true — you can get all of the keys in a given keyed container by getting that keyed container with a key type that allows _any_ `String` or `Int` value. [This is how `Dictionary` decodes itself](https://github.com/apple/swift/blob/master/stdlib/public/core/Codable.swift.gyb#L1859), using [_`DictionaryCodingKey`](https://github.com/apple/swift/blob/master/stdlib/public/core/Codable.swift.gyb#L1805)

dlbuckley:
> > Unfortunately it's not possible to just obtain all of the keys like that. Due to the nature of Codable you need to specify the keys you expect when getting the keyed container. Using the `allKeys` property on a keyed container only returns all of the keys you give it, not the other way around.
> 
> @dlbuckley Just FYI, that's not strictly true — you can get all of the keys in a given keyed container by getting that keyed container with a key type that allows _any_ `String` or `Int` value. [This is how `Dictionary` decodes itself](https://github.com/apple/swift/blob/master/stdlib/public/core/Codable.swift.gyb#L1859), using [_`DictionaryCodingKey`](https://github.com/apple/swift/blob/master/stdlib/public/core/Codable.swift.gyb#L1805)

Ah ok, every day is a school day 😁

proposals/0000-codable-range.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Alternatives considered
+
+One area of concern mentioned during the discussion is that of
+potential data corruption. Consider this: if an application implements its own
+`Codable` conformance for `Range` or `ClosedRange` (which is, by the way, not
+recommended, as both the protocol and the type are defined in the standard
+library), there might be data permanently stored somewhere (in the database, in
+a JSON or PLIST file, etc.) which was serialized using that conformance. Unless
+the serialization format is exaclty the same as the one used in the standard
```
benrimmington:
```suggestion
the serialization format is exactly the same as the one used in the standard
```

moiseev: I am now looking for a spell-checking plugin for the editor I used. Thanks, @benrimmington.

tkremenek:
This proposal is now going to be run through a review.  I will merge and patch up the document for review.

Given the updates @airspeedswift and @moiseev have done on this proposal as well, I think it makes sense to add them as proposal authors as well to address any questions related to technical concerns and data compatibility issues.

PR #923, opened 2018-10-03
[Proposal] Collection conditionals SR-8743

johnno1962:
Hi Apple, 

A lightning proposal for a solution to [SR-8743](https://bugs.swift.org/browse/SR-8743) to allow conditionals inside collection literals. Hopefully this will not require a full public review as it is a very minor, intuitively obvious extension to current Swift syntax but this proposal documents the approach and motivation for the Core Team to make their own call on these matters.

Existing PR: https://github.com/apple/swift/pull/19347

### DIFF
--- a/proposals/NNNN-collection-conditionals.md
+++ b/proposals/NNNN-collection-conditionals.md
@@ -1,0 +1,61 @@
+# Conditionals in Collections
+
+* Proposal: [SE-NNNN](NNNN-collection-conditionals.md)
+* Authors: [John Holdsworth](https://github.com/johnno1962)
+* Review Manager: TBD
+* Status: **Awaiting decision**
+* Implementation: [apple/swift#19347](https://github.com/apple/swift/pull/19347)
+* Decision Notes: TBD
+* Bugs: [SR-8743](https://bugs.swift.org/browse/SR-8743)
+
+## Introduction
+
+This is a lightning proposal to extend the existing Swift language slightly to allow `#if` conditional inclusion of elements in array and dictionary literals. For example:
+
+```swift
+let array = [
+	1,
+	#if os(Linux)
+	2,
+	#endif
+	3]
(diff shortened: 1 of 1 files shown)


proposals/NNNN-collection-conditionals.md
```diff
@@ -0,0 +1,61 @@
...
+
+As conditionals themselves and inactive elements are not included in the parser AST representation, a new data structure, the "Conditionals Map" is maintained on the `CollectionExpr` which is used to support features such as the AST dump and stripping conditionals from module interfaces. The syntax model of libSyntax also required minor modification.
+
+## Source compatibility
+
+N/A. This is an purely additive proposal for syntax that is not currently valid in Swift.
+
+## Effect on ABI stability
+
+N/A. This is a compile time alteration of a collections's elements. The resulting collection is a conventional container as it would have been without the conditional though exactly which elements are included can affect the collection's type.
```
getaaron: In an example like this:

```
let params = [
#if DEBUG
  42,
#else
  "release",
#endif
]
```

The compiler would infer a type of `[Any]`, and an error would state `params: [Any]` needs to be explicitly declared, correct?

However in this example:

```
let params = [
#if DEBUG
  "localEnvironment",
#else
  "releaseEnvironments": ["integration.api.example.com", "uat.api.example.com", "api.example.com"],
#endif
]
```

The compiler can't reliably determine if this is even an array or a dictionary, so perhaps this should be an error. Currently the error for

```
let params = [
"localEnvironment",
"releaseEnvironments": ["integration.api.example.com", "uat.api.example.com", "api.example.com"],
]
```

is 

> Expected ',' separator
> Expected expression in container literal

So maybe a more helpful error message would be useful as part of this proposal, perhaps something like:

> Collection literal cannot include both `ArrayLiteralElement` and `KeyValuePairs`

PR #924, opened 2018-10-05
Swift 5.0 -> Swift 5?

linqingmo:
Swift 5


benrimmington:
Thanks, now it will be filtered by: [Implemented (Swift 5)](https://apple.github.io/swift-evolution/#?version=5)

PR #929, opened 2018-10-11
[Proposal] Deprecate Tuple Shuffles

CodaFi:
[Implementation is here](https://github.com/apple/swift/pull/19826)

### DIFF
--- a/proposals/NNNN-deprecate-tuple-shuffles.md
+++ b/proposals/NNNN-deprecate-tuple-shuffles.md
@@ -1,0 +1,141 @@
+Deprecate Tuple Shuffles
+========================
+
+-   Proposal: [SE-NNNN](NNNN-filename.md)
+-   Authors: [Robert Widmann](https://github.com/codafi), [David Hart](https://github.com/hartbit)
+-   Review Manager: TBD
+-   Status: **Awaiting review**
+
+Introduction
+------------
+
+This proposal seeks to deprecate certain forms of a little-known feature of Swift called a "Tuple Shuffle"†. A Tuple Shuffle is an undocumented feature of Swift in which one can reference tuple labels out of order in certain expressions.
+
+### Tuple-Shuffle Examples
+
+Tuple Shuffles are best seen in action:
+
+* Assigning to a tuple with the same labels in a different order:
+
+```swift
+let a = (x: 1, y: 2)
+let b: (y: Int, x: Int)
+b = a
+print(b) // prints (y: 2, x: 1)
+```
+
+* Destructuring a tuple by referencing labels out of order:
+
+```swift
+let a = (x: 1, y: 2)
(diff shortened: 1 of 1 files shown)


proposals/NNNN-deprecate-tuple-shuffles.md
```diff
@@ -0,0 +1,140 @@
...
+Proposed solution
+-----------------
+
+Construction of Tuple Shuffles will become a warning in Swift 4 compatibility mode and will be a hard error in Swift 5.
+
+Detailed design
+---------------
+
+†Throughout this proposal, we have referred to "Tuple Shuffles" as a monolithic feature.  However,
+the compiler currently models many features - mostly related to argument matching - with *Tuple Shuffle
```
slavapestov: Note that argument lists do not support re-ordering. It’s best to keep the two concepts separate here. In fact I’d like to split off an ArgumentShuffleExpr from TupleShuffleExpr one way, so that only the latter needs to support default and variadic arguments.

hartbit:
@tkremenek can this proposal be reviewed in time before Swift 5 branches off?

tkremenek:
@hartbit Sorry for the delay in responding.  I'm discussing this with other members of the Core Team on whether this is too late for Swift 5.  As long as the change is minimally impacting on users — and the compiler — it may still be eligible for Swift 5.

CodaFi:
@tkremenek Any movement here?

tkremenek:
Sorry for the silence.  I'm soul-searching after a conversation with the Core Team.

The reality is we can't rip this out of the compiler any time soon.  Swift 4 mode will be around for a long time, and possibly indefinitely.  There is very little carrot here, and mostly stick, for our users.  There's also a very high bar at this point for taking source-breaking changes into the language.

The primary motivation for this proposal is reducing complexity in the compiler's implementation.   But it is not clear to me if that goal will be ever achieved, per my comment above about the longevity of Swift 4 mode.  There are some user benefits to removing this feature, as it is confusing for many users, but it also isn't causing a lot of active harm either.

We are at the point where we cannot continue to just make piecemeal refinements to the language like this that have a source compatibility impact.  If this can tie in with a broader set of refinements that lead to a clear gain to the user experience — perhaps a win with compile time — then holistically those changes can be justified.  Right now for a user this feels like all stick and no carrot.

That said, I'm still mulling this over.  But I think we need to be clearer about the motivations here, and whether or not they meet the criteria.  Note that my conclusions here would have applied months ago; it's not just because we are past the `swift-5.0-branch` date.

lattner:
Regardless of whether this reduces complexity in the compiler (it surely would if the compat mode could ever be dropped...), I'm still +1 on this proposal.  It reduces the complexity of the *language*, and should improve diagnostics in dialects that don't have this enabled.

rjmccall:

PR #933, opened 2018-10-19
Reform Set Algebra

airspeedswift:
ABI/move-only type-motivated proposal to update/split up `SetAlgebra`.

### DIFF
--- a/proposals/NNNN-reform-set.md
+++ b/proposals/NNNN-reform-set.md
@@ -1,0 +1,222 @@
+# Reform SetAlgebra
+
+* Proposal: [SE-NNNN](NNNN-reform-set-algebra.md)
+* Authors: [Ben Cohen](https://github.com/airspeedswift)
+* Review Manager: TBD
+* Status: **Awaiting implementation**
+* Implementation: [apple/swift#NNNNN](https://github.com/apple/swift/pull/NNNNN)
+
+## Introduction
+
+This proposal makes two changes to the `SetAlgebra` protocol, as well as corresponding changes to conforming types in the Standard Library (and Foundation):
+
+- amend `insert` to return the index to the current element in the set, not the element itself; and
+- split `SetAlgebra` into two protocols, for intensional and extensional sets.
+
+These are **source breaking** changes. The primary motivation for breaking source now is ensuring that the ABI and API allow for future types and language features – specifically move-only types.
+
+Swift-evolution thread: [Discussion thread topic for that proposal](https://forums.swift.org/)
+
+## Motivation
+
+### Preparation for move-only types
+
+As described in [SE-0232](https://github.com/apple/swift-evolution/blob/master/proposals/0232-remove-customization-points.md), requirements on a protocol where an element is returned, but not removed, from a collection will be a problem for collections that want to be able to hold move-only types (because returning the element means it would have to be moved out of the collection, so it can't remain in the collection).
+ There is one such method on the `SetAlgebra` protocol, `insert`:
+
+```swift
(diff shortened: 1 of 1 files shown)


proposals/NNNN-reform-set.md
```diff
@@ -0,0 +1,222 @@
...
+///   other means.
+@discardableResult
+mutating func insert(_ newMember: Element) -> (inserted: Bool, memberAfterInsert: Element)
+```
+
+The operation returns the new value if an equal value wasn't already present; and _discards_ the new value and instead returns the existing old value if one was.
+
+This will be a problem for a set of move-only types. If the inserted type is move-only, it could not possibly be both inserted into the set _and_ returned.
+
+Contrast this with `Dictionary`'s `updateValue(_:forKey:)`, which replaces the existing value if there is one and returns that – something that would work fine for move-only types:
```
lorentey: `SetAlgebra` has an `update(with:)` requirement with similar behavior; it would probably make for a closer example.

proposals/NNNN-reform-set.md
```diff
@@ -0,0 +1,216 @@
...
+  // In-place operations can just be implemented
+  // with self-assignment
+}
+```
+
+But some of them are impossible to implement.
+
+Some requirements effectively require the ability to enumerate the elements. `isEmpty` requires you to know `contains` will always return false. This is impossible for an opaque closure. `Equatable` conformance has the same problem.
+
+`init<S : Sequence>(_ sequence: S) where S.Element == Element` requires the ability to initialize from a sequence. At a minimum, this would require that `Element` be `Equatable` (and even then, testing would take `O(n)` — `Comparable` or `Hashable` is probably the minimum). In most cases, this is taken care of by the confoming type (e.g. `Set` requires `Element: Hashable`). But a predicate set _might not_ need this requirement – the predicate the user supplied could use some other means. Ideally, `SetAlgebra` wouldn't impose any requirement on `Element`, but also wouldn't require an unconstrained `init` from a sequence. `ExpressibleByArrayLiteral` has the same problem.
```
benrimmington: Typo: `confoming` => `conforming`

proposals/NNNN-reform-set.md
```diff
@@ -0,0 +1,216 @@
...
+
+This is a source-breaking change.
+
+The main driver is to ensure that `SetAlgeba` is ready for move-only types, which requires fixing `insert`. This will unavoidably break conformances. Once that possibility is admitted, it is not significantly worse to introduce a second break for implementors.
+
+Some problems can be mitigated by keeping `SetAlgebra` as a typealias for `ExtensionalSet`. This will cover any extensions on `SetAlgebra`.
+
+Callers of `insert` that make use of the `Element` return type will have to update their code to add the `subscript` call.
+
+There are no instances of either of these in the compatability suite currently.
```
benrimmington: Typo: `compatability` => `compatibility` (1 of 2)

proposals/NNNN-reform-set.md
```diff
@@ -0,0 +1,216 @@
...
+
+This is an ABI-breaking change. The motivation is to get the ABI right for sets of move-only types, which cannot implement `insert` currently.
+
+## Effect on API resilience
+
+None.
+
+## Alternatives considered
+
+It is reasonable to argue that the generic `SetAlgebra` is not pulling its weight – that it isn't common to want to write generic algorithms using `SetAlgebra`. The lack of use in the compatability suite supports this.
```
benrimmington: Typo: `compatability` => `compatibility` (2 of 2)

PR #936, opened 2018-10-21
Synthesize default values for the memberwise initializer

Azoy:


tkremenek:
The Core Team is planning to run this review imminently.

@Azoy: you OK if we run this review very soon, such as this coming week?

Azoy:
@tkremenek Yes, that's completely okay with me!

tkremenek:
Launching the review now.

PR #937, opened 2018-10-22
[Proposal]: Add Result to the Standard Library

jshier:
This PR tracks the proposal for adding Result to the standard library.


proposals/NNNN-add-result.md
```diff
@@ -0,0 +1,242 @@
...
+    guard let data = data, let response = response else { return // Impossible? }
+    
+    handleResponse(response, data: data)
+}
+```
+
+While this code is only a few lines long, it exposes Swift's complete lack of automatic error handling for asynchronous APIs. Not only was the `error` forcibly unwrapped (or perhaps handled using a slightly less elegant `if` statement), but a possibly impossible scenario was created. What happens if `response` or `data` are `nil`? Is it even possible? It shouldn't be, but Swift currently lacks the ability to express this impossibility. Using `Result<Value, Error>` for the same scenario allows for much more elegant code:
+
+```swift
+URLSession.shared.dataTask(with: url) { (result: Result<(URLResponse?, Data?), Error>) in // Type added for illustration purposes.
```
Dante-Broggi: See [this article](https://oleb.net/blog/2018/03/making-illegal-states-unrepresentable/) by Ole Begemann for why the correct signature for this closure is actually `Result<(Data, URLResponse), (Error, URLResponse?)>`, which also is an example of a `Result` with a non `Swift.Error` error case.
```suggestion
URLSession.shared.dataTask(with: url) { (result: Result<(Data, URLResponse), (Error, URLResponse?)>) in // Type added for illustration purposes.
```

proposals/NNNN-add-result.md
```diff
@@ -0,0 +1,242 @@
...
+    
+    ... 
+}
+
+```
+
+This can be made somewhat cleaner by using a `(string: String?, error: Error?)` to store the result, but it would still have the same usage issues as in the asynchronous case. Using a `Result` is the appropriate answer here, especially with the convenience API for creating a result from a `throw`ing function.
+
+```swift
+let configuration = Result { try String(contentsOfFile: configuration) }
```
Dante-Broggi: I propose we add special syntax to do this, enabling us to avoid the `try` and extraneous closure.
```suggestion
let configuration = catch String(contentsOfFile: configuration)
```

jshier: I've specifically kept any special syntax out of this proposal, both to keep the proposal and  implementation simpler and to keep the language and usage impacts more predictable. I think some special syntax might be called for, but I'd like to leave it as a separate proposal. But if it's important to get this proposal into review, I can certainly add it. I would require help to implement it however.

proposals/NNNN-add-result.md
jshier: I disagree, but I do think the proposed type illustrates one of the advantages of an unconstrained `Result`. Thanks!

proposals/NNNN-add-result.md
```diff
@@ -0,0 +1,242 @@
...
+    guard let data = data, let response = response else { return // Impossible? }
+    
+    handleResponse(response, data: data)
+}
+```
+
+While this code is only a few lines long, it exposes Swift's complete lack of automatic error handling for asynchronous APIs. Not only was the `error` forcibly unwrapped (or perhaps handled using a slightly less elegant `if` statement), but a possibly impossible scenario was created. What happens if `response` or `data` are `nil`? Is it even possible? It shouldn't be, but Swift currently lacks the ability to express this impossibility. Using `Result<Value, Error>` for the same scenario allows for much more elegant code:
+
+```swift
+URLSession.shared.dataTask(with: url) { (result: Result<(Data, URLResponse), (Error, URLResponse?)>) in // Type added for illustration purposes.
```
nh7a: Unfortunately, this contradicts the implementation of `init(_ throwing: () throws -> Value)` below because you cannot throw `(Error, URLResponse?)`.

jshier: I've updated the implementation in the proposal to match the current state of the PR, which places the throwing initializer in a constrained extension.

proposals/NNNN-add-result.md
```diff
@@ -0,0 +1,243 @@
...
+
+func doSomethingWithConfiguration() {
+    ...
+}
+
+```
+
+#### Separating Errors
+
+It's occasionally useful to be able to run `throw`able functions in such way as to allow the developer to disambiguate between the sources of the errors, especially if the errors don't contain the information necessary to do, or the developer doesn't want to implement such a check. For instance, if we needed to disambiguate between the errors possible when reading files:
```
spadafiva:
```suggestion
It's occasionally useful to be able to run `throw`able functions in such way as to allow the developer to disambiguate between the sources of the errors, especially if the errors don't contain the information necessary to do so, or the developer doesn't want to implement such a check. For instance, if we needed to disambiguate between the errors possible when reading files:
```

Is this a typo?

jshier: Yes, thanks!

proposals/NNNN-add-result.md
```diff
@@ -0,0 +1,365 @@
...
+
+## Proposed solution
+
+```swift
+public enum Result<Value, Error> {
+    case success(Value), failure(Error)
+}
+```
+
+`Result<Value, Error>` is a pragmatic compromise between competing error handling concerns both present and future. It is unconstrained for the `Error` type captured in the `.failure` case to both allow for a possible type-`throws` future and allow failures to return values that don't conform to `Swift.Error`.
```
natecook1000:
```suggestion
`Result<Value, Error>` is a pragmatic compromise between competing error handling concerns both present and future. It is unconstrained for the `Error` type captured in the `.failure` case both to allow for a possible typed-`throws` future and to allow failures to return values that don't conform to `Swift.Error`.
```

proposals/NNNN-add-result.md
```diff
@@ -0,0 +1,293 @@
...
+  /// Evaluates the given transform closure when this `Result` instance is
+  /// `.success`, passing the value as a parameter and flattening the result.
+  ///
+  /// - Parameter transform: A closure that takes the successful value of the
+  ///   instance.
+  /// - Returns: A new `Result` instance, either from the transform or from
+  ///   the previous error value.
+  public func flatMap<NewValue>(
+    _ transform: (Value) throws -> NewValue
+  ) -> Result<NewValue, Swift.Error>
```
DevAndArtist: The error type does not have to be `Swift.Error` because of the constrain from the extension.
```suggestion
  ) -> Result<NewValue, Error>
```

proposals/NNNN-add-result.md
```diff
@@ -0,0 +1,293 @@
...
+
+* Proposal: [SE-NNNN](NNNN-add-result.md)
+* Authors: [Jon Shier](https://github.com/jshier)
+* Review Manager: TBD
+* Status: **Awaiting implementation**
+* Implementation: [apple/swift#19982](https://github.com/apple/swift/pull/19982)
+
+## Introduction
+
+Swift's current error-handling, using `throws`, `try`, and `catch`, offers automatic and synchronous handling of errors through explicit syntax and runtime behavior. However, it lacks the flexibility needed to cover all error propagation and handling in the language. `Result` is a type commonly used for manual propagation and handling of errors in other languages and within the Swift community. Therefore this proposal seeks to add such a type to the Swift standard library.
```
fabb: `async throw` functions would also relieve the mentioned pain points, maybe reference the other proposal draft that that seeks to introduce Coroutines? https://gist.github.com/lattner/429b9070918248274f25b714dcfc7619

proposals/NNNN-add-result.md
jshier: Fixed.

tkremenek:
@lattner will run a review for this proposal shortly.

lattner:
The core team discussed this today and would like to run this proposal ASAP.  Is that ok with you @jshier ?

jshier:
@lattner Sure! Just let me know if there's anything else you need from me.

proposals/NNNN-add-result.md
jshier: While something like `async throw` may address the automatic handling of errors for async work, it wouldn't address anything for the other use cases, and would likely still need a manually propagating version (like a `Future` returning a `Result`).

PR #941, opened 2018-10-29
Improved zip API

dennisvennink:
This PR tracks the *Improved `zip` API* proposal.

### DIFF
--- a/proposals/nnnn-improved-zip-api.md
+++ b/proposals/nnnn-improved-zip-api.md
@@ -1,0 +1,976 @@
+# Improved `zip` API
+
+- Proposal: [SE-NNNN](nnnn-improved-zip-api.md)
+- Author: [Dennis Vennink](https://github.com/dennisvennink)
+- Review Manager: To Be Determined
+- Status: **Awaiting Review**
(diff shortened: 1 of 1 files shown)


airspeedswift:
Feedback from the core team: this is a definitely an appropriate proposal, but it depends on the ability to mark conformances to protocols with availability. So it is blocked on that feature becoming available before it can be run. I've added a label to mark it as such.

rjmccall:
This proposal was recently unblocked by the addition of conformance availability.  The Language Workgroup is trying to be better about not allowing proposals to sit for long periods without making progress, but I will leave this proposal open for a grace period of a few weeks before closing it as stale, in case someone is still interested in bringing it through evolution.

rjmccall:
The Language Workgroup discussed this today, and we think this idea is potentially appropriate for the standard library.  However, we don't think we should take it there until it can be rebuilt on top of variadic generics.  In the meantime, it can be considered for the Algorithms package, which is not controlled by the evolution process.  Therefore, we're closing this PR for now, and it can be reconsidered in the future when everything is in plan.

PR #942, opened 2018-10-31
[Proposal] Make Numeric Refine a new AdditiveArithmetic Protocol

rxwei:
Pitch thread: [Make Numeric Refine a new AdditiveArithmetic Protocol](https://forums.swift.org/t/make-numeric-refine-a-new-additivearithmetic-protocol/17269)


proposals/0215-additive-arithmetic-protocol.md
```diff
@@ -0,0 +1,209 @@
...
+```swift
+static func * (lhs: Vector, rhs: Vector) -> Vector { ... }
+static func * (lhs: Vector, rhs: Scalar) -> Vector { ... }
+```
+
+This compiles, but does not work in practice. The following trivial use case would fail to compile, because literal  `1`  can be implicitly converted to both a  `Scalar`  and a  `Vector` , and  `*`  is overloaded for both  `Vector`  and  `Scalar` .
+
+```swift
+let x = Vector<Int>(...)
+x * 1 // Ambiguous! Can be both `x * Vector(integerLiteral: 1)` and `x * (1 as Int)`.
```
natecook1000: I meant to note that this isn't ambiguous, since the type checker will always first try to use an literal's default type (`Int` for integer literals). If there's a use that supports that, the default wins. When there's nothing available using the default, then there's an ambiguity, so making this example use a `Vector<Double>` would be a better illustration. There's no precedence in that case between `Vector<Double>` and `Vector<Double>.Scalar` (i.e. `Double`), so there's an ambiguity.

PR #949, opened 2018-11-03
[SE-0233] Revise based on review comments

rxwei:
Addressing comment https://forums.swift.org/t/se-0233-make-numeric-refine-a-new-additivearithmetic-protocol/17583/11.


rxwei:
Ok, better :)

PR #966, opened 2018-12-10
Extending operator declarations with designated types

rudkx:


proposals/0239-operator-designated-types.md
```diff
@@ -0,0 +1,137 @@
...
+* Proposal: [SE-0239](0239-operator-designated-types.md)
+* Authors: [Mark Lacey](https://github.com/rudkx)
+* Review Manager: TBD
+* Status: **Implemented under staging options**
+
+*During the review process, add the following fields as needed:*
+
+* Implementation: This is currently in-tree, disabled by default, and enabled by staging options (`-enable-operator-designated-types`, `-solver-enable-operator-designated-types`).
+* Decision Notes: [Rationale](https://forums.swift.org/), [Additional Commentary](https://forums.swift.org/)
+* Bugs: [SR-NNNN](https://bugs.swift.org/browse/SR-NNNN), [SR-MMMM](https://bugs.swift.org/browse/SR-MMMM)
```
benrimmington:
```suggestion
* Bug: [SR-7847](https://bugs.swift.org/browse/SR-7847)
```

proposals/0239-operator-designated-types.md
```diff
@@ -0,0 +1,137 @@
...
+- Expressions involving collection literals sometimes fail even if they have only a handful of elements (e.g. < 20).
+- Chains of single-expression closures, which in addition to exacerbating the issues with operators and literals, also create new problems due to a lack of explicit types for their parameters.
+
+This proposal aims to address only the first of these issues.
+
+## Proposed solution
+
+We introduce syntax to define *designated types* in operator declarations. These types will be used as the first place to look up operator overloads for a given operator. If the expression successfully typechecks with an overload found via those types, we stop looking at the remaining overloads for the operator.
+
+The intent is to use this within `Policy.swift` in the standard library in order to provide a hint to the type checker as to which overload should be preferred for typechecking an operator. You can see the current usage of this syntax [here](https://github.com/apple/swift/blob/master/stdlib/public/core/Policy.swift#L398).
```
benrimmington: Use permalink? e.g. [here](https://github.com/apple/swift/blob/191a71e10dab37aa12b53f04b9496642dd3604b1/stdlib/public/core/Policy.swift#L382-L485).

proposals/0239-operator-designated-types.md
```diff
@@ -0,0 +1,137 @@
...
+
+Users can also add their own designated types to operators that they declare, but they will not have a way to add their own types as designated types to operators already defined in `Policy.swift`.
+
+In effect this ties a particular operator to the semantics of one (or occassionally multiple) particular named entity(ies), usually a protocol, but sometimes a concrete type. For example, if you're implementing `<<`, making `BinaryInteger` the designated type for that operator says that the expectation is that you will be implementing this for a type conforming to `BinaryInteger` and behaving in the way `BinaryInteger` would expect it to.
```
benrimmington: occassionally => occasionally

```suggestion
In effect this ties a particular operator to the semantics of one (or occasionally multiple) particular named entity(ies), usually a protocol, but sometimes a concrete type. For example, if you're implementing `<<`, making `BinaryInteger` the designated type for that operator says that the expectation is that you will be implementing this for a type conforming to `BinaryInteger` and behaving in the way `BinaryInteger` would expect it to.
```

proposals/0239-operator-designated-types.md
```diff
@@ -0,0 +1,137 @@
...
+## Experimental results
+
+The implementation for this is already in-tree, and is enabled as such:
+
+- `-Xfrontend -enable-operator-designated-types`: enables parsing the extended operator declarations
+- `-Xfrontend -solver-enable-operator-designated-types`: enables having the constraint solver make use of the declarations
+
+The current implementation has been tested across 25 test cases that were known to be slow and/or result in the type checker giving up after expending substantial effort attempting to type check the expression.
+
+In addition to enabling this new *designated type* functionality, this testing disables the `shrink` portion of the expression type checker (via `-Xfrontend -solver-disable-shrink`).  This portion of the expression type checker is known to help occassionally but often results in substantially increasing the expression type checking time since it frequently doesn't provide any benefit but does a lot of work. We also disable a number of other problematic hacks (via `-Xfrontend -disable-constraint-solver-performance-hacks`).
```
benrimmington: occassionally => occasionally

```suggestion
In addition to enabling this new *designated type* functionality, this testing disables the `shrink` portion of the expression type checker (via `-Xfrontend -solver-disable-shrink`).  This portion of the expression type checker is known to help occasionally but often results in substantially increasing the expression type checking time since it frequently doesn't provide any benefit but does a lot of work. We also disable a number of other problematic hacks (via `-Xfrontend -disable-constraint-solver-performance-hacks`).
```

proposals/0239-operator-designated-types.md
```diff
@@ -0,0 +1,137 @@
...
+
+```
+
+A number (25+) of other test cases were previously compiling quickly were also tested using this functionality and with the existing hacks disabled, and those cases are as fast or faster than previously.
+
+## Source compatibility
+
+Because of the fact that we fall back to behavior that is very similar to the current behavior of the type checker (exhaustively testing combinations of operators in the cases), this change is very compatible with the current behavior. 
+
+One failure was seen in the source compatibility suite when running with this new behavior enabled by default in all language modes. The failure was in `BlueSocket`, and that same failure also manifests if you build the affected code with `-swift-version 4`. Specificially, for `let mask = 1 << (2 as Int32)`, we now infer `mask` as `Int` rather than `Int32`.
```
benrimmington: Specificially => Specifically

```suggestion
One failure was seen in the source compatibility suite when running with this new behavior enabled by default in all language modes. The failure was in `BlueSocket`, and that same failure also manifests if you build the affected code with `-swift-version 4`. Specifically, for `let mask = 1 << (2 as Int32)`, we now infer `mask` as `Int` rather than `Int32`.
```

proposals/0239-operator-designated-types.md
rudkx: @benrimmington Did you confirm that this is fixed with these changes enabled?

proposals/0239-operator-designated-types.md
rudkx: @benrimmington Fixed in https://github.com/apple/swift-evolution/pull/966/commits/c051e12766d568dab70002374fdfccce0d8d0467, thanks!

proposals/0239-operator-designated-types.md
benrimmington: Sorry, I didn't check. Maybe @dan-zheng can help.

rudkx: Okay, there are a bunch of cases where this will help and I'd rather not list ones that we don't know are helped for certain. I might make a quick pass at listing some that I can validate still repro without this change and are fixed by it, if I can make the time.

proposals/0239-operator-designated-types.md
```diff
@@ -0,0 +1,159 @@
...
+
+We can see this in the [`Anchorage`](https://github.com/RaizLabs/Anchorage) project on GitHub.
+
+```
+public func == <T: BinaryFloatingPoint>(lhs: NSLayoutDimension, rhs: T) -> NSLayoutConstraint {
+    return finalize(constraint: lhs.constraint(equalToConstant: CGFloat(rhs)))
+}
+```
+
+Here, `NSLayoutDimension` and `T` are both `Equatable`, and as a result code like:
```
benrimmington: Are you sure this is correct?

The homogeneous `==` (two `Self` parameters) doesn't seem to be interchangeable with the generic heterogeneous `==` (`NSLayoutDimension` and `T: BinaryFloatingPoint` parameters).

If the operands are two `.widthAnchor: NSLayoutDimension & Equatable` values, I'd have thought the alternatives are:

```swift
public protocol Equatable {
  static func == (lhs: Self, rhs: Self) -> Bool
}
```

and [this one](https://github.com/Raizlabs/Anchorage/blob/89c3b605638a2883e954cc0b7bea2c2f19daf7fd/Source/Anchorage.swift#L57-L59):

```swift
public func == (lhs: NSLayoutDimension, rhs: NSLayoutDimension) -> NSLayoutConstraint
```

proposals/0239-operator-designated-types.md
```diff
@@ -0,0 +1,159 @@
...
+The current implementation has been tested across 25 test cases that were known to be slow and/or result in the type checker giving up after expending substantial effort attempting to type check the expression.
+
+In addition to enabling this new *designated type* functionality, this testing disables the `shrink` portion of the expression type checker (via `-Xfrontend -solver-disable-shrink`).  This portion of the expression type checker is known to help occassionally but often results in substantially increasing the expression type checking time since it frequently doesn't provide any benefit but does a lot of work. We also disable a number of other problematic hacks (via `-Xfrontend -disable-constraint-solver-performance-hacks`).
+
+In 14 cases, the expressions in question now typecheck in less than 30ms (often as little as 5ms). In the 11 remaining cases, the expression is either one that would fail to typecheck if we were able to typecheck it quickly, or involves other things that are known to be problematic, like closures. 
+
+
+Some illustrative examples:
+
+```
```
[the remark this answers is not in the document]
rudkx: Fixed all three in e054585.

proposals/0239-operator-designated-types.md
rudkx: Fixed in e054585.

proposals/0239-operator-designated-types.md
rudkx: Fixed in 0c9072860781b5e28bffdbe771d3741c97df16c6.

proposals/0239-operator-designated-types.md
```diff
@@ -0,0 +1,164 @@
+# Extending operator declarations with designated types
+
+* Proposal: [SE-0239](0239-operator-designated-types.md)
```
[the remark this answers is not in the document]
airspeedswift: It should be NNNN until it's merged. I updated the PR.

rudkx:
@airspeedswift Thanks.

rudkx:
Closing. It can be reopened if there is interest in trying to push it through review in lieu of some other creative new ideas.

PR #975, opened 2019-01-13
added loading state in first page load

uxitten:
sometime server is delay to response or internet connection is low

### DIFF
--- a/index.html
+++ b/index.html
@@ -77,8 +77,7 @@ See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
                 <article class="proposals">
                     <div id="proposals-count"><span id="proposals-count-number"></span></div>
                     <ul class="proposals-list">
-                        <!-- {proposals} -->
-                        <li>Loading ...</li>
+                        <!-- {proposals} -->                        
                     </ul>
                 </article>
             </main>
--- a/index.html
+++ b/index.html
@@ -78,6 +78,7 @@ See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
                     <div id="proposals-count"><span id="proposals-count-number"></span></div>
                     <ul class="proposals-list">
                         <!-- {proposals} -->
+                        <li>Loading ...</li>
                     </ul>
                 </article>
             </main>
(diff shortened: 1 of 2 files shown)


krilnon:
Why add `debounce`? I don't think that rendering while typing in the filter box poses a performance issue.

uxitten:
@krilnon yes i agree with you, but if there's a big record, it's a problem

tkremenek:
@krilnon you OK with merging this?

index.html
```diff
@@ -77,7 +77,7 @@ <h1 id="title"><a href="#">Swift Evolution</a></h1>
                 <article class="proposals">
                     <div id="proposals-count"><span id="proposals-count-number"></span></div>
                     <ul class="proposals-list">
-                        <!-- {proposals} -->
+                        <!-- {proposals} -->
```
krilnon: Why add whitespace here?

index.js
```diff
@@ -128,9 +128,10 @@ function init () {
   })
 
   req.addEventListener('error', function (e) {
-    document.querySelector('#proposals-count').innerText = 'Proposal data failed to load.'
+    document.querySelector('#proposals-count-number').innerText = 'Proposal data failed to load.'
   })
 
+  document.querySelector('#proposals-count-number').innerHTML = 'Loading ...';
```
krilnon: Remove the semicolon here and other places where they appear to follow the [style guide](https://standardjs.com).

index.html
uxitten: thanks, fixed

index.js
uxitten: fixed

uxitten:
i think we need `debounce` because the volume of data will increase in the future

uxitten:
is not this resolved?

tkremenek:
@krilnon?

krilnon:
There are still some style guide rules being broken here. You fixed the two places with semicolons that I pointed out explicitly, but you should just remove all the ones you added. There's [a tool to check these rules automatically](https://standardjs.com/#usage) that you might find useful.

> i think we need `debounce` because the volume of data will increase in the future

Is there a place where you've seen a performance issue here, or, if you haven't seen one yet, have you done any measurements that indicate *the conditions in which* we'd need debounce? I'm guessing it'll be a couple more years until we even approach 1000 proposals; and I don't think there'd be performance problems even then.

uxitten:
@krilnon all right, i am removing debounce

uxitten:
@krilnon removed now

krilnon:
Alright, looks good now. Thanks for the improvement and for working through this with me!

uxitten:
@krilnon you're welcome

PR #977, opened 2019-01-17
Key Path Expressions as Functions

stephencelis:
Proposal to allow `\Root.value` key path expressions where functions `(Root) -> Value` are allowed.


Pearapps:
yes

jVirus:
I find this proposal quite handy from the practical perspective

tkremenek:
The Core Team will kick off a review of this on March 18.  @airspeedswift will manage.

PR #980, opened 2019-01-25
Fix markdown block not being closed

adibfara:
Close the markdown for swift (typo).


krilnon:
Thanks for spotting and fixing this!

adibfara:
> 
> 
> Thanks for spotting and fixing this!

No problem :)

PR #983, opened 2019-01-29
Add `0242-presencable-protocol` proposal

dimpiax:
Hello, please review the proposal.
@krilnon Link to the working copy: https://github.com/dimpiax/0242-presencable-protocol/blob/master/0242-presencable-protocol.playground/Contents.swift


DougGregor:
Re-reading the pitch thread, there was not sufficient support for bringing this up for a review. If you'd like this to be reconsidered, please re-pitch.

PR #986, opened 2019-02-19
Update my name in proposals

lilyball:


erica:
I have no objections to Lily's proposed changes that affect documents I've worked on.

millenomi:
Thank you for fixing dangling pointers.

PR #997, opened 2019-03-05
Add second proposal for collection-rotation methods

CTMacUser:
I modernized SE-0078 for Swift 4 and 5.  It isn't a straight adaptation; I added my own spin to the idea.


rjmccall:

This proposal may also now be redundant with the Algorithms package.

PR #1005, opened 2019-03-22
Updates for math functions proposal.

stephentyrone:
Moved static implementations out of the "Math" pseudo-namespace and back onto the types themselves. This lets us drop a layer of protocol indirection that would have been fixed in the ABI "forever".


stephentyrone:
@rjmccall this is all set, but I'll let you merge.

PR #1007, opened 2019-03-25
Proposal: Static and class subscripts

brentdax:
The implementation is not quite completely tested yet, but it's far enough along that I think we can prepare the proposal for review.

### DIFF
--- a/proposals/0253-static-subscripts.md
+++ b/proposals/0253-static-subscripts.md
@@ -1,6 +1,6 @@
 # Static and class subscripts
 
-* Proposal: [SE-0252](0252-static-subscripts.md)
+* Proposal: [SE-0253](0253-static-subscripts.md)
 * Authors: [Brent Royal-Gordon](https://github.com/brentdax)
 * Review Manager: TBD
 * Status: **Awaiting review**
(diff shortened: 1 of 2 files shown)


tkremenek:
@brentdax what are the current limitations of the implementation?

brentdax:
@tkremenek The last open question that would affect the proposal's drafting is whether to keep metatype key paths in its scope. I'm currently working on emitting the metadata needed to support them in public resilient modules; I think I'll know by end of day tomorrow whether I want to include or defer them.

Besides that:

* I haven't tested existential metatypes very much yet, and I wouldn't be surprised if some of the current implementation only works for concrete metatypes.
* I haven't started seriously working on code completion or other editor features.

brentdax:
I think there's enough doubt about resilient cross-module key paths that they should be removed from the proposal. I'll probably write the proposal revisions tonight, but certainly by lunch tomorrow.

proposals/0253-static-subscripts.md
```diff
@@ -0,0 +1,132 @@
+# Static and class subscripts
+
+* Proposal: [SE-0253](0253-static-subscripts.md)
```
krilnon: Usually it makes sense to delay numbering a proposal until it's merged. E.g. there's already an [SE-0253](https://github.com/apple/swift-evolution/blob/master/proposals/0253-callable.md).

brentdax: Yes, I'm quickly learning that lesson. 😅

DougGregor: No worries, I'll fix it

PR #1009, opened 2019-03-26
Add the callable proposal.

rxwei:
This PR adds the proposal for (static) callables (see [pitch thread](https://forums.swift.org/t/pitch-introduce-static-callables/21732)).

### DIFF
--- a/proposals/0252-callable.md
+++ b/proposals/0252-callable.md
@@ -596,15 +596,3 @@ np.random.randint(-10, 10, dtype: np.float)
 // `PythonObject` with `call` members. The empty strings are killer.
 np.random.randint(["": -10, "": 10, "dtype": np.float])
 ```
- -10, "": 10, "dtype": np.float])
-```
-at is crucial for usability.
-
-```swift
-// Let `PythonObject` be a `@dynamicMemberLookup` type with callable functionality.
-let np: PythonObject = ...
-// `PythonObject` with `@dynamicCallable.
-np.random.randint(-10, 10, dtype: np.float)
-// `PythonObject` with `call` members. The empty strings are killer.
-np.random.randint(["": -10, "": 10, "dtype": np.float])
-```
--- a/proposals/0252-callable.md
+++ b/proposals/0252-callable.md
@@ -45,7 +45,7 @@ Here are some concrete sources of motivation.
 
 ### Values representing functions
 
-Values of some types exactly represent functions: in the mathematical sense (a mapping from inputs to outputs), or in the context of programming languages.
+Values of some nominal types exactly represent functions: in the mathematical sense (a mapping from inputs to outputs), or in the context of programming languages.
 
 Here are some examples:
 
--- a/proposals/0252-callable.md
+++ b/proposals/0252-callable.md
@@ -37,9 +37,9 @@ Currently, in Swift, only a few kinds of values are syntactically callable:
 - Type names (e.g. `T` can be called like `T(...)`, which is desugared to `T.init(...)`).
 - Values with a `@dynamicCallable` type.
 
-However, callable syntax can also be useful for other values, primarily those that behave like functions. This includes:
+However, call-syntax can also be useful for other values, primarily those that behave like functions. This includes:
 - Values that represent functions: mathematical functions, function expressions, etc.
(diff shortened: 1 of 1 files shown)


proposals/0252-callable.md
```diff
@@ -0,0 +1,611 @@
...
+```
+
+## Motivation
+
+Currently, in Swift, only a few kinds of values are syntactically callable:
+- Values with function types.
+- Type names (e.g. `T` can be called like `T(...)`, which is desugared to `T.init(...)`).
+- Values with a `@dynamicCallable` type.
+
+However, callable syntax can also be useful for other values, primarily those that behave like functions. This includes:
```
dan-zheng: Unify terminology: rename "callable syntax"and "call-syntax" to "call syntax"?
There are four mentions of "callable syntax".

proposals/0252-callable.md
```diff
@@ -0,0 +1,610 @@
...
+- Values that represent functions: mathematical functions, function expressions, etc.
+- Values that have one main use and want to provide a simple callable interface: neural network layers, parsers, efficient bound closures, etc.
+
+Here are some concrete sources of motivation.
+
+### Values representing functions
+
+Values of some types exactly represent functions: in the mathematical sense (a mapping from inputs to outputs), or in the context of programming languages.
```
dan-zheng: `Values of some types exactly represent functions` is a bit confusing, particularly `some types`.
```suggestion
Some types exactly represent functions: in the mathematical sense (a mapping from inputs to outputs), or in the context of programming languages.
```

proposals/0252-callable.md
```diff
@@ -0,0 +1,610 @@
...
+
+  call() { return function(value) }
+}
+
+let x = "Hello world!"
+let closure = BoundClosure(function: { print($0) }, value: x)
+closure() // prints "Hello world!"
+```
+
+A callable syntactic sugar would enable `BoundClosure` instances to be applied like normal functions.
```
dan-zheng:
```suggestion
A call syntax sugar would enable `BoundClosure` instances to be applied like normal functions.
```

proposals/0252-callable.md
```diff
@@ -0,0 +1,610 @@
...
+struct Adder {
+    var base: Int
+    @callDelegate
+    func addingWithBase(_ x: Int) -> Int {
+        return base + x
+    }
+}
+```
+
+This approach achieves a similar effect as `call` declarations, except that methods can have a custom name and be directly referenced by that name. This is useful for types that want to make use of the callable syntactic sugar, but for which the name "call" does not accurately describe the callable functionality.
```
dan-zheng:
```suggestion
This approach achieves a similar effect as `call` declarations, except that methods can have a custom name and be directly referenced by that name. This is useful for types that want to make use of the call syntax sugar, but for which the name "call" does not accurately describe the callable functionality.
```

proposals/0252-callable.md
dan-zheng: Maybe hyphenize as `call-syntax sugar`.

proposals/0252-callable.md
rxwei: "Types representing functions" is incorrect. I intentionally changed the original "some types ..." to "values of some types".

Functions are values, not types. Thus types cannot represent functions.

proposals/0252-callable.md
rxwei: Done.

proposals/0252-callable.md
dan-zheng: How about `Some values exactly represent functions?`

rxwei: That is technically correct, but it is correct more because "some values" can just _be_ functions, which feels like tautology and doesn't go into more details than the section title.
I changed it to "Values of some nominal types exactly represent functions".

proposals/0252-callable.md
```diff
@@ -0,0 +1,598 @@
...
+```
+
+#### Bound closures
+
+Variable-capturing closures can be modeled explicitly as structs that store the bound variables. This representation is more performant and avoids the type-erasure of closure contexts.
+
+```swift
+// Represents a nullary function capturing a value of type `T`.
+struct BoundClosure<T> {
+  var function: (T) -> Void
```
dan-zheng: 4 space indentation

PR #1015, opened 2019-04-02
[SE-NNNN] Implicit Returns from Single-Expression Functions

nate-chandler:


proposals/NNNN-omit-return.md
```diff
@@ -0,0 +1,361 @@
...
+let ages = persons.map { $0.name }
+```
+
+Their prior exposure to functions which implicitly return the single expression in their bodies will lead them to conclude that the closure being passed to map is returning the expression `$0.name` and that the return type of the closure is `String` (the type of `name`, here).
+
+## Detailed design
+
+Interpret the bodies of function-like entities which consist of a single expression as returning that expression--unless the entity's return type is `Void` or the single expression's type is uninhabited.
+
+#### Function-Like Entities
```
krilnon: You bounce back and forth between title case and sentence case for your headings. This heading and the proposal title are title case, the rest are sentence case.

nate-chandler: Nice catch.  As of https://github.com/apple/swift-evolution/pull/1015/commits/32f9fb3fdf0a8f0c3b5aea8ba2aaa54712a36421 they're all sentence case.

PR #1024, opened 2019-04-10
Add proposal for allowing implicit self in more circumstances

Jumhyn:


proposals/NNNN-implicit-self-explicit-capture.md
```diff
@@ -0,0 +1,189 @@
...
+* Authors: [Frederick Kellison-Linn](https://github.com/jumhyn)
+* Review Manager: TBD
+* Status: **Awaiting Review**
+* Implementation: [apple/swift#23934](https://github.com/apple/swift/pull/23934/)
+* Bug: [SR-10218](https://bugs.swift.org/browse/SR-10218)
+* Forum threads: [Discussion](https://forums.swift.org/t/review-capture-semantics-of-self/22017), [Pitch](https://forums.swift.org/t/allow-implicit-self-in-escaping-closures-when-self-is-explicitly-captured/22590)
+
+## Introduction
+
+Modify the rule that all uses of `self` in escaping closures must be explicit by allowing for implicit uses of `self` in situations where the user has already made their intent explicit, or where strong reference cycles are otherwise unlikely to occur. There are two situations covered by this proposal. The first is when the user has explicitly captured `self` in the closures capture list, so that the following would compile without error:
```
dabrahams:
```suggestion
Modify the rule that all uses of `self` in escaping closures must be explicit by allowing for implicit uses of `self` in situations where the user has already made their intent explicit, or where strong reference cycles are otherwise unlikely to occur. There are two situations covered by this proposal. The first is when the user has explicitly captured `self` in the closure's capture list, so that the following would compile without error:
```

dabrahams: Markdown with manual line breaks is more source-control-friendly, FWIW.

proposals/NNNN-implicit-self-explicit-capture.md
```diff
@@ -0,0 +1,189 @@
...
+    func execute(_ work: @escaping () -> Void) {
+        work()
+    }
+    func method() {
+        execute { [self] in
+            x += 1
+        }
+    }
+}
+```
```
dabrahams: That example really makes the case well!

proposals/NNNN-implicit-self-explicit-capture.md
```diff
@@ -0,0 +1,189 @@
...
+
+The error is also overly conservative—it will prevent the usage of implicit `self` even when it is very unlikely that the capture of `self` will somehow cause a reference cycle. With function builders in SwiftUI, the automatic resolution to the "make capture semantics explicit" error (and indeed, the fix-it that is currently supplied to the programmer) is just to slap a `self.` on wherever the compiler tells you to. While this likely fine when `self` has value semantics, building this habit could cause users to ignore the error and apply the fix-it in cases where it is not in fact appropriate.
```
dabrahams: Strong argument!

proposals/NNNN-implicit-self-explicit-capture.md
```diff
@@ -0,0 +1,189 @@
...
+```swift
+execute { // <- Fix-it: capture 'self' explicitly to enable implicit 'self' in this closure. Fix-it: insert '[self] in'
+    let foo = doFirstThing()
+    performWork(with: bar)
+    doSecondThing(with: foo)
+    cleanup()
+}
+```
+
+Furthermore, if the type of `self` has value semanics, we will not require any explicit usage of `self` (at the call/use site or in the capture list), so that if `self` were a `struct` or `enum` then the above could be written as simply:
```
dabrahams:
```suggestion
Second, if the type of `self` has value semanics, we will not require any explicit usage of `self` (at the call/use site or in the capture list), so that if `self` were a `struct` or `enum` then the above could be written as simply:
```

proposals/NNNN-implicit-self-explicit-capture.md
```diff
@@ -0,0 +1,189 @@
...
+    let foo = doFirstThing()
+    performWork(with: bar)
+    doSecondThing(with: foo)
+    cleanup()
+}
+```
+
+## Detailed design
+
+Whenever `self` is declared explicitly in an escaping closure's capture list, or when `self` is of a type with value semantics, any code inside that closure can use names which resolve to members of the enclosing type, without specifying `self.` explicitly. In nested closures, the *innermost* escaping closure must capture `self`, so the following code would be invalid:
```
dabrahams:
```suggestion
Whenever `self` is declared explicitly in an escaping closure's capture list, or its type is a value type, any code inside that closure can use names which resolve to members of the enclosing type, without specifying `self.` explicitly. In nested closures, the *innermost* escaping closure must capture `self`, so the following code would be invalid:
```

proposals/NNNN-implicit-self-explicit-capture.md
```diff
@@ -0,0 +1,189 @@
...
+
+```swift
+execute { [self] in
+    execute { // Fix-it: capture 'self' explicitly to enable implicit 'self' in this closure.
+        x += 1 // Error: reference to property 'x' in closure requires explicit use of 'self' to make capture semantics explicit. Fix-it: reference 'self.' explicitly.
+    }
+}
+```
+
+This new behavior will be unavailable when `self` is captured weakly (due to the optional type of `self` requiring special handling), but for closures with  `unowned(safe)` and `unowned(unsafe)` captures of `self` the programmer has clearly declared their intent to capture `self`, and so implicit `self` will be enabled in these contexts (as for strong captures).
```
dabrahams: The reasoning for not including "weak self" here is unclear.

proposals/NNNN-implicit-self-explicit-capture.md
```diff
@@ -0,0 +1,189 @@
...
+
+The existing errors and fix-its have their language updated accoringly to indicate that there are now multiple ways to resolve the error. In addition to the changes noted above, we will also have:
```
benrimmington: Typo: `accoringly`

```suggestion
The existing errors and fix-its have their language updated accordingly to indicate that there are now multiple ways to resolve the error. In addition to the changes noted above, we will also have:
```

proposals/NNNN-implicit-self-explicit-capture.md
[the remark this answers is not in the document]
Jumhyn: Opted for the second suggestion to be consistent throughout the proposal in discussing value types and not value semantics.

Jumhyn:
@dabrahams Thank you for the feedback! I've improved the wording in the places you've noted, and clarified the reasoning for excluding `weak self` at this point (along with calling it out in future directions).

tkremenek:
@Jumhyn this is short notice, but would you be OK with kicking this review off on Monday, October 28?

PR #1027, opened 2019-04-15
[SE-0258] Clarify that property delegates cannot start with a lowercase letter

DougGregor:


rjmccall:
Alright, since this was always intended to be a baseline for the proposal, I'll merge it in.

PR #1031, opened 2019-04-16
initialValue should have label

an0:


DougGregor:
This doesn't look correct? We use the label below.

an0:
That's exactly why we shouldn't use `_`, in order to keep the label.

DougGregor:
Hah, I read the patch backwards. Thank you!

PR #1032, opened 2019-04-19
Update SE-0249 design with core team feedback

stephencelis:
It was suggested that key path expressions should be evaluated rather than captured, so I'm updating the proposal with said feedback.

### DIFF
--- a/proposals/0249-key-path-literal-function-expressions.md
+++ b/proposals/0249-key-path-literal-function-expressions.md
@@ -62,7 +62,7 @@ The following are equivalent:
 ``` swift
 let f: (User) -> String = \User.email
 
-let g: (User) -> String = { kp in { $0[keyPath: \User.email } }(\User.email)
+let g: (User) -> String = { kp in { $0[keyPath: kp] } }(\User.email)
 ```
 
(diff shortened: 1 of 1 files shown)


rjmccall:
I hadn't really thought of this as requiring an amendment, but it's best to be clear.  I'll raise this with the Core Team tomorrow.

rjmccall:
The Core Team agrees that this change can happen without further review; we'll just ping the thread when it's applied.  We'd like some drafting changes first, though:

- The right way to describe the high-level semantics is that we evaluate the key path expression immediately and then capture the resulting `KeyPath` in a function which applies it.  (That doesn't have to be how we *implement* it, but the implementation should be consistent with those semantics.  The best implementation would probably be to simply capture any subscript values and then directly emit a reference to the storage in the function without building a key path.)

- It would be good to have an example which shows this semantic difference.

rjmccall:
@stephencelis Are you planning to revise this PR, or would you like me to just make the edits?  I don't want to ping the forum without this change already having been applied.

stephencelis:
@rjmccall Sorry about that, please feel free to make the edits! I won’t have time to get around to them for a bit.

rjmccall:
This can be closed; an equivalent edit was made in #1038.

PR #1036, opened 2019-04-25
Minor fix ages -> names

palimondo:
Fix the illogical variable name that tripped my concentration when reading the proposal… ;-)


lattner:
Thanks!

PR #1037, opened 2019-04-26
[Revision] SE-0253: Introduce callables.

rxwei:
Implementation: https://github.com/apple/swift/pull/24299

### DIFF
--- a/proposals/0253-callable.md
+++ b/proposals/0253-callable.md
@@ -1,4 +1,4 @@
-# User-defined non-function callable values
+# Callable values of user-defined nominal types
 
 * Proposal: [SE-0253](https://github.com/apple/swift-evolution/blob/master/proposals/0253-callable.md)
 * Authors: [Richard Wei](https://github.com/rxwei), [Dan Zheng](https://github.com/dan-zheng)
--- a/proposals/0253-callable.md
+++ b/proposals/0253-callable.md
@@ -344,9 +344,14 @@ finally a dynamic call.
 
 ### Implementation
 
-The implementation is very simple: [less than 200 lines of
+The implementation is very simple and non-invasive: [less than 200 lines of
 code](https://github.com/apple/swift/pull/24299) in the type checker that
 performs lookup and expression rewrite.
+
+```swift
+let add1 = Adder(base: 1)
+add1(0) // Rewritten to `add1.call(0)` after type checking.
+```
 
 ## Source compatibility
 
--- a/proposals/0253-callable.md
+++ b/proposals/0253-callable.md
@@ -3,8 +3,9 @@
 * Proposal: [SE-0253](https://github.com/apple/swift-evolution/blob/master/proposals/0253-callable.md)
 * Authors: [Richard Wei](https://github.com/rxwei), [Dan Zheng](https://github.com/dan-zheng)
 * Review Manager: [Chris Lattner](https://github.com/lattner)
-* Status: **Returned for revision**
-* Implementation: [apple/swift#23517](https://github.com/apple/swift/pull/23517)
+* Status: **Awaiting review**
+* Implementation: [apple/swift#24299](https://github.com/apple/swift/pull/24299)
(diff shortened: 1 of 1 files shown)


proposals/0253-callable.md
```diff
@@ -3,8 +3,9 @@
 * Proposal: [SE-0253](https://github.com/apple/swift-evolution/blob/master/proposals/0253-callable.md)
```
lattner: The proposal name is very generic, and Swift already has callable things.  How about "User Defined Callable Types"?

proposals/0253-callable.md
```diff
@@ -3,8 +3,9 @@
...
 * Authors: [Richard Wei](https://github.com/rxwei), [Dan Zheng](https://github.com/dan-zheng)
 * Review Manager: [Chris Lattner](https://github.com/lattner)
-* Status: **Returned for revision**
-* Implementation: [apple/swift#23517](https://github.com/apple/swift/pull/23517)
+* Status: **Awaiting review**
+* Implementation: [apple/swift#24299](https://github.com/apple/swift/pull/24299)
+* Previous Revisions: [[1](https://github.com/apple/swift-evolution/blob/36ea8be09508db9380949954d0c7a101fdb15226/proposals/0253-callable.md)]
 * Decision Notes: [Rationale](https://forums.swift.org/t/returned-for-revision-se-0253-static-callables/23290)
 
 ## Introduction
```
lattner: Similarly in the intro, Swift already has callable values.

lattner: " the keyword call:" -> not anymore

proposals/0253-callable.md
rxwei: I'm intentionally avoiding "callable types" because 1. `func call` is not restricted to primary type declarations and there's no type declaration attributes, and 2. _types_ being callable sound like initializers.

rxwei: I changed the title to "User-defined callable values". What do you think?

proposals/0253-callable.md
rxwei: Done.

proposals/0253-callable.md
lattner: Isn't fp a user defined callable value?
```
var fp : () -> () = {}
```

lattner:
A few other observations:

 - I'd recommend saying something about how simple and non-invasive the implementation is.

- "Implicit conversions to function" should mention that you can already explicitly convert them through partial application.

- I don't think that "static/class call methods" are a likely future direction personally, for exactly the reasons you say.  Why not drop that or move to alternatives considered?

- The first entry of 'alternatives considered' should be other names for this, including "call" "invoke" "apply" or other things that will come up in the thread.

proposals/0253-callable.md
rxwei: That's a good point. How about "non-function callable values"?

lattner: Metatypes.

Why not call this "Allowing user defined types to be callable"?

rxwei: Metatypes are not callable today actually, though I understand in the implementation they are.

```swift
let x = Float.self
x(2)
```
```console
error: repl.swift:2:2: error: initializing from a metatype value must reference 'init' explicitly
x(2)
 ^
 .init
```

rxwei: I changed it to "Callable values of user-defined nominal types" to be a bit more accurate. Hope that sounds okay to you!

rxwei:
> * I'd recommend saying something about how simple and non-invasive the implementation is.

Mentioned in the "detailed design" section.

> * "Implicit conversions to function" should mention that you can already explicitly convert them through partial application.

Done.

> * I don't think that "static/class call methods" are a likely future direction personally, for exactly the reasons you say.  Why not drop that or move to alternatives considered?

Moved to "alternatives considered".

> * The first entry of 'alternatives considered' should be other names for this, including "call" "invoke" "apply" or other things that will come up in the thread.

Done.

PR #1039, opened 2019-05-05
Fix: 0228-fix-expressiblebystringinterpolation.md

Nirma:
While reading through this file I noticed that the closing parentheses is missing in the example code.

I also sent over a small pull request to apple/swift to fix the comment in stdlib/public/core/StringInterpolation.swift as well.

https://github.com/apple/swift/pull/24509


lattner:
Thanks!

PR #1042, opened 2019-05-17
[SE-0244] Switch literal assignment to explicit Int 

airspeedswift:
Since `BinaryInteger` is literal-expressible, this example compiles because it doesn't assign an `Int`. Trying to assign an explicit `Int` does fail as described.


jckarter:
Thanks!

PR #1049, opened 2019-06-13
Revised proposal for part one of SE0243 - Character Literals.

johnno1962:
Hi Apple,

Re-submitting the ill-fated SE-0243 as the first of two parts in line with the review decision which indicated there seemed to be support for the syntax change. This part deals only with changing to single quote syntax not integer convertibility or any of the other ideas that were rejected during the review.

Thanks.

John


rjmccall:

PR #1051, opened 2019-06-15
Fixing old renamed property name.

DevAndArtist:
This PR fixes a few oversights in the proposal which otherwise would confuse the reader.

cc @DougGregor @rjmccall


DougGregor:
Thanks!

PR #1061, opened 2019-07-15
Add demangle function

Azoy:
As discussed here: https://forums.swift.org/t/demangle-function/25416


tkremenek:
This proposal will go to review on Monday, July 22.  @jckarter will be the review manager.

PR #1062, opened 2019-07-16
Add syntax highlighting to Real stub

gganley:
Very simple, this block of code lacks the "swift" after the triple grave that tells MD what syntax highlighting language to use.


DougGregor:
Thanks!

PR #1065, opened 2019-07-28
Add proposal: Prefix * Varargs Splat Operator for Arrays

owenv:
This is a revised version of the 'pass arrays as varargs proposal' to address core team & community feedback.

Pitch thread: [here](https://forums.swift.org/t/another-attempt-at-passing-arrays-as-varargs-with-implementation/26718/127)

Original implementation: [here](https://github.com/apple/swift/pull/25997)

If this new version looks like it might move forward, I can update the implementation to reflect the revised proposal.


owenv:
closing out old PRs

PR #1067, opened 2019-08-06
[Proposal] Make funcion parameter parentheses optional.

Nirma:
This is currently being discussed here:

https://forums.swift.org/t/pitch-allow-function-definitions-to-omit-parentheses-if-no-parameters


Coeur:
Multiple typos:
- funcion -> function
- proerties -> properties
- fucntion -> function
- ...

DougGregor:
The Core Team has determined that it will not make this fundamental change to the function declaration syntax.

PR #1068, opened 2019-08-16
Add the didSet Semantics proposal

theblixguy:
As discussed here: https://forums.swift.org/t/pitch-didset-semantics/27858

This has been on the todo list for a really long time!


theblixguy:
@jckarter @rjmccall would either of you want to manage this?

PR #1087, opened 2019-11-08
[Proposal] Discontiguous collection operations

natecook1000:
This proposal is the result of the discussion here: https://forums.swift.org/t/pitch-add-rangeset-and-related-collection-operations/29961


natecook1000:
I mixed up these branches somehow 🙃

PR #1088, opened 2019-11-08
[Proposal] Discontiguous collection operations

natecook1000:
This proposal is the result of the discussion here: https://forums.swift.org/t/pitch-add-rangeset-and-related-collection-operations/29961

### DIFF
--- a/proposals/NNNN-rangeset-and-collection-operations.md
+++ b/proposals/NNNN-rangeset-and-collection-operations.md
@@ -1,0 +1,616 @@
+# Add Collection Operations on Noncontiguous Elements
+
+* Proposal: [SE-NNNNNNNN](NNNN-rangeset-and-friends.md)
+* Author: [Nate Cook](https://github.com/natecook1000)
+* Review Manager: TBD
+* Status: **Awaiting review**
+* Implementation: [apple/swift#28161](https://github.com/apple/swift/pull/28161)
+
+## Introduction
+
+We can use a `Range<Index>` to refer to a group of consecutive positions in a collection, but the standard library doesn't currently provide a way to refer to discontiguous positions in an arbitrary collection. I propose the addition of a `RangeSet` type that can store any number of positions, along with collection algorithms that operate on those positions.
+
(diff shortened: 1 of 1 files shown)


proposals/NNNN-rangeset-and-collection-operations.md
```diff
@@ -0,0 +1,616 @@
...
+    /// comprises the range set fully contains `range`. This example shows
+    /// ranges that are and are not contained by a range set:
+    /// 
+    ///     let set: RangeSet = [0..<5, 10..<15]
+    ///     print(set.contains(1..<4))
+    ///     // Prints "true"
+    ///     print(set.contains(4..<7))
+    ///     // Prints "false"
+    ///     print(set.contains(4..<12))
+    ///     // Prints "false"
```
dabrahams: IMO this set of examples is unclear, in part because of the word "fully" above.  It leaves open what the answer would be for `set.contains(1..<5)`. I think you should at least drop "fully", because in no sense does `1..<3` "contain" `2..<4`.

proposals/NNNN-rangeset-and-collection-operations.md
```diff
@@ -0,0 +1,616 @@
...
+    /// Removes the given range from the range set.
+    public mutating func remove(_ range: Range<Bound>)
+}
+```
+
+`RangeSet` conforms to `Equatable`, to `CustomStringConvertible`, and, when its `Bound` type conforms to `Hashable`, to `Hashable`. `RangeSet` also has `ExpressibleByArrayLiteral` conformance, using arrays of ranges as its literal type.
+
+#### `SetAlgebra`-like methods
+
+`RangeSet` implements a subset of the `SetAlgebra` protocol. Because of its efficient representation for contiguous ranges, there is no way to insert a single element into a `RangeSet` .
```
dabrahams: If `RangeSet` is indeed a *set of ranges* as noted in the doc comment, then *elements* of the set have to be ranges. So isn't what you mean that there's no way to insert a range containing just one `Bound` value?

dabrahams:
It's also unclear from the proposal whether a RangeSet is empty after inserting an empty range.  I hope the answer is no, but I think the proposal should make the answer crystal clear.

PR #1091, opened 2019-11-11
Preview package implementation [3/3]

natecook1000:
This PR represents the initial contents of a `swift-evolution-staging` repository that would be created upon acceptance of the Swift Preview package proposal.


tkremenek:
SE-0264 will be accepted.  This PR is no longer needed.

PR #1097, opened 2019-12-02
[SE-0268] Amend proposal to include the deprecation of implicit oldValue

theblixguy:
Based on core team feedback, I have pulled the deprecation of the implicit `oldValue` into the proposed solution and updated the implementation to emit a warning notifying the user that the implicit `oldValue` is now deprecated, along with a fix-it to explicitly add `(oldValue)` to the `didSet`.

Apart from that, two minor tweaks:

1. Move the original bug associated with this proposal (SR-5982) to the top.
2. A small tweak to the language in the source compatibility section.


theblixguy:
cc @airspeedswift

PR #1104, opened 2019-12-18
[SE-0268] Amend proposal to remove the oldValue deprecation change

theblixguy:
This effectively reverts #1097 and removes the `oldValue` deprecation change from the proposal text based on core team decision.

Proposal acceptance announcement: https://forums.swift.org/t/accepted-se-0268-refine-didset-semantic/31822


theblixguy:
cc @airspeedswift

PR #1113, opened 2020-01-16
Amend changes to SE-0275

adellibovi:
As per title, I would like to amend some clarification for SE-0275

### DIFF
--- a/proposals/0275-allow-more-characters-like-whitespaces-and-punctuations-for-escaped-identifiers.md
+++ b/proposals/0275-allow-more-characters-like-whitespaces-and-punctuations-for-escaped-identifiers.md
@@ -14,19 +14,39 @@
 Swift has a beautiful concise yet expressive syntax.
 As part of that, escaped identifiers are adopted to allow usage of reserved keywords.
 This proposal wants to extend the character allowance for escaped identifiers with more Unicode scalars, like whitespace and punctuation.
-It will enable to have method names (or other identifiers) with a more readable and natural language like the following:
-```swift
-func `test validation should succeed when input is less than ten`()
-```
+
 ## Motivation
+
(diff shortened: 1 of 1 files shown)


proposals/0275-allow-more-characters-like-whitespaces-and-punctuations-for-escaped-identifiers.md
```diff
@@ -14,19 +14,39 @@
...
 ## Motivation
+
 Naming could be hard and having descriptive methods, like in tests, may result in declarations that are hard to read because of its lack of whitespace and punctuations or other symbols. Enabling natural language would improve readability.
 
+```swift
+func `test validation should succeed when input is less than ten`() // currently not possible
+// vs
+func `testValidationShouldSucceedWhenInputIsLessThanTen`()
+```
+
```
theblixguy: I think you should also add these examples:

```swift
func test_Validation_Should_Succeed_When_Input_IsLess_Than_Ten() {}
func test_validationShouldSucceed_whenInputIs_lessThanTen() {}
```
(add a few more variations if needed)

for a fairer comparison, because this is what a lot of people (including myself) currently write test functions as (and I think they are quite readable as well).

proposals/0275-allow-more-characters-like-whitespaces-and-punctuations-for-escaped-identifiers.md
```diff
@@ -52,7 +83,7 @@ identifier → ` identifier-head identifier-characters opt `
 with:
 ```
 identifier → ` escaped-identifier `
-escaped-identifier -> Any Unicode scalar value except U+000A or U+000D or U+0060
+escaped-identifier -> Any Unicode scalar value except U+000A, U+000B, U+000C, U+000D, U+0085, U+2028, U+2029 or U+0060 (back-tick)
```
theblixguy: Could you please add a small description next to the rest of the Unicode characters. For example: `U+0085 (nel)` or `U+0085 (next line)`, just like you did with U+0060.

PR #1120, opened 2020-02-05
Float16

stephentyrone:
First version of Float16 proposal for Swift Evolution.

### DIFF
--- a/proposals/0000-float16.md
+++ b/proposals/0000-float16.md
@@ -48,7 +48,7 @@ The entire API falls out from that. I.e. `Float16` will provide exactly the oper
 
 The nitty-gritty of implementation is slightly more involved; we need to update a few
 python files that drive `.gyb` generation, to introduce availability annotations for the type
-and operations (all the other floating-point types are unconditionally availble, so there's
+and operations (all the other floating-point types are unconditionally available, so there's
 a fair amount of minor churn associated with this change), and write implementations of
 a few operations that the compiler doesn't know how to do on its own (conversion from
 `Float80` to `Float16`, formatting `Float16` as a string, etc). All of this is essentially
--- a/proposals/0000-float16.md
+++ b/proposals/0000-float16.md
@@ -115,3 +115,10 @@ forcing all arithmetic to be performed in another format. However, this would me
 it would be much harder, in some cases, to get the same numerical result on the CPU and
 GPU (when GPU computation is performed in half-precision). It's a very nice convenience
 to be able to do a computation in the REPL and see what a GPU is going to do.
+
+Q: Why not put it in Swift Numerics?
+
+A: The biggest reason to add the type is for interoperability with C-family and GPU programs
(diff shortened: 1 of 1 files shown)


proposals/0000-float16.md
```diff
@@ -0,0 +1,134 @@
...
+Introduce the `Float16` type conforming to the `BinaryFloatingPoint` and `SIMDScalar`
+protocols, binding the IEEE 754 *binary16* format (aka *float16*, *half-precision*, or *half*),
+and bridged by the compiler to the C `_Float16` type.
+
+* Old pitch thread: [Add `Float16`](https://forums.swift.org/t/add-float16/19370).
+
+## Motivation
+
+The last decade has seen a dramatic increase in the use of floating-point types smaller
+than (32-bit) `Float`. The most widely implemented is `Float16`, which is used
```
benrimmington: It might be clearer to use `Float32` and `Float64` throughout.

```suggestion
than `Float32`. The most widely implemented is `Float16`, which is used
```

proposals/0000-float16.md
```diff
@@ -0,0 +1,134 @@
...
+struct Float16: BinaryFloatingPoint, SIMDScalar {
+  // ...
+}
+```
+The entire API falls out from that. I.e. `Float16` will provide exactly the operations that
+`Float` and `Double` and `Float80` do for their conformance to these protocols.
+
+The nitty-gritty of implementation is slightly more involved; we need to update a few
+python files that drive `.gyb` generation, to introduce availability annotations for the type
+and operations (all the other floating-point types are unconditionally availble, so there's
```
benrimmington: Typo: `availble` ➡️ `available`

```suggestion
and operations (all the other floating-point types are unconditionally available, so there's
```

proposals/0000-float16.md
```diff
@@ -0,0 +1,134 @@
...
+
+A: Alternative half-precision is no longer supported; ARMv8.2-FP16 only uses the IEEE 754
+encoding. Bfloat is something that we should *also* support eventually, but it's a separate
+proposal. and we should wait a little while before doing it. Conformance to IEEE 754 fully
+defines the semantics of `Float16` (and several hardware vendors, including Apple have
+been shipping processors that implement those semantics for a few years). By contrast,
+there are a few proposals for hardware that implements bfloat16; ARM and Intel designed
+different multiply-add units for it, and haven't shipped yet (and haven't defined any 
+operations *other* than a non-homogeneous multiply-add). Other companies have HW in
+use, but haven't (publically) formally specified their arithmetic. It's a moving target, and it
```
benrimmington: Typo: `publically` ➡️ `publicly`

```suggestion
use, but haven't (publicly) formally specified their arithmetic. It's a moving target, and it
```

proposals/0000-float16.md
stephentyrone: I think most people reviewing the proposal are more familiar with the names `Float` and `Double` (indeed, several people were surprised that the other names exist).

PR #1121, opened 2020-02-06
Record SE-110 as implemented in its current state

jckarter:
Update the proposal text to describe the special case for function arguments.


slavapestov:
LGTM!

PR #1127, opened 2020-02-24
Add support for Encoding and Decoding nested JSON keys

calda:
- Swift Evolution thread: [Pitch: Add support for Encoding and Decoding nested JSON keys](https://forums.swift.org/t/pitch-add-support-for-encoding-and-decoding-nested-json-keys/34039)
 - Prototype implementation: [diff](https://github.com/calda/NestedKeyEncodingStrategy/pull/1/files#diff-8ff2eba96e32f178462fed931f39208bR205) with [passing tests](https://github.com/calda/NestedKeyEncodingStrategy/blob/master/NestedKeyEncodingStrategyTests/NestedKeyEncodingStrategyTests.swift)

I can put up a PR against `apple/swift/stdlib/public/Darwin/Foundation/JSONEncoder.swift` with stdlib tests if the core team thinks this proposal is fitting for a formal review.

### DIFF
--- a/proposals/XXX-nested-json-keys.md
+++ b/proposals/XXX-nested-json-keys.md
@@ -1,0 +1,156 @@
+# Add support for Encoding and Decoding nested JSON keys
+
+* Authors: [Cal Stephens](https://github.com/calda)
+* Review Manager: TBD
+* Status: **Awaiting review**
+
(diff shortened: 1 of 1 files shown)


airspeedswift:
Hi @calda – thanks for the proposal. However, Swift Evolution only covers the Swift Standard Library, and `JSONEncoder` is part of Foundation. @parkera may be able to offer advice on how to propose this as a Foundation addition, but for now I'll close this evolution PR.

PR #1129, opened 2020-03-04
[SE-0280] Update the phrasing of the 'API resilience' section

theblixguy:
As mentioned by Jordan and Slava in the [review](https://forums.swift.org/t/se-0280-enum-cases-as-protocol-witnesses/34257) thread, switching between cases and static properties/methods can break resilience due to ABI and mangling differences. So, re-phrase the "API resilience" section to reflect that.


theblixguy:
cc @rjmccall

PR #1131, opened 2020-03-13
Narrative amendments for SE-0267

AnthonyLatsis:
These are some final corrections just to make sure the proposal is well-written.

cc @jckarter


jckarter:
Thanks @AnthonyLatsis!

PR #1134, opened 2020-03-26
Move goals from README.md to releases/swift-5_0.md

benrimmington:
The goals in README.md haven't been updated since August 2017.
Since the goals and release notes are available as blog posts,
add a simple table of links, which should be easier to update.


tkremenek:
Thank you so much for doing this!

PR #1135, opened 2020-03-27
Reset filter after click on Logo or Swift Evolution title

uxitten:
Imagine we choice some filter and after that, we expect reset all filter after a click on `Logo` or `Swift Evolution` title

### DIFF
--- a/index.html
+++ b/index.html
@@ -31,7 +31,7 @@ See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
         <div class="app">
             <header>
                 <span class="header-contents">
-                    <a id="logo" href="#" title="Swift Evolution" role="img" aria-label="Swift Evolution"></a>
+                    <a id="logo" href="." title="Swift Evolution" role="img" aria-label="Swift Evolution"></a>
                     <h1 id="title"><a href="#">Swift Evolution</a></h1>
                 </span>
             </header>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -31,8 +31,8 @@
         <div class="app">
             <header>
                 <span class="header-contents">
-                    <a id="logo" href="#" title="Swift Evolution" role="img" aria-label="Swift Evolution"></a>
-                    <h1 id="title"><a href="#">Swift Evolution</a></h1>
+                    <a id="logo" href="." title="Swift Evolution" role="img" aria-label="Swift Evolution"></a>
```
krilnon, line 34: Although from a coding perspective it’s expedient to do this, does it really make sense to reload the entire page just to toggle off the filter?

Seems like we should add a function for this, roughly: 

```javascript
if(!document.querySelector('.filter-toggle').classList.contains('hidden')) { toggleFiltering() }
```

benrimmington:
@krilnon Linking to the repository was [my suggestion](https://github.com/apple/swift-evolution/pull/592#issuecomment-274271980) from a few years ago, but I agree with [your reply](https://github.com/apple/swift-evolution/pull/592#issuecomment-274634696) that it would be too much of a context switch.

I think clicking the header (logo/title) of the status page should *clear* the filters, because you can already toggle them on/off by clicking the filter-button. But either way would be an improvement on the current behavior.

To continue your previous comparison, if you go to [help.github.com](https://help.github.com/) and type in the search field, the website address changes (`?query=example`) and the search results are displayed. Clicking the logo in the top-left corner resets both the page and website address.

index.html
uxitten: Currently, When clicking on logo/title, query string in URL has been removing but there are some filter still

krilnon: Hmm, fair point. Let’s just go with your original approach, then.

krilnon:
Alright, this seems like the way to go.

PR #1142, opened 2020-04-10
Add proposal for low-level atomics

lorentey:
This is based on the [Low-Level Atomic Operations](https://forums.swift.org/t/low-level-atomic-operations/34683) pitch thread on the Swift Evolution forum.

### DIFF
--- a/proposals/NNNN-atomics.md
+++ b/proposals/NNNN-atomics.md
@@ -397,7 +397,7 @@ extension UnsafeAtomic where Value == Int {
   func wrappingIncrement(
     by operand: Int,
     ordering: AtomicUpdateOrdering
-  ) -> Int {
+  ) {
     var done = false
     var current = load(ordering: .relaxed)
     while !done {
@@ -999,9 +999,9 @@ public func atomicMemoryFence(ordering: AtomicUpdateOrdering)
 
 @frozen
 public struct AtomicLoadOrdering: Equatable, Hashable, CustomStringConvertible {
-  public static var relaxed: AtomicLoadOrdering { get }
-  public static var acquiring: AtomicLoadOrdering { get }
-  public static var sequentiallyConsistent: AtomicLoadOrdering { get }
+  public static var relaxed: Self { get }
+  public static var acquiring: Self { get }
+  public static var sequentiallyConsistent: Self { get }
 
   public static func ==(left: Self, right: Self) -> Bool
   public func hash(into hasher: inout Hasher)
@@ -1010,9 +1010,9 @@ public struct AtomicLoadOrdering: Equatable, Hashable, CustomStringConvertible {
 
 @frozen
 public struct AtomicStoreOrdering: Equatable, Hashable, CustomStringConvertible {
-  public static var relaxed: AtomicStoreOrdering { get }
-  public static var releasing: AtomicStoreOrdering { get }
-  public static var sequentiallyConsistent: AtomicStoreOrdering { get }
+  public static var relaxed: Self { get }
+  public static var releasing: Self { get }
+  public static var sequentiallyConsistent: Self { get }
 
   public static func ==(left: Self, right: Self) -> Bool
   public func hash(into hasher: inout Hasher)
@@ -1021,11 +1021,11 @@ public struct AtomicStoreOrdering: Equatable, Hashable, CustomStringConvertible 
 
 @frozen
 public struct AtomicUpdateOrdering: Equatable, Hashable, CustomStringConvertible {
-  public static var relaxed: AtomicUpdateOrdering { get }
-  public static var acquiring: AtomicUpdateOrdering { get }
-  public static var releasing: AtomicUpdateOrdering { get }
-  public static var acquiringAndReleasing: AtomicUpdateOrdering { get }
-  public static var sequentiallyConsistent: AtomicUpdateOrdering { get }
+  public static var relaxed: Self { get }
+  public static var acquiring: Self { get }
+  public static var releasing: Self { get }
+  public static var acquiringAndReleasing: Self { get }
+  public static var sequentiallyConsistent: Self { get }
 
   public static func ==(left: Self, right: Self) -> Bool
   public func hash(into hasher: inout Hasher)
@@ -1396,7 +1396,7 @@ We defer these for future proposals.
 
 ### Default Orderings
 
-We considered providing defaulting all atomic operations throughout the `Atomics` module to sequentially consistent ordering. While we concede that doing so would make atomics slightly more approachable, implicit ordering values tend to interfere with highly performance-sensitive use cases of atomics (which is *most* use cases of atomics). Sequential consistency tends to be relatively rarely used in these contexts, and implicitly defaulting to it would allow accidental use to easily slip through code review.
+We considered defaulting all atomic operations throughout the `Atomics` module to sequentially consistent ordering. While we concede that doing so would make atomics slightly more approachable, implicit ordering values tend to interfere with highly performance-sensitive use cases of atomics (which is *most* use cases of atomics). Sequential consistency tends to be relatively rarely used in these contexts, and implicitly defaulting to it would allow accidental use to easily slip through code review.
 
 Users who wish for default orderings are welcome to define their own overloads for atomic operations:
 
@@ -1414,15 +1414,21 @@ extension UnsafeAtomic {
     exchange(desired, ordering: .sequentiallyConsistent)
   }
   
-  func compareExchange(expected: Value, desired: Value) -> Value {
+  func compareExchange(
+    expected: Value,
+    desired: Value
+  ) -> (exchanged: Bool, original: Value) {
     compareExchange(
       expected: expected, 
       desired: desired, 
       ordering: .sequentiallyConsistent)
   }
 
-  func compareExchangeWeak(expected: Value, desired: Value) -> Value {
-    compareExchangeWeak(
+  func weakCompareExchange(
+    expected: Value,
+    desired: Value
+  ) -> (exchanged: Bool, original: Value) {
+    weakCompareExchange(
       expected: expected, 
       desired: desired, 
       successOrdering: .sequentiallyConsistent,
@@ -1458,7 +1464,7 @@ struct UnsafeDemo {
     // Initialize a new storage instance by converting the given value.
     // The conversion may involve side effects such as unbalanced retain/release
     // operations; to ensure correct results, the resulting storage instance
-    // must be used to initialize exactly once to initialize a memory location.
+    // must be used to initialize exactly one memory location.
     init(_ initialValue: __owned Value)
     
     // Dispose of this storage instance, and return the last stored value.
@@ -1480,9 +1486,8 @@ struct UnsafeDemo {
   init(at address: UnsafeMutablePointer<Storage>)
   
   // Return a new instance by allocating and initializing a dynamic variable
-  // dedicated to holding its storage. Must be paired with a call to `destroy()`
-  // when 
-  static func create(initialValue: Value) -> UnsafeDemo
+  // dedicated to holding its storage. Must be paired with a call to `destroy()`.
+  static func create(initialValue: Value) -> Self
   
   // Destroy an instance previously created by `Self.create(initialValue:)`,
   // deinitializing and deallocating the dynamic variable that backs it,
--- a/proposals/NNNN-atomics.md
+++ b/proposals/NNNN-atomics.md
@@ -353,14 +353,14 @@ extension UnsafeAtomic {
   public func compareExchange(
     expected: Value,
     desired: __owned Value,
-    ordering: AtomicUpdateOrdering,
+    successOrdering: AtomicUpdateOrdering,
     failureOrdering: AtomicLoadOrdering
   ) -> (exchanged: Bool, original: Value)
 
   public func weakCompareExchange(
     expected: Value,
     desired: __owned Value,
-    ordering: AtomicUpdateOrdering,
+    successOrdering: AtomicUpdateOrdering,
     failureOrdering: AtomicLoadOrdering
   ) -> (exchanged: Bool, original: Value)
 }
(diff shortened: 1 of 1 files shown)


proposals/NNNN-atomics.md
```diff
@@ -0,0 +1,1702 @@
...
+The `weakCompareExchange` form may sometimes return false even when the original and expected values are equal. (Such failures may happen when some transient condition prevents the underlying operation from succeeding -- such as an incoming interrupt during a load-link/store-conditional instruction sequence.) This variant is designed to be called in a loop that only exits when the exchange is successful; in such loops using `weakCompareExchange` may lead to a performance improvement by eliminating a nested loop in the regular, "strong", `compareExchange` variants.
+
+The compare-exchange primitive is special: it is a universal operation that can be used to implement all other atomic operations, and more. For example, here is how we could use `compareExchange` to implement a wrapping increment operation over `UnsafeAtomic<Int>` values:
+
+```swift
+extension UnsafeAtomic where Value == Int {
+  func wrappingIncrement(
+    by operand: Int,
+    ordering: AtomicUpdateOrdering
+  ) -> Int {
```
benrimmington, line 401: No return value?

```suggestion
  ) {
```

proposals/NNNN-atomics.md
```diff
@@ -0,0 +1,1702 @@
...
+| `wrappingDecrementThenLoad(by:ordering:)`  | new value  |`a &-= b`   |
+| `bitwiseAndThenLoad(with:ordering)`        | new value  |`a &= b`    |
+| `bitwiseOrThenLoad(with:ordering)`         | new value  |`a \|= b`   |
+| `bitwiseXorThenLoad(with:ordering)`        | new value  |`a ^= b`    |
+| `wrappingIncrement(by:ordering:)`          | none   | `a &+= b` |
+| `wrappingDecrement(by:ordering:)`          | none   | `a &-= b` |
+
+The `wrappingIncrement` and `wrappingDecrement` operations are provided as a convenience for incrementing/decrementing values in the common case when a return value is not required.
+
+While we require all atomic operations to be free of locks, we don't require wait-freedom. Therefore, on architectures that don't provide direct hardware support some or all of these operations, we still require them to be implemented using `compareExchange` loops like the one for `wrappingIncrement` above.
```
benrimmington, line 437: Missing "for"?

```suggestion
While we require all atomic operations to be free of locks, we don't require wait-freedom. Therefore, on architectures that don't provide direct hardware support for some or all of these operations, we still require them to be implemented using `compareExchange` loops like the one for `wrappingIncrement` above.
```

proposals/NNNN-atomics.md
```diff
@@ -0,0 +1,1702 @@
...
+// Thread B: Store a new reference in the atomic unmanaged value and 
+//           release the previous reference.
+let new = Unmanaged.passRetained(...)
+let old = myAtomicRef.exchange(new, ordering: .acquiringAndReleasing)
+old.release() // RACE CONDITION
+```
+
+If thread B happens to release the same object that thread A is in the process of loading, then thread A's `takeUnretainedValue` may attempt to retain a deallocated object.
+
+Such problems make `AtomicUnmanaged` exceedingly difficult to use in all but the simplest situations. The section on [*Atomic Strong References*](#atomic-strong-references-and-the-problem-of-memory-reclamation) below describes some new constructs we may introduce in future proposals to assist with this issue.
```
benrimmington, line 489: `AtomicUnmanaged` ➡️ `UnsafeAtomic<Unmanaged<T>>`

```suggestion
Such problems make `UnsafeAtomic<Unmanaged<T>>` exceedingly difficult to use in all but the simplest situations. The section on [*Atomic Strong References*](#atomic-strong-references-and-the-problem-of-memory-reclamation) below describes some new constructs we may introduce in future proposals to assist with this issue.
```

proposals/NNNN-atomics.md
```diff
@@ -0,0 +1,1702 @@
...
+
+For now, we provide the standalone type `UnsafeAtomicLazyReference`; this is an example of a useful construct that could be built on top of `UnsafeAtomic<Unmanaged>` operations. Of all the various atomic constructs introduced in this proposal, only `UnsafeAtomicLazyReference` represents a regular strong reference to a class instance -- the other pointer/reference types leave memory management entirely up to the user.)
```
benrimmington, line 491: Add or remove parenthesis?

```suggestion
For now, we provide the standalone type `UnsafeAtomicLazyReference`; this is an example of a useful construct that could be built on top of `UnsafeAtomic<Unmanaged>` operations. (Of all the various atomic constructs introduced in this proposal, only `UnsafeAtomicLazyReference` represents a regular strong reference to a class instance -- the other pointer/reference types leave memory management entirely up to the user.)
```

proposals/NNNN-atomics.md
```diff
@@ -0,0 +1,1702 @@
...
+
+An `UnsafeAtomicLazyReference` holds an optional reference that is initially set to `nil`. The value can be set exactly once, but it can be read an arbitrary number of times. Attempts to change the value after the first `storeIfNilThenLoad` call are ignored, and return the current value instead.
```
benrimmington, line 493: `storeIfNilThenLoad` ➡️ `storeIfNil` to match implementation?

```suggestion
An `UnsafeAtomicLazyReference` holds an optional reference that is initially set to `nil`. The value can be set exactly once, but it can be read an arbitrary number of times. Attempts to change the value after the first `storeIfNil` call are ignored, and return the current value instead.
```

proposals/NNNN-atomics.md
```diff
@@ -0,0 +1,1702 @@
...
+    public mutating func dispose() -> Value
+  }
+
+  public init(at address: UnsafeMutablePointer<Storage>)
+
+  public static func create() -> Self
+  @discardableResult 
+  public func destroy() -> Value
+
+  public func storeIfNilThenLoad(_ desired: __owned Instance) -> Instance
```
benrimmington, line 512:
```suggestion
  public func storeIfNil(_ desired: __owned Instance) -> Instance
```

proposals/NNNN-atomics.md
```diff
@@ -0,0 +1,1702 @@
...
+```swift
+var _foo: UnsafeAtomicLazyReference<Foo> = ...
+
+// This is safe to call concurrently from multiple threads.
+var atomicLazyFoo: Foo {
+  if let foo = _foo.load() { return foo }
+  // Note: the code here may run concurrently on multiple threads.
+  // All but one of the resulting values will be discarded.
+  let foo = Foo()
+  return _foo.storeIfNilThenLoad(foo)
```
benrimmington, line 530:
```suggestion
  return _foo.storeIfNil(foo)
```

proposals/NNNN-atomics.md
```diff
@@ -0,0 +1,1702 @@
...
+}
+```
+
+The Standard Library has been internally using such a pattern to implement deferred bridging for `Array`, `Dictionary` and `Set`.
+
+Note that unlike the rest of the atomic types, `load` and `storeIfNilThenLoad(_:)` do not expose `ordering` parameters. (Internally, they map to acquiring/releasing operations to guarantee correct synchronization.)
```
benrimmington, line 536:
```suggestion
Note that unlike the rest of the atomic types, `load` and `storeIfNil(_:)` do not expose `ordering` parameters. (Internally, they map to acquiring/releasing operations to guarantee correct synchronization.)
```

proposals/NNNN-atomics.md
```diff
@@ -0,0 +1,1702 @@
...
+}
+
+@frozen
+struct AtomicUpdateOrdering {
+  static var relaxed: AtomicUpdateOrdering { get }
+  static var acquiring: AtomicUpdateOrdering { get }
+  static var releasing: AtomicUpdateOrdering { get }
+  static var acquiringAndReleasing: AtomicUpdateOrdering { get }
+  static var sequentiallyConsistent: AtomicUpdateOrdering { get }
+}
```
benrimmington, line 592: Use `Self` to match implementation?

```suggestion
@frozen
struct AtomicLoadOrdering {
  static var relaxed: Self { get }
  static var acquiring: Self { get }
  static var sequentiallyConsistent: Self { get }
}

@frozen
struct AtomicStoreOrdering {
  static var relaxed: Self { get }
  static var releasing: Self { get }
  static var sequentiallyConsistent: Self { get }
}

@frozen
struct AtomicUpdateOrdering {
  static var relaxed: Self { get }
  static var acquiring: Self { get }
  static var releasing: Self { get }
  static var acquiringAndReleasing: Self { get }
  static var sequentiallyConsistent: Self { get }
}
```

proposals/NNNN-atomics.md
```diff
@@ -0,0 +1,1702 @@
...
+        ordering: ordering)
+    }
+  }
+}
+
+// Elsewhere:
+counter.checkedIncrement(by: 1, ordering: .relaxed)
+```
+
+If for whatever reason the Swift compiler isn't able (or willing) to inline the `checkedIncrement` call, then the value of `ordering` won't be known at compile time to the body of the function, so even though `compileExchange` will still get inlined, its switch statement won't be eliminated. This leads to a potentially significant performance regression that could interfere with the scalability of the operation.
```
benrimmington, line 841: `compileExchange` ➡️ `compareExchange`

```suggestion
If for whatever reason the Swift compiler isn't able (or willing) to inline the `checkedIncrement` call, then the value of `ordering` won't be known at compile time to the body of the function, so even though `compareExchange` will still get inlined, its switch statement won't be eliminated. This leads to a potentially significant performance regression that could interfere with the scalability of the operation.
```

proposals/NNNN-atomics.md
```diff
@@ -0,0 +1,1702 @@
...
+
+There are a variety of approaches to tackle this problem, some of which may be general enough to consider in future proposals. (One potential solution can be built on top of double-wide atomic operations, by offloading some of the reference counting operations into the second word of a double-wide atomic reference.)
+
+(It'd be straightforward to use locks to build an atomic strong reference; while such a construct obviously wouldn't be lock-free, it is still a useful abstraction, so it may be a worthy addition to the Standard Library. However, locking constructs are outside the scope of this proposal.)
+
+### Additional Low-Level Atomic Features
+
+To enable usecases that require even more fine-grained control over atomic operations, it may be useful to introduce additional low-level atomics features:
+
+* support for additonal kinds of atomic values (such as double-wide atomics or floating-point atomics [[P0020]]),
```
benrimmington, line 1387: `additonal` ➡️ `additional`

```suggestion
* support for additional kinds of atomic values (such as double-wide atomics or floating-point atomics [[P0020]]),
```

proposals/NNNN-atomics.md
```diff
@@ -0,0 +1,1702 @@
...
+
+Note that this restriction wasn't introduced by our new low-level atomic primitives -- it is a preexisting property of the language.
+
+This is one of the reasons why `NaïveFuture` was declared as a class above. Class instance methods are allowed to mutate their stored properties without declaring themselves `mutating`, and thus they are outside the scope of the Law of Exclusivity. (Of course, their implementation must still guarantee that the Law is upheld for any variables they access.)
+
+> **Note:** A more fundamental reason why `NaïveFuture` is a class is that synchronization constructs are difficult to model with value types -- their instances tend to have an inherent identity that prevents copies from working like the original, they often need to be backed by a stable memory location, etc. The [Ownership Manifesto]'s non-copiable types may eventually provide a more efficient and safer model for such constructs, but in today's Swift, we need to represent them with some reference type instead: typically, either a class (like `NaïveFuture`) or some unsafe pointer type (like `UnsafeAtomic`).
```
benrimmington, line 922: The `NaïveFuture` class was removed (in version `2020-03-30/2`).

proposals/NNNN-atomics.md
```diff
@@ -0,0 +1,1702 @@
...
+```text
+warning: inout expression creates a temporary pointer, but argument 'at' should be 
+a pointer that outlives the call to 'init(at:)'
+    _value = UnsafeAtomic<Int>(at: &_storage)
+                                   ^~~~~~~~~
+```
+
+This is implemented using a preexisting diagnostic based on a compiler heuristic. Ideally this warning would be promoted to a compile-time error.
+
+> **Note:** For an idea on how to add proper language support for taking the address of certain kinds of variables, see the discussion on the hypothetical `@raw` attribute in [Memory\-Safe Atomic Constructs](#memory-safe-atomic-constructs).
```
benrimmington, line 985: `@raw` ➡️ `@addressable`

```suggestion
> **Note:** For an idea on how to add proper language support for taking the address of certain kinds of variables, see the discussion on the hypothetical `@addressable` attribute in [Memory\-Safe Atomic Constructs](#memory-safe-atomic-constructs).
```

proposals/NNNN-atomics.md
lorentey: This is quite intentional. There are separate variants for increments/decrements with a return value.

proposals/NNNN-atomics.md
lorentey: Ah, good eye -- I'm actually changing the implementation to match.

proposals/NNNN-atomics.md
benrimmington: So, should the `-> Int` be removed for this variant?

lorentey: 🤦‍♂️

proposals/NNNN-atomics.md
```diff
@@ -0,0 +1,1701 @@
...
+
+```swift
+struct UnsafeDemo {
+  typealias Value
+
+  struct Storage {
+    // Initialize a new storage instance by converting the given value.
+    // The conversion may involve side effects such as unbalanced retain/release
+    // operations; to ensure correct results, the resulting storage instance
+    // must be used to initialize exactly once to initialize a memory location.
```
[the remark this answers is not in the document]
lorentey: Ah, this is garbled, what I want to say is

```suggestion
    // must be used to initialize exactly one memory location.
```

lorentey:
@benrimmington thanks for all the edits!

PR #1144, opened 2020-04-16
Revise SE-0279 to propose an alternative syntax

rjmccall:


proposals/0279-multiple-trailing-closures.md
```diff
@@ -1,422 +1,369 @@
...
 
-It's possible to do an early syntax transformation which would consider each block to be a regular argument and adjust locations of parentheses if necessary.
-For type checker perspective this means that calls with new syntax would just be regular calls which require no special handling:
+Note that labeled trailing closures are required to match labels with a parameter.  A labeled trailing closure can use the special label `_` to indicate that it matches an unlabeled parameter, but this *only* matches an unlabeled parameter; it does not have the special label-omission powers of the initial unlabeled trailing closure.  For example:
 
 ```swift
-when(2 < 3) {
-  then: { ... }
-  else: { ... }
+func pointFromClosures(x: () -> Int, _ y: () -> Int) -> (Int, Int) { (x,y) }
```
benrimmington, line 205: * Call `x()` and `y()` in body
* Reformat as it's now longer than 80 characters

```suggestion
func pointFromClosures(
  x: () -> Int,
  _ y: () -> Int
) -> (Int, Int) {
  (x(), y())
}
```

proposals/0279-multiple-trailing-closures.md
```diff
@@ -1,422 +1,369 @@
...
+pointFromClosures { 10 } _: { 20 }  // Ok
+
+func performAsync(action: @escaping () -> (),
+                  completionOnMainThread: @escaping () -> ()) { ... }
```
benrimmington, line 209: * Return `-> Void` instead of `-> ()`

```suggestion
func performAsync(
  action: @escaping () -> Void,
  completionOnMainThread: @escaping () -> Void
) {
  ...
}
```

proposals/0279-multiple-trailing-closures.md
```diff
@@ -1,422 +1,369 @@
...
 ```swift
-when(2 < 3,
-  then: { ... },
-  else: { ... }
-)
+func when<T>(_ condition: @autoclosure () -> Bool,
+             then: () -> T,
+             `else`: () -> T) -> T {
+  ...
+}
```
benrimmington, line 228: * Add the ternary expression

```suggestion
func when<T>(
  _ condition: @autoclosure () -> Bool,
  then: () -> T,
  `else`: () -> T
) -> T {
  condition() ? then() : `else`()
}
```

proposals/0279-multiple-trailing-closures.md
```diff
@@ -1,422 +1,369 @@
...
 }
 ```
 
-Is equivalent to:
+is equivalent to:
 
-```swift
-foo({ ... }, { ... })
+```
+`when(2 < 3, then: { print("then") }, else: { print("else") })`
```
benrimmington, line 244: * Enable syntax highlighting
* Remove single backticks

```suggestion
```swift
when(2 < 3, then: { print("then") }, else: { print("else") })
```

proposals/0279-multiple-trailing-closures.md
```diff
@@ -1,422 +1,369 @@
...
-This feature is implementable entirely in the parser, as a syntactic transformation on call expressions. It, therefore, has no impact on the ABI.
+foo {
+  42
+} c: {
+  ...
+}
+```
 
-## Effect on API resilience
+Since new type-checking rule dictates a backwards scan starting for the last (labeled) trailing closure before attempting to match an unlabeled argument, this call is equivalent to a following "old" syntax:
```
benrimmington, line 275: * Add missing "the"

```suggestion
Since the new type-checking rule dictates a backwards scan starting for the last (labeled) trailing closure before attempting to match an unlabeled argument, this call is equivalent to a following "old" syntax:
```

proposals/0279-multiple-trailing-closures.md
```diff
@@ -1,422 +1,369 @@
...
-  },
-  else: {
-    ...
-    ...
-  }
-)
+func resolve(id: UUID,
+             action: (Object) -> (),
+             completion: (() -> ())? = nil,
+             onError: ((Error) -> ())? = nil)
```
benrimmington, line 289: * Return `-> Void` instead of `-> ()`

```suggestion
func resolve(
  id: UUID,
  action: (Object) -> Void,
  completion: (() -> Void)? = nil,
  onError: ((Error) -> Void)? = nil
) {
  ...
}
```

proposals/0279-multiple-trailing-closures.md
```diff
@@ -1,422 +1,369 @@
...
-  ...
-}
-label {   // call to a function named "label", not part of the Button creation
-  Text("Hello!")
+UIView.animate(withDuration: 0.3) {
+  self.view.alpha = 0
+} completion: { _ in
+  self.view.removeFromSuperview()
 }
+`
```
benrimmington, line 337: * Remove single backtick

```suggestion
```

proposals/0279-multiple-trailing-closures.md
```diff
@@ -1,422 +1,369 @@
...
-  action {
-    ...
-    ...
-    ...
-  }
+The proposed syntax could be extended to allow users to optionally label the first trailing closure:
 
-  label {
-    Text("Hello!")
+```
```
benrimmington, line 344: * Enable syntax highlighting

```suggestion
```swift
```

proposals/0279-multiple-trailing-closures.md
```diff
@@ -1,422 +1,369 @@
...
-[clarity at the point of use](https://swift.org/documentation/api-design-guidelines/#fundamentals) for
-many existing APIs, without requiring the API authors to make changes.
+This introduces a zero-lookahead ambiguity between the start of a labeled trailing closure and the start of either a new expression, a labeled statement, or a `default:` label followed by `'{'`.  The first two ambiguities can be resolved by looking forward at most two tokens, because (`identifier|keyword|'_') ':'` can never start an expression, only a labeled statement, and the statement in a labeled statement can never start with `'{'` while `expr-closure` must start with it.  The ambiguity with `default:` can be resolved by not allowing the unescaped `default` keyword as a label in this syntax; it can still be used if necessary by escaping it (i.e. \``default`\``:`).  Source compatibility requires the existing use of `default:` to be preferred, and it's better to do this uniformly (even if the syntax does not appear in a `switch`) in order to discourage the use of `default` as a trailing-closure label in APIs, rather than leaving a trap in the language if an API like that is used within a `switch`.
```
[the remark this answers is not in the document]
benrimmington: Two changes:

* (`identifier|keyword|'_') ':'` ➡️ `(identifier|keyword|'_') ':'`
* (i.e. \``default`\``:`) ➡️ (i.e. `` `default`: ``)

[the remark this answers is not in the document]
rjmccall: I spend quite a bit of time failing to get that to work, thanks!

benrimmington:
@rjmccall There are two unresolved suggestions (hidden above). To incorporate multiple suggestions into one commit, you can use *Add suggestion to batch* on the *Files changed* tab (but not on this *Conversation* tab).

rjmccall:
> @rjmccall There are two unresolved suggestions (hidden above). To incorporate multiple suggestions into one commit, you can use _Add suggestion to batch_ on the _Files changed_ tab (but not on this _Conversation_ tab).

Thanks, sorry.  In the future I'll batch the changes into a commit.

proposals/0279-multiple-trailing-closures.md
```diff
@@ -1,422 +1,383 @@
...
     ...
   }
-}
+  ```
+
+When using multiple trailing closure syntax, these APIs are all [clear at the point of use](https://swift.org/documentation/api-design-guidelines/#fundamentals), without the need to label the first trailing closure.
+
+If labelling the first trailing closure were allowed, users would have to evaluate whether to include the label on a case by case basis, which would inevitably lead to linter and style guide rules to prohibit it. So, in conjunction with the new syntax rules, we propose an amendment to the [API Design Guidelines](https://swift.org/documentation/api-design-guidelines):
+
+> Name functions assuming that the argument label of the first trailing closure will be dropped. Include meaningful argument labels for all subsequent trailing closures.
```
anandabits, line 177: Can you add a discussion of how existing standard library API that doesn't follow this convention will be handled?  `Sequence.first(where:)` is a good example of this.

rjmccall: Pinging @airspeedswift for that part.

airspeedswift: `drop(while:)` is probably the urbug in the std lib.

airspeedswift: But going back and fixing existing issues (as opposed to setting a new standard) should be left to a separate discussion rather than mixed into this one.

airspeedswift: Applying this advice would allow us to amend and reintroduce `countWhere` as well.

anandabits: Right, but I think the proposal should make it clear that this is a viable future direction if that’s the case.  One sentence would be enough.

airspeedswift: IMO it would be out of place to suggest only-tangentially-related proposals in this proposal text.

PR #1153, opened 2020-05-28
[Proposal] ExpressibleByTupleLiteral

JonnyGamer:
Added `ExpressibleByTupleLiteral` protocol idea.
This would be extremely magical.

```
struct Foo: ExpressibleByTupleLiteral {
     typealias FloatLiteralType = (Int, String)
     init(tupleLiteral value: (Int, String)) { }
}
```
```
let bar: Foo = (1_000, "foo")
```


DougGregor:
This belongs in a forum discussion, not as an addition to Future Directions for an accepted proposal.

PR #1165, opened 2020-07-17
Forward-scan matching for trailing closures

DougGregor:


proposals/NNNN-forward-scan-trailing-closures.md
```diff
@@ -0,0 +1,295 @@
...
+
+However, the backward-scan matching rule makes it hard to write good API that uses trailing closures, especially multiple trailing closures. This proposal replaces the backward scan with a forward scan, which is simpler, more in line with normal argument matching in a call, and works better for APIs that support trailing closures (whether single or multiple) and default arguments. This change introduces a *minor source break* for code involving multiple, defaulted closure parameters, but the practical impacts of this source break are low, and in its current state SE-0279 multiple trailing closures aren't very useful.
+
+## Motivation
+
+Several folks noted the downsides of the "backward" matching rule. The rule itself is described in the [detailed design](https://github.com/apple/swift-evolution/blob/master/proposals/0279-multiple-trailing-closures.md#detailed-design) section of SE-0279 (search for "backward"). To understand the problem with the backward rule, let's try to declare the UIView [`animate(withDuration:animation:completion:)`](https://developer.apple.com/documentation/uikit/uiview/1622515-animate) method in the obvious way to make use of SE-0279:
```
benrimmington, line 19: `animation:` → `animations:`

```suggestion
Several folks noted the downsides of the "backward" matching rule. The rule itself is described in the [detailed design](https://github.com/apple/swift-evolution/blob/master/proposals/0279-multiple-trailing-closures.md#detailed-design) section of SE-0279 (search for "backward"). To understand the problem with the backward rule, let's try to declare the UIView [`animate(withDuration:animations:completion:)`](https://developer.apple.com/documentation/uikit/uiview/1622515-animate) method in the obvious way to make use of SE-0279:
```

proposals/NNNN-forward-scan-trailing-closures.md
```diff
@@ -0,0 +1,295 @@
...
+UIView.animate(withDuration: 0.3) {
+  self.view.alpha = 0
+}
+```
+
+Now, the backward rule matches the unnamed trailing closure to `completion:`. The compiler produces an error:
+
+```
+error: missing argument for parameter 'animations' in call
+  animate(withDuration: 3.1415) {
```
benrimmington, line 53:
```suggestion
error: missing argument for parameter 'animations' in call
  animate(withDuration: 0.3) {
```

proposals/NNNN-forward-scan-trailing-closures.md
```diff
@@ -0,0 +1,295 @@
...
+```
+
+Note that the "real" UIView API actually has two different methods---`animate(withDuration:animation:completion:)` and `animate(withDuration:animation:)`---where the latter looks like this:
```
benrimmington, line 56: `animation:` → `animations:`

```suggestion
Note that the "real" UIView API actually has two different methods---`animate(withDuration:animations:completion:)` and `animate(withDuration:animations:)`---where the latter looks like this:
```

proposals/NNNN-forward-scan-trailing-closures.md
```diff
@@ -0,0 +1,295 @@
...
+    withDuration duration: TimeInterval, 
+    animations: @escaping () -> Void
+)
+```
+
+This second overload only has a closure argument, so the backward-matching rule handles the single-trailing-closure case. These overloads exist because these UIView APIs were imported from Objective-C, which does not have default arguments. A new Swift API would not be written this way---except that SE-0279 forces it due to the backward-matching rule.
+
+## Proposed solution
+
+The idea of the forward-scan matching rule is to match trailing closure arguments to parameters in the same forward, left-to-right manner that other arguments are matched to parameters. The unlabeled trailing closure will be matched to the next parameter that is either unlabeled or has a declared type that structurally resembles a function type (defined below). For example above, this means the following:
```
benrimmington, line 69: `For example above` → `For the example above`

```suggestion
The idea of the forward-scan matching rule is to match trailing closure arguments to parameters in the same forward, left-to-right manner that other arguments are matched to parameters. The unlabeled trailing closure will be matched to the next parameter that is either unlabeled or has a declared type that structurally resembles a function type (defined below). For the example above, this means the following:
```

proposals/NNNN-forward-scan-trailing-closures.md
```diff
@@ -0,0 +1,295 @@
...
+  self.view.alpha = 0
+} completion: { _ in
+  self.view.removeFromSuperview()
+}
+// equivalent to
+UIView.animate(withDuration: 0.3, animation: {
+    self.view.alpha = 0
+  }, completion: { _ in
+  self.view.removeFromSuperview()
+  })
```
benrimmington, line 94:
```suggestion
UIView.animate(withDuration: 0.3, animation: {
  self.view.alpha = 0
}, completion: { _ in
  self.view.removeFromSuperview()
})
```

proposals/NNNN-forward-scan-trailing-closures.md
```diff
@@ -0,0 +1,295 @@
...
+Note that the unlabeled trailing closure matches `animations:` in both cases; specifying additional trailing closures fills out later parameters but cannot shift the unlabeled trailing closure to an earlier parameter.
+
+Note that you can still have the unlabeled trailing closure match a later parameter, by specifying earlier ones:
+
+```swift
+UIView.animate(withDuration: 0.3, animations: self.doAnimation) { _ in
+  self.view.removeFromSuperview()
+}
+// equivalent to
+UIView.animate(withDuration: 0.3, animations: self.doAnimation, completion: _ in
```
benrimmington, line 106:
```suggestion
UIView.animate(withDuration: 0.3, animations: self.doAnimation, completion: { _ in
```

proposals/NNNN-forward-scan-trailing-closures.md
```diff
@@ -0,0 +1,295 @@
...
+* Remove all outer "optional" types.
+
+Following this rule, the `withDuration` parameter (a `TimeInterval`) does not resemble a function type. However, `@escaping () -> Void` does, so the unlabeled trailing closure matches `animations`. `@autoclosure () -> ((Int) -> Int)` and `((Int) -> Int)?` would also resemble a function type.
+
+### Mitigating the source compatibility impact
+
+The forward-scanning rule, as described above, is source-breaking. A run over Swift's [source compatibility suite](https://swift.org/source-compatibility/) with this change enabled in all language modes turned up source compatibility breaks in three projects. The first problem occurs with a SwiftUI API [`View.sheet(isPresented:onDismiss:content:)`](https://developer.apple.com/documentation/swiftui/view/sheet(ispresented:ondismiss:content:)):
+
+```
+func sheet(isPresented: Binding<Bool>, onDismiss: (() -> Void)? = nil, content: @escaping () -> Content) -> some View
```
benrimmington, line 163:
```suggestion
```swift
func sheet(
  isPresented: Binding<Bool>,
  onDismiss: (() -> Void)? = nil,
  content: @escaping () -> Content
) -> some View
```

proposals/NNNN-forward-scan-trailing-closures.md
```diff
@@ -0,0 +1,295 @@
...
+```
+
+Note that `onDismiss` and `content` both structurally resemble a function type. This API fits well with the backward-matching rule, because the unlabeled trailing closure in the following example is always ascribed to `content:`. The `onDismiss:` argument gets the default argument `nil`:
+
+```swift
+sheet(isPresented: $isPresented) { Text("Hello") }
+```
+
+With the forward-scanning rule, the unlabeled trailing closure matches the `onDismiss:` parameter, and there is no suitable argument for `content:`. Therefore, the well-formed code above would be rejected by the riule as proposed above.
```
benrimmington, line 172: `riule` → `rule`

```suggestion
With the forward-scanning rule, the unlabeled trailing closure matches the `onDismiss:` parameter, and there is no suitable argument for `content:`. Therefore, the well-formed code above would be rejected by the rule as proposed above.
```

proposals/NNNN-forward-scan-trailing-closures.md
```diff
@@ -0,0 +1,295 @@
...
+
+However, it is clear from the function signature that (1) `onDismiss:` could have used the default argument, and (2) `content:` therefore won't have an argument if it is not paired with the unlabeled trailing closure. We can turn this into an heuristic to accept more existing code, reducing the source breaking impact of the proposal. Specifically, if
+
+* the parameter that would match the unlabeled trailing closure
+argument does not requirement an argument (because it is variadic or has a default argument), and
```
benrimmington, line 177: `requirement ` → `require`

```suggestion
argument does not require an argument (because it is variadic or has a default argument), and
```

proposals/NNNN-forward-scan-trailing-closures.md
```diff
@@ -0,0 +1,295 @@
...
+
+This heuristic is remarkably effective: in addition to fixing 2 of the 3 failures from the Swift source compatibility suite (the remaining failure will be discussed below), it resolved 17 of the 19 failures we observed in a separate (larger) testsuite comprising a couple of million lines of Swift.
+
+## Source compatibility
+
+Even with the heuristic, the forward-scan matching rule will still fail to compile some existing and can change the meaning of some code, when there are multiple, defaulted parameters of closure type. As an example, the remaining source compatibility failure in the Swift source compatibility suite, a project called [ModelAssistant](https://github.com/ssamadgh/ModelAssistant), is due to [this API](https://github.com/ssamadgh/ModelAssistant/blob/c96335280a3aba5f8e14955ecaf38dc25a0872b6/Source/Libraries/AOperation/Observers/BlockObserver.swift#L22-L26):
```
benrimmington, line 186: `existing` → `existing code,`

```suggestion
Even with the heuristic, the forward-scan matching rule will still fail to compile some existing code, and can change the meaning of some code, when there are multiple, defaulted parameters of closure type. As an example, the remaining source compatibility failure in the Swift source compatibility suite, a project called [ModelAssistant](https://github.com/ssamadgh/ModelAssistant), is due to [this API](https://github.com/ssamadgh/ModelAssistant/blob/c96335280a3aba5f8e14955ecaf38dc25a0872b6/Source/Libraries/AOperation/Observers/BlockObserver.swift#L22-L26):
```

proposals/NNNN-forward-scan-trailing-closures.md
```diff
@@ -0,0 +1,295 @@
...
+
+```swift
+init(startHandler: ((AOperation) -> Void)? = nil, produceHandler: ((AOperation, Foundation.Operation) -> Void)? = nil, finishHandler: ((AOperation, [NSError]) -> Void)? = nil) {
```
benrimmington, line 189:
```suggestion
init(
    startHandler: ((AOperation) -> Void)? = nil,
    produceHandler: ((AOperation, Foundation.Operation) -> Void)? = nil,
    finishHandler: ((AOperation, [NSError]) -> Void)? = nil
) {
```

proposals/NNNN-forward-scan-trailing-closures.md
```diff
@@ -0,0 +1,295 @@
...
+```
+
+The forward-scan matching rule provides more predictable results, making it easier to understand how to use this API properly. 
+
+### Workaround via overload sets
+
+APIs like the above that will be broken by the forward scan can be "fixed" by removing the default arguments, then adding additional overloads to create the same effect. For example, drop the default argument of `finishHandler` so that the heuristic will kick in to fix calls with a single unlabeled trailing closure:
+
+```swift
+init(startHandler: ((AOperation) -> Void)? = nil, produceHandler: ((AOperation, Foundation.Operation) -> Void)? = nil, finishHandler: ((AOperation, [NSError]) -> Void)) {
```
benrimmington, line 263:
```suggestion
init(
    startHandler: ((AOperation) -> Void)? = nil,
    produceHandler: ((AOperation, Foundation.Operation) -> Void)? = nil,
    finishHandler: ((AOperation, [NSError]) -> Void)?
) {
```

proposals/NNNN-forward-scan-trailing-closures.md
```diff
@@ -0,0 +1,295 @@
...
+    self.produceHandler = produceHandler
+    self.finishHandler = finishHandler
+}
+```
+
+One can then add overloads to handle other cases, e.g., the zero-argument case:
+
+```swift
+init() {
+  self.init(startHandler: nil, produceHandler: nil, finishHandler:nil)
```
benrimmington, line 274:
```suggestion
  self.init(startHandler: nil, produceHandler: nil, finishHandler: nil)
```

proposals/NNNN-forward-scan-trailing-closures.md
```diff
@@ -0,0 +1,295 @@
...
+}
+```
+
+An investation into APIs with multiple trailing closures uncovered one SwiftUI API that will break in this manner, because both closure parameters have defaults:
```
benrimmington, line 278: `investation` → `investigation`

```suggestion
An investigation into APIs with multiple trailing closures uncovered one SwiftUI API that will break in this manner, because both closure parameters have defaults:
```

proposals/NNNN-forward-scan-trailing-closures.md
```diff
@@ -0,0 +1,295 @@
...
+
+```swift
+extension TextField where Label == SwiftUI.Text {
+  public init(_ titleKey: SwiftUI.LocalizedStringKey, text: SwiftUI.Binding<Swift.String>, onEditingChanged: @escaping (Swift.Bool) -> Swift.Void = { _ in }, onCommit: @escaping () -> Swift.Void = {})
+}
```
benrimmington, line 283:
```suggestion
extension TextField where Label == Text {
  public init(
    _ titleKey: LocalizedStringKey,
    text: Binding<String>,
    onEditingChanged: @escaping (Bool) -> Void = { _ in },
    onCommit: @escaping () -> Void = {}
  )
}
```

DougGregor:
Thank you @benrimmington !

PR #1171, opened 2020-08-24
[Proposal] Improve Result with .success and recoverFailure

hfhbd:
Hey, 
I want to propose these changes to the Result type.

Implementations: https://github.com/apple/swift/pull/27908, https://github.com/apple/swift/pull/26471

Swift-evolution threads: [Never-failing Result Type](https://forums.swift.org/t/never-failing-result-type/30249/5), [Convenience member on Result when when Success is Void](https://forums.swift.org/t/convenience-member-on-result-when-when-success-is-void/36134)

But I have a problem to answer the question regarding the ABI resilience:
What does this changes do, 2 public functions and 1 static var via extensions to the frozen enum, to the ABI?

### DIFF
--- a/proposals/Results-Improvements.md
+++ b/proposals/Results-Improvements.md
@@ -1,0 +1,76 @@
+# Feature name
+
+* Proposal: [SE-NNNN](NNNN-Results-Improvements.md)
+* Authors: [hfhbd](https://github.com/hfhbd), [Author 2](https://github.com/swiftdev)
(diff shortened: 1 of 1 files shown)


DougGregor:
The Core Team discussed this and has determined that we do not want to bring this proposal up for review while Swift Concurrency is under discussion and may change the way in which Swift code interacts with `Result`. These can be reconsidered at a later time.

PR #1180, opened 2020-09-08
[Proposal] Extend Property Wrappers to Function and Closure Parameters

filip-sakel:


proposals/XXXX-extend-property-wrappers-to-functions-and-closures.md
```diff
@@ -0,0 +1,391 @@
...
+## Alternatives Considered
+
+**TBD**
+
+
+## Future Directions
+
+### Property Wrappers Applied to Local Declarations
+
+Should this be included? Property wrappers seem to require more structure than is provided in local variable declarations. My concern is that local property wrappers could be misused taking property wrappers' simplicity away.
```
hborla, line 390: Local property wrappers were proposed and accepted (but not implemented) as part of the original property wrapper proposal, so they don't need to be included here. Though, the implementation for property wrapper parameters makes implementing local property wrappers pretty straightforward 🙂

proposals/XXXX-extend-property-wrappers-to-functions-and-closures.md
```diff
@@ -0,0 +1,391 @@
...
+    Becomes:
+
+```swift
+func c(_foo: WrapperObject<Int>) {
+    var foo: Int { 
+        _foo.wrappedValue 
+    }
+}
+
+// with the overload:
```
hborla, line 225: I think it's important here to not show the transformation as the compiler generating overloads of the function, but rather showing how calling the function is transformed to insert the `init(wrappedValue:)` call around the property wrapper (wrapped value) argument, e.g.

```swift
func fn(@WrapperObject value: Int) { ... }

fn(value: 10)
```

Becomes

```swift
func fn(value _value: WrapperObject<Int>) { ... }

fn(value: WrapperObject(wrappedValue: 10))
```

This is how the transformation is implemented, but it also better fits the mental model of initializing a property wrapper from a wrapped value, rather than thinking about it as the compiler choosing an overload.

proposals/XXXX-extend-property-wrappers-to-functions-and-closures.md
```diff
@@ -0,0 +1,391 @@
+# Extend Property Wrappers to Functions and Closures
```
hborla, line 1: Nit: Let's change the title slightly to "Extend Property Wrappers to Function and Closure Parameters" to clarify that we aren't proposing property wrappers on functions themselves

proposals/XXXX-extend-property-wrappers-to-functions-and-closures.md
```diff
@@ -0,0 +1,391 @@
...
+) { ... }
+
+struct Foo {
+    @Wrapper
+    var count = 0
+    
+    func bar() {
+        foo(count: $count)
+        
+        fooInClosure { @Wrapper count in
```
theblixguy, line 89: Would it be possible to support attaching it to the closure parameter at declaration site like you do for non-closure parameters? For example - `fooInClosure(@Wrapper block: (Wrapper<Int>) -> Int)`, because it would allow a user to use anonymous parameters and access count via `$0` for example.

hborla: That wouldn't work if you have multiple property wrappers parameters in the function type. Listing the wrapper attributes for different parameters next to each other before the parameter also wouldn't mesh well with property wrapper composition.

I'm against inferring the property wrapper attribute on closure parameters for several reasons. I also think it's useful to not be forced into the property wrapper syntax. However, I think we should discuss design decisions on the forums

hborla: Also, I would _really_ hate to have to refer to the backing wrapper and projected value with `_$0` and `$$0` 😆

theblixguy: Oh, okay! It might be good to add an example involving anonymous parameters and mentioning that it would make accessing the backing and projected value a bit awkward :)

proposals/XXXX-extend-property-wrappers-to-functions-and-closures.md
```diff
@@ -0,0 +1,391 @@
...
+
+* Proposal: [SE-NNNN](NNNN-extend-property-wrappers-to-functions-and-closures.md)
+* Authors: [Holly Borla](https://github.com/hborla), [Filip Sakel](https://github.com/filip-sakel)
+* Review Manager: TBD
+* Status: **Awaiting implementation**
+
+
+## Introduction
+
+Property Wrappers were [introcuded in Swift 5.1](https://github.com/apple/swift-evolution/blob/master/proposals/0258-property-wrappers.md), and have since become a popular feature abstracting away common accessor patterns for properties. Currently, applying a property wrapper is solely permitted on properties inside of a type context. However, with increasing adoption demand for extending _where_ property wrappers can be applied has emerged.
```
mpapple-swift, line 11: 'introduced' instead of introcuded

proposals/XXXX-extend-property-wrappers-to-functions-and-closures.md
filip-sakel: Oops, I need to freshen up on the original proposal. Though, I still feel that they're not that good of an idea.

proposals/XXXX-extend-property-wrappers-to-functions-and-closures.md
filip-sakel: Okey, I've changed that. The only transformation that still has an overload is the `inout` one. Let me know how the version discussed in the proposal seems to you. Otherwise, we can remove any `inout` logic from function parameters and just mention `inout` in a Future Directions section.

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
```diff
@@ -0,0 +1,385 @@
...
+```swift
+@propertyWrapper
+struct Wrapper<T> {
+  var wrappedValue: T
+  var projectedValue: ProjectionType
+
+  init(projectedValue: ProjectionType) { ... }
+}
+
+func fn(@Wrapper value: Int) { ... }
```
filip-sakel, line 372: Would you mind if we changed the function name to something like `foo` instead of the abbreviation `fn`?

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
```diff
@@ -0,0 +1,385 @@
...
+
+## Future Directions
+
+### Support property wrapper initialization from a projected value
+
+Today, a property wrapper can be initialized from an instance of its `wrappedValue` type if the wrapper provides a suitable `init(wrappedValue:)`. The same initialization strategy is used in this proposal for property wrapper parameters to allow users to pass a wrapped value as a property wrapper argument. We could extend this model to support initializing a property wrapper from an instance of its `projectedValue` type by allowing property wrappers to define an `init(projectedValue:)` that follows the same rules as `init(wrappedValue:)`. This could allow users to additionally pass a projected value as a property wrapper argument, like so:
+
+```swift
+@propertyWrapper
+struct Wrapper<T> {
```
filip-sakel, line 365: I think we should rename the generic parameter `T` to `Value` for better clarity.

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
```diff
@@ -0,0 +1,385 @@
...
+  var wrappedValue: T
+  var projectedValue: ProjectionType
```
filip-sakel, line 367: Should we maybe specify the projection type, like this:

```swift
@propertyWrapper
struct Wrapper<Value> {

  typealias ProjectionType = ...

  var projectedValue: ProjectionType 


  init(projectedValue: ProjectionType) { ... }

  
  var wrappedValue: Value

}
```

I think we should include the type alias so that the reader doesn't have to search in the proposal to find if we refer to something specified previously. Instead, they'll see the type alias and understand that the actual type is irrelevant without being uncertain of whether it's been specified. 

> **NOTE**: The other changes to layout are just personal choice; feel free to leave them out.

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
hborla: I'm not against changing the example, but I actually am against using `foo` - instead I think we should use a more realistic example throughout the whole document, such as `@Lazy` or `@Clamped`. It would also be nice to use the same example throughout and build upon it to avoid too much context switching for the reader

The reason I'm against using `foo` and `bar` is that "foobar" is derived from some very negative slang.

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
hborla: You're totally right that this example isn't clear. Do you think using the same example throughout the proposal and building on it would clear up the confusion? We could still include a repeated definition of whichever wrapper type we choose so that the reader doesn't have to jump back and forth between different parts of the proposal. Also, using a more realistic example would better illustrate what these future directions enable in terms of expressivity instead of just showing the mechanics of how it would work

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
filip-sakel: #### [ ... ] realistic example throughout the whole document, such as @Lazy or @Clamped

I agree, I'll use the relevant names when rewriting the examples.

#### The reason I'm against using foo and bar is that "foobar" is derived from some very negative slang.

I had no idea that this was where foo originated from. Thanks for pointing it out!

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
filip-sakel: Using real-world wrappers (but maybe with simplified implementations) seems like a really good idea and I like how exploring future directions through such wrappers emphasizes their usefulness. As you can imagine, though, we won't be able to use just one wrapper in all the sections. However, I think we could stick to `@Clamped` for the Motivation and whatever other section qualifies. One such section, for example, is Detailed Design where we could use `@Clamped` instead of `@Wrapper` is some examples and of course it some Future Direction sub-sections would also apply - as you mentioned.

berrylawchingman (approved):

berrylawchingman (approved):

berrylawchingman (approved):

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
filip-sakel: I've changed all the sections up to and including the Proposed Solution. Arguably it's a bit more confusing than using meaningless wrappers; however, the do make the motivation more persuasive. Let me know if there's something you think we could simplify!

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
filip-sakel: I've change the example in the Future Direction section you wrote to this:

---

```swift
@propertyWrapper
struct Clamped<Value: Comparable> {
  ...

  init(projectedValue: Self) { ... }
}

func distanceFromUpperBound(
  @Clamped clamped: Int
) { ... }

distanceFromUpperBound(
  $clamped: Clamped(to: 0 ... 100, wrappedValue: 30)
) // returns: 70
```

---

It's something that one could actually see in the real world. Anyway, let me know if you like this better or if you prefer that we revert to the previous one.

proposals/XXXX-extend-property-wrappers-to-functions-and-closures.md
hborla: ✅

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
```diff
@@ -0,0 +1,388 @@
...
+2. The internal parameter name will be prefixed with an underscore, and the type of this parameter is the backing property wrapper type.
+3. A local computed property representing  `wrappedValue` will be synthesized by the compiler and named per the original (non-prefixed) parameter name. The accessors will mirror the `wrappedValue` accessors. A setter will only be synthesized for the local property if the `wrappedValue` setter is `nonmutating`, or if the wrapper is a reference type.
+4. If the property wrapper defines a `projectedValue`, a local computed property representing  `projectedValue` will be synthesized by the compiler and named per the original parameter name prefixed with a dollar sign (`$`). The same accessor rules for `wrappedValue` apply to `projectedValue`.
+5. When passing an argument to a property wrapper parameter, the compiler will wrap the argument in the appropriate `init(wrappedValue:)` call.
+
+#### Transformation Example:
+
+```swift
+@propertyWrapper
+struct Percentage {
```
hborla, line 178: I'm not attached to this example - I wanted something relatively simple that didn't need additional arguments in the wrapper attribute (which ruled out `Clamped`). @filip-sakel please feel free to change this if you think of something better 🙂

filip-sakel: How would you feel about a `@Prefixed` wrapper:

```swift
@propertyWrapper
struct Prefixed {

  var prefixValue: String
  var originalValue: String

 
  init(
    wrappedValue originalValue: String,
    by prefixValue: String
  ) { ... }

   
  var wrappedValue: String { 
    get { ... }
  }

  var projectedValue: Self { ... }

}
```

So that we can write:

```swift
func log(
  @Prefixed(by: "Error: ") error: String
) { ... }


log(error: "Couldn't fetch images.")
// prints: Error: Couldn't fetch images.

log(_error: Prefixed(by: "UIError: ", wrappedValue: "Couldn't fetch images."))
// prints: UIError: Couldn't fetch images.
```

I'm not sure if that's any better. If `@Prefixed` isn't good enough we could explore something error-related. Anyway, let me know what you think of `@Prefixed`.

hborla: `@Prefixed` still requires additional arguments in the wrapper attribute (the `(by: "Error")` part), and I'm leaning toward not allowing arguments in the wrapper attribute initially with this proposal

filip-sakel: Right, I just started writing the wrapper and completely forgot the single-argument part 😅. I'll give this some more thought to see if I can come up with something better than `@Percentage`.

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
```diff
@@ -0,0 +1,473 @@
...
+  (value: property, id: $property.keyPath.hashValue)
+}
+```
+
+It's important to note that allowing use of such a feature in function parameters would entail some limitations. For example, a parameter makred with a wrapper type referencing enclosing `Self` would not be initializable with the sugared function call that utilizes `init(wrappedValue: ...)`. That's because it would require access to the enclosing `Self`; however, in the case of calling a function this would be undefined.
+
+
+### Add Wrapper Types in the Standard Library
+
+The has been some discussion of adding wrapper types to the Standard Library, such as an `@Atomic` wrapper, so as to facilitate certain APIs. For example, `@UnsafePointer` could be a very useful one, due to the way one accesses the `pointee` property of an `UnsafePointer` type:
```
filip-sakel, line 446: Do you have any other ideas about wrappers we could use here? I'll look into the possibility of adding a `@Lazy` wrapper here, but other than that I can only think of `@Atomic` and `@ThreadSpecific` – which I'm not certain should be included in the Standard Library.

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
```diff
@@ -0,0 +1,473 @@
...
+  @State 
+  private var shoppingItems: [Item]
+
+
+  var body: some View {
+    ForEach($shoppingItems) { @Binding shoppingItem in
+    
+      Text(shoppingItem.name)
+        .onTapGesture {
+          shoppingItem.quanity += 1
```
hborla, line 138: Can we add a `TextField` or something else that requires a binding via `$shoppingItem` to this example to show that using an `inout Item` parameter would not really solve the issue?

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
hborla: Hmm, not off the top of my head but I'll keep thinking about it

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
```diff
@@ -0,0 +1,456 @@
...
+We propose to extend the contexts were application of property-wrapper types is permitted. Namely, application of such types will be allowed on function and closure parameters:
+
+
+## Detailed design
+
+Property wrappers are essentially sugar wrapping a given property with compiler synthesized code. This proposal retains this principle employing the following rules for transformation.
+
+### Property Wrappers on Function Parameters
+
+Function parameters marked with compatible property-wrapper custom attributes must conform to the following rules:
```
hborla, line 148: @filip-sakel I noticed that you changed some of the language in these rules to be more consistent (thank you!), but I think some details got lost along the way. I wrote "a **set** of property wrapper custom attributes" to make it clear that these rules are generalized for property wrapper composition, e.g.

```swift
func test(@Wrapper @Wrapper @Wrapper param: Int) { ... }
``` 

Similarly, in the first rule below, it's important to say that **each** property wrapper type must provide a suitable `init(wrappedValue:)`.

filip-sakel: Does this look better?

---

Function parameters marked with a set of compatible property-wrapper custom attributes must conform to the following rules:

1. For initalization through `wrappedValue`'s type, all property-wrapper types must provide a suitable `init(wrappedValue:)`.
2. Each `wrappedValue` getter shall be `nonmutating`.
3. Default values for such parameters must be expressed in terms of the innermost `wrappedValue` type.

hborla: Yes, thank you!

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
```diff
@@ -0,0 +1,513 @@
...
+  $value.withMemoryRebound(to: UInt64.self {
+    ... 
+  }
+  
+}
+```
+
+As a result, unsafe code is not dominated by visually displeasing accesses to `pointee` members; rather, more natural and clear code is enabled. 
+
+What's more, a `@Lazy` type could be added so as to alleviate the need for custom behavior built into the compiler. Instead, `@Lazy` would act as more comprehensible, easy to maintain type, that would also allow for a more streamlined way of resetting its storage:
```
hborla, line 457: I think if we are going to include this, we should use an example that uses `@Lazy` on a parameter. Otherwise, it seems strange that this is included in the "Property Wrapper Parameters" proposal

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
```diff
@@ -0,0 +1,513 @@
...
+
+
+reportProgress(of: Percentage(wrappedValue: 50))
+```
+The changing type of the `@autoclosure` is incredibly misleading, as it's not obvious that `@autoclosure` applies to the backing wrapper rather than the wrapped value. Therefore, using `@autoclosure` for a property wrapper function parameter will be a compiler error.
+
+
+### Property Wrappers on Closure Parameters
+
+Since a property wrapper custom attribute is applied directly to the declaration that will be wrapped, application of a property wrapper type is only available within a closure expression. That is, the signature of a function that contains a closure cannot include the property wrapper attibute. Instead the application of the attrbute will be up to the caller of the function, which supplies the closure argument.
```
benrimmington, line 236:
```suggestion
Since a property wrapper custom attribute is applied directly to the declaration that will be wrapped, application of a property wrapper type is only available within a closure expression. That is, the signature of a function that contains a closure cannot include the property wrapper attribute. Instead the application of the attribute will be up to the caller of the function, which supplies the closure argument.
```

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
benrimmington: A simple example might be a property wrapper for when you always want to be given a lowercased string.

```swift
func example(@Lowercased _ urlScheme: String) -> Void { ... }
```

Would the `Lowercased` initializer be allowed to have a `with locale: Locale? = nil` default argument?

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
```diff
@@ -0,0 +1,513 @@
...
+    // Great!
+  }
+
+  mutating func adding(_ offset: Int) {
+    percent += min(100, max(0, offset))
+    //         ^~~~~~~~~~~~~~~~~~~~~~~~
+    // We are forced to manually adjust 'percent' 
+    // instead of utilizing the abstraction 
+    // property wrappers offer.
+  }
```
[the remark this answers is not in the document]
hborla: The existing example is intentional. The behavior of `percent += offset` and `percent += min(100, max(0, offset))`  is different if you pass a negative offset. Although the example is contrived (and there's no real reason to adjust the value if it's above the upper bound), it's meant to remind readers of assertions or adjustments they write on parameter values that would be useful to 1) have an abstraction for, and 2) be able to communicate to callers through the API itself.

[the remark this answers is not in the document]
benrimmington: > The existing example is intentional.

@hborla Thanks, I've updated the suggested change.

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
hborla: @benrimmington these are great questions. Could you please ask them on the pitch thread?

hborla:
@benrimmington thank you for the review and suggested changes! @filip-sakel when you get a chance, would you mind incorporating the changes? I'm trying to get the implementation PR up 🙂

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
```diff
@@ -0,0 +1,513 @@
...
+
+* Proposal: [SE-NNNN](NNNN-extend-property-wrappers-to-function-and-closure-parameters.md)
+* Authors: [Holly Borla](https://github.com/hborla), [Filip Sakel](https://github.com/filip-sakel)
+* Review Manager: TBD
+* Status: **Awaiting implementation**
+
+
+## Introduction
+
+Property Wrappers were [introduced in Swift 5.1](https://github.com/apple/swift-evolution/blob/master/proposals/0258-property-wrappers.md), and have since become a popular feature abstracting away common accessor patterns for properties. Currently, applying a property wrapper is solely permitted on properties inside of a type context. However, with increasing adoption demand for extending _where_ property wrappers can be applied has emerged. This proposal aims to extend property wrappers to function and closure parameters.
```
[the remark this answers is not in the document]
filip-sakel: Is this better? 

---

Property **w**rappers were [introduced in Swift 5.1](https://github.com/apple/swift-evolution/blob/main/proposals/0258-property-wrappers.md), and have since become a popular feature abstracting away common accessor patterns for properties. Currently, applying a property wrapper is solely permitted **at local and** type context. However, with increasing adoption demand for extending _where_ property wrappers can be applied has emerged. This proposal aims to extend property wrappers to function and closure parameters.

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
```diff
@@ -0,0 +1,513 @@
...
+
+
+## Motivation
+
+Property wrappers have undoubtably been very successful. Applying a property wrapper to a property is enabled by an incredibly lightweight and expressive syntax. Therefore, library authors can expose complex behavior through easily understandable property-wrapper types in an efficient manner. For instance, frameworks such as [SwiftUI](https://developer.apple.com/documentation/swiftui/) and [Combine](https://developer.apple.com/documentation/combine) introduce property wrappers such as [`State`](https://developer.apple.com/documentation/swiftui/state), [`Binding`](https://developer.apple.com/documentation/swiftui/binding) and [`Published`](https://developer.apple.com/documentation/combine/published) respectively to expose elaborate behavior through a succint interface, helping craft expressive yet simple APIs. However, property wrappers are only applicable to type properties, shattering the illusion that they helped realize in the first place:
```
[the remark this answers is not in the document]
filip-sakel: Updated:

---

Property wrappers have undoubtably been very successful. Applying a property wrapper to a property is enabled by an incredibly lightweight and expressive syntax. Therefore, library authors can expose complex behavior through easily understandable property-wrapper types in an efficient manner. For instance, frameworks such as [SwiftUI](https://developer.apple.com/documentation/swiftui/) and [Combine](https://developer.apple.com/documentation/combine) introduce property wrappers such as [`State`](https://developer.apple.com/documentation/swiftui/state), [`Binding`](https://developer.apple.com/documentation/swiftui/binding) and [`Published`](https://developer.apple.com/documentation/combine/published) respectively, to expose elaborate behavior through a succin**c**t interface, helping craft expressive yet simple APIs. However, property wrappers are only applicable to **local and** type properties, shattering the illusion that they helped realize in the first place:

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
filip-sakel: @hborla what do you think of `@Lowercased`?

In my opinion, we can be more creative with `@Lowercased` as the current example is supposed to report some progress, whereas we could use lowercase for something like a username or the 'urlScheme' example that @benrimmington proposed.

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
filip-sakel: I've given this some thought and I wanted to get your opinion. 

You proposed adding `@Lazy` as a replacement for `lazy` and `@autoclosure`. That's an amazing idea, but really difficult to implement if not impossible in the current language semantics. It's difficult to implement `@Lazy` as a replacement for `@autoclosure`, as it requires an immutable getter – which `@Lazy` lacks by design. There are also various other hurtles that `@Lazy` would have to overcome, which would make justifying the implementation effort difficult. 

Therefore, that leaves us with `lazy`, which I used in the examples. I do think, though, that you're right; we should end the proposal with an example that utilizes the feature we actually propose. Thus, I think we should look for other wrapper types that fit the criteria of being acceptable for the standard library and that utilize property wrappers in function or closure parameters.

Thoughts?

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
hborla: I like that example!

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
hborla: You're right that if we want to use `@Lazy` on a parameter, it can't have a `mutating` getter. If we want to use this example, we'd have to implement `Lazy` differently than in the original property wrapper proposal. Using reference types is the way to do this, either by making `Lazy` itself a reference type, or making it store an instance of some reference type that represents the initialization state.

filip-sakel: Well, yeah there are various ways of achieving reference-semantics; what I’m skeptical about  is whether it’s worth going through a lot of trouble for a simple ‘@Lazy’ wrapper instead of finding a better wrapper for the job. Another wrapper type would avoid complicated discussion about what will just be a future direction. Namely, I expect review questions asking about aspects like the following:

1. Lazy properties currently exhibit value semantics; how will a single reference-semantics ‘@Lazy’ wrapper handle that?

2. Won’t ‘@autoclosure’ still be required?

3. Why spend so much resources on syntactic sugar?

hborla: That's fair - it's not worth raising a bunch of questions just for a future direction. I think we should find another example or just remove the section. Somebody on the pitch thread had suggested creating a "manifesto" document for all of the future directions for property wrappers that have been floating around, and we could put all the work you've done for this section in there so it doesn't get lost!

filip-sakel: Yeah, I read that post and I think I this is a _great_ idea! There’s so much potential with property wrappers that we just can’t fit into a future-directions section. As for the wrapper to be used here, any suggestion is welcome; I’ll also be searching for a good wrapper type in personal and public projects.

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
```diff
@@ -0,0 +1,521 @@
...
+    }
+  }
+  
+}
+```
+
+
+## Proposed solution
+
+We propose to extend the contexts where application of property-wrapper types is permitted. Namely, application of such types will be allowed on function and closure parameters:
```
filip-sakel, line 140: Is there any reason why we removed the example that was here or was it removed by accident?

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
```diff
@@ -0,0 +1,521 @@
...
+  
+}
+
+
+postUrl(urlString: Lowercased(wrappedValue: "mySite.xyz/myUnformattedUsErNAme"))
+```
+
+#### Restrictions on property wrapper function parameters
+
+##### `@autoclosure`
```
filip-sakel, line 211: This part seems to belong to __Alternatives considered__; _not_ __Detailed design__. I propose we simple say that `@autoclosure` is not supported, maybe note that this is discussed in more detailed at the  __Alternatives considered__ section, and place it in that section.

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
```diff
@@ -0,0 +1,521 @@
...
+The above code would be transformed to:
+```swift
+func postUrl(
+  urlString _urlString: @autoclosure () -> Lowercased
+) { ... }
+
+
+postUrl(urlString: Lowercased(wrappedValue: "mySite.xyz/myUnformattedUsErNAme"))
+```
+As you can see, the changing type of the `@autoclosure` is incredibly misleading, as it's not obvious that `@autoclosure` applies to the backing wrapper rather than the wrapped value. For these reasons, using `@autoclosure` for a property-wrapper function-parameter will be invalid.
```
filip-sakel, line 232: @hborla, could you clarify what you mean by "the changing type of the `@autoclosure`". Do you mean the type that results from adding the `@autoclosure` attribute, the action of adding the attribute, or something else entirely?

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
```diff
@@ -0,0 +1,521 @@
...
+  ...
+}
+```
+
+This approach enables inference of the wrapper attribute on the closure parameter from context. However, this breaks the property-wrapper declaration-model, and would force callers to use the property-wrapper syntax. This approach, also, raises questions about anonymous closure-parameters that have an inferred property-wrapper custom-attribute. That is, if an anonymous closure parameter `$0` has the `wrappedValue` type, accessing the backing wrapper and projected value would naturally use `_$0` and `$$0`, which are _far_ from readable. Furthermore, suppose `$0` is bound to the backing wrapper-type; this would mean that naming the parameter would cause the value to change types, which would be very unexpected from a user standpoint. All in all, the property-wrapper syntax is purely an implementation detail for the closure body, which does _not_ belong to the API signature.
+
+
+## Future directions
+
+### Support property-wrapper initialization from a projected value
```
hborla, line 352: The hyphenation of property wrapper is inconsistent throughout the proposal and it's a little distracting. I think we should not hyphenate it at all - it isn't hyphenated anywhere in the original proposal or in TSPL, even when using it as an adjective, e.g. "property wrapper type". For example, in the [original proposal](https://github.com/apple/swift-evolution/blob/master/proposals/0258-property-wrappers.md#proposed-solution):

> This implements the property `foo` in a way described by the _property wrapper type_ for Lazy:

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
hborla: I removed it because it was really similar to the example above with `(@Binding shoppingItem)` but it was less compelling, so I didn't think it added anything. One thing we could do instead is move the `Binding` example here and explain it a little more, and just leave the "before" examples in the motivation

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
hborla: This is a justification for why `@autoclosure` is not supported. It's not really an alternative. We could explicitly label this as **Justification** or **Rationale**; I saw that in one of the [concurrency proposals](https://github.com/DougGregor/swift-evolution/blob/async-await/proposals/nnnn-async-await.md#asynchronous-functions) and I liked that the justification for a design decision was made explicit.

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
```diff
@@ -0,0 +1,521 @@
...
+  @Clamped(to: 0 ... 100)
+  var percent = 0
+     
+     
+  mutating func increment() {
+    percent += 1
+    // Great!
+  }
+
+  mutating func adding(_ offset: Int) {
```
hborla, line 48: I'm thinking we should remove this part of the example. I think it's caused a lot of people on the pitch thread to incorrectly develop a mental model for property wrapper parameters by comparing it to passing a wrapped value type. In reality, you should think of a property wrapper parameter as passing the backing wrapper type, while providing a convenient way to initialize that backing wrapper at the call-site.

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
```diff
@@ -0,0 +1,537 @@
...
+}
+```
+
+However, this can take flexibility away from the call-site if the property wrapper has other `init` overloads, because the call-site can cannot choose a different initializer. Further, if the property wrapper is default initialized in the struct, then the memberwise initializer will choose the backing wrapper type, even if the wrapper supports `init(wrappedValue:)`. This creates unnecessary boilerplate at call-sites that do want to use `init(wrappedValue:)`:
+
+```swift
+import SwiftUI
+
+struct TextEditor {
+  @State() var document: Optional<URL>
```
filip-sakel, line 43: Are the parenthesis right after `@State` intentional? If yes, I'm kind of lost as to why; if there _is_ a reason, maybe we could add a comment right below it to explain why the paranthesis are there.

hborla: Yes, empty parenthesis after the wrapper attribute is the way to do default initialization of the backing wrapper. I thought this was clear from the sentence above:

> Further, if the property wrapper is default initialized in the struct ...

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
filip-sakel: Yeah, I guess it could help clarify the mental. Not to mention that`min(100, max(0, offset))`  may confuse some people. That is, this code is essentially a clamping algorithm – which is what a proposal to add `clamping(to:)` to the standard library uses. What I mean by all this, is that people who have not used the above code to clamp a value, will most likely not understand it and, probably, get slightly confused by it.

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
filip-sakel: Oh, okey. I saw the colon right after "parameters" and I thought that I might have deleted that by accident. As far as the moving of the "after" examples to the Proposed Solution section, I think that's a really good idea. There's, also, not a lot of space between the "before" examples and the Proposed Solution section, so I think it's going to work out really well!

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
filip-sakel: I think this _really_ looks like an Alternatives Considered topic; __however__, if you think it should be included in the Detailed Design section as a sort of Rationale, I think that's okey. We do, though, – at least in my opinion – have to make it slightly smaller and in more of a note style. So, for example, if we shortened it keeping the most important parts, and put it in a note style as is done in the concurrency proposal, think that would be better.

#### Just to see how it would look like 


>*__RATIONALE:__* Function parameters with a property-wrapper custom-attribute cannot have an `@autoclosure` type, because `@autoclosure` is _unnecessary_ for the wrapped value itself. That is, the wrapped-value argument at the call-site will _always_ be wrapped in a call to `init(wrappedValue:)`, which can already support `@autoclosure` arguments. Furthermore, using `@autoclosure` for the backing wrapper-type would be misleading, since the type spelled out for the parameter would not be the true autoclosure type; consider the `reportProgress` example from above with an `@autoclosure` parameter:
>
 > ```swift
> func postUrl(
>  @Lowercased urlString: @autoclosure () -> String
> ) { ... }
>
>
> postUrl(urlString: "mySite.xyz/myUnformattedUsErNAme")
> ```
 > The above code would be transformed to:
 > ```swift
> func postUrl(
>   urlString _urlString: @autoclosure () -> Lowercased
> ) { ... }
>
>
> postUrl(urlString: Lowercased(wrappedValue: "mySite.xyz/myUnformattedUsErNAme"))
> ```
 > As you can see, the changing type of the `@autoclosure` is incredibly misleading, as it's not obvious that `@autoclosure` applies to the backing wrapper rather than the wrapped value. For these reasons, using `@autoclosure` for a property-wrapper function-parameter will be invalid.

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
filip-sakel: I think we should use [this Grammarly article](https://www.grammarly.com/blog/hyphen/) as a guide. Maybe in the following, a hyphen isn't required in the "custom-attributes" part:

```
property-wrapper custom-attributes
                       ^ This might not be necessary
```

However, when using "property wrappers" as an adjective, it _is_ required. For example, saying a "to do list" is kind of confusing, but definitely ungrammatical. Likewise, saying a "property wrapper attribute" is also ungrammatical. It would be valid if "property" and "wrapper" were both used as adjectives and if the former modified the latter. However, that's not the case; it's like they go together. Thus the correct usage of "property wrappers" when used as an adjective is with a hyphen, such as in "property-wrapper type/attribute".

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
hborla: Yeah, absolutely. I think I'm going to change this example to use the `@Lowercased` property wrapper instead since we use that example later anyway, and I'll fix the rest in the process. Thanks!

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
hborla: Totally agree that the rationale could be smaller. How about we explain the first part about using `@autoclosure` for the wrapped value type here:

Function parameters with a property-wrapper custom attribute cannot have an @autoclosure type.

> RATIONALE: `@autoclosure` is unnecessary for the wrapped value itself. That is, the wrapped-value argument at the call-site will always be wrapped in a call to `init(wrappedValue:)`, which can already support `@autoclosure` arguments.

And the rest, which is really talking about supporting `@autoclosure` for the backing wrapper, can go in Alternatives Considered. What do you think?

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
hborla: Yeah, I understand that it's grammatically correct to hyphenate when "property-wrapper" is used as an adjective. Let's at least remove the hyphenation on nouns, e.g. the phrases "backing property-wrapper" without "type" after, "function-parameter," and "custom-attribute." I'll also fix the hyphenation in the section I just added about memberwise initialization!

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
filip-sakel: Maybe we could change it to: 

> Further, if the property wrapper is default initialized in the struct – with an `init()` overload – ...

I say that because "default initialized" may not instinctively lead the reader to assume that we mean the plain `State` initializer: `init()`; not to mention, that this isn't the form we usually see `State` initialized in the documentation.

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
filip-sakel: Yeah, right 😅; I might have overdone it with the nouns, in which (I think) hyphens are not necessary. So, I agree we'd better leave "property wrappers" without any hyphens, when used as a noun. Likewise, we can change the examples you mentioned above. I'll also check for hyphenation on adjectives, so that we catch any errors.

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
```diff
@@ -0,0 +1,408 @@
...
+
+
+## Future directions
+
+### Support property-wrapper initialization from a projected value
+
+Today, a property wrapper can be initialized from an instance of its `wrappedValue` type if the wrapper provides a suitable `init(wrappedValue:)`. The same initialization strategy is used in this proposal for property wrapper parameters to allow users to pass a wrapped value as a property wrapper argument. We could extend this model to support initializing a property wrapper from an instance of its `projectedValue` type by allowing property wrappers to define an `init(projectedValue:)` that follows the same rules as `init(wrappedValue:)`. This could allow users to additionally pass a projected value as a property wrapper argument, like so:
```
filip-sakel, line 351: In the text there's this phrase: 
> "[...] wrapper provides a suitable `init(wrappedValue:)`"

Do you think it would be better if we changed it to: 
>  "[...] wrapper provides a suitable initializer: `init(wrappedValue:)`"

Otherwise, if we mean something else here, maybe we should clarify that.

filip-sakel (commented):
Overall, I really like the updated restrictions part.

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
```diff
@@ -0,0 +1,424 @@
...
+
+### Memberwise initialization
+
+Currently, property-wrapped struct properties interact with function parameters, through the struct's synthesized memberwise initializer. However, property-wrapper attributes are _not_ supported on function parameters. This leads to really complicated and nuanced rules for which type, between the wrapped type and the backing property-wrapper type, the memberwise initializer accepts. Namely, the compiler will choose the wrapped-value type to offer a convenience to the call-site, when the property wrapper supports initalization through the suitable initializer: `init(wrappedValue:)`, as seen here:
```
filip-sakel, line 21: > "complicated and nuanced rules for which type the memberwise initializer accepts"

After trying to wrap my head around what you mean by this, I think I figured it out. So, I changed this paragraph in order to simplify it as much as I could. Let me know if you think I missed something; also, any changes that simplify this portion are definitely welcome (especially because this is the first paragraph of Motivation).

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
```diff
@@ -0,0 +1,426 @@
...
```
filip-sakel, line 21: Here, and in some other places in the text, you use "call-site" hyphenated, yet in some other places it's unhyphenated. So, I'm asking you which one do you think is more correct, so that we can use one of the two throughout the proposal. (I assume it's "call-site", since you used it in Memberwise Initializers, which you wrote more recently; but I wanted to ask you just in case)

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
filip-sakel: I think it turned out very good! As for any further explanation that, you mentioned, could be moved in Alternatives Considered, I don't think it's necessary; what's already in Detailed Design is already sufficient – in my opinion.

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
hborla: I've seen them used interchangeably and I think I'm indifferent; both seem correct to me. Let's use "call-site" since that's apparently what I wrote everywhere else haha (I think it was out of habit, rather than an intentional spelling)

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
hborla: Ah sorry, changed "suitable `init(wrappedValue:)` in Detailed Design to be a little more descriptive, I'll just use the same replacement here

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
```diff
@@ -132,7 +132,7 @@ func postUrl(@Lowercased urlString: String) {
 postUrl(urlString: "mySite.xyz/myUnformattedUsErNAme")
 ```
 
-What's more, if `Binding` conditionally conformed to `RandomAccessCollection`, property-wrapper parameters could be used with `ForEach` to access collection elements directly in the `content` closure and enable property-wrapper syntax in the closure body:
+What's more, if `Binding` conditionally conformed to `RandomAccessCollection`, property-wrapper parameters could be used with `ForEach` to access collection elements directly in the `content` closure; thus enabling property-wrapper syntax in the closure body:
```
hborla, line 135: "thus enabling" doesn't make sense here; "accessing collection elements directly" and "enabling property-wrapper syntax" are two separate benefits. The first one does not imply the second one.

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
hborla: Clarifying that the compiler is choosing between the wrapped type and the backing wrapper type is great, thank you! 

We're trying to eliminate the phrase "suitable initializer" or "suitable `init(wrappedValue:)`", though, so I'll replace that part (sorry if it was me who originally wrote that 🙂 ).

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
hborla: Whoops, I was mixing up my terminology. This is explicit initialization via `init()`. Fixed in the description above the example.

proposals/XXXX-extend-property-wrappers-to-function-and-closure-parameters.md
filip-sakel: It's okey, don't worry about that. I think it is much better know, anyway!

filip-sakel:
I think we're ready for review; feel free to mark this PR as such.

hborla:
@filip-sakel Agreed, and will do! I'm fixing one last bug (I think) in the implementation and then I will un-draft this PR

lattner:
I'm taking this as review manager and will kick it off shortly.

PR #1183, opened 2020-09-13
Small grammatical fix

byaruhaf:


proposals/0289-function-builders.md
```diff
@@ -226,7 +226,7 @@ return body {
 
 The above has to be embedded into the ordinary language somehow, which means that at least the outermost layer must obey something like ordinary language rules.  Under ordinary language rules, this is a function call to `body `passing a trailing closure.  It makes sense, then, that what we're doing is taking the body of the anonymous function and apply some sort of transformation to it.  This raises a series of separate questions:
```
benrimmington, line 227: - `` `body ` `` → `` `body` ``
- `apply` → `applying`

```suggestion
The above has to be embedded into the ordinary language somehow, which means that at least the outermost layer must obey something like ordinary language rules.  Under ordinary language rules, this is a function call to `body` passing a trailing closure.  It makes sense, then, that what we're doing is taking the body of the anonymous function and applying some sort of transformation to it.  This raises a series of separate questions:
```

benrimmington:
Cc: @DougGregor

DougGregor:
Thanks!

PR #1186, opened 2020-09-16
[Proposal] Demo Code Compile Error in SE0258

Mas0nSun:
There is a compile error in [0258-property-wrappers](https://github.com/apple/swift-evolution/blob/master/proposals/0258-property-wrappers.md).
The line of 610 `wrappedValue = value.copy()`
Compiler Error: `Use of unresolved identifier 'value'`

Screen Shot 2020-09-16 at 4 48 37 PM

I don't know whether my modification is consistent with previous meaning, please review it thanks.


DougGregor (approved):
Thanks!

PR #1188, opened 2020-09-17
Fix the "Swift Next" filter on the status page

benrimmington:


benrimmington:
@krilnon This is the change you suggested in https://github.com/apple/swift-evolution/pull/1187#issuecomment-693700915, and I've tested it locally.

PR #1196, opened 2020-10-02
Fix double addressing

an0:


lorentey (approved):
Good catch!

jckarter:
Thanks @an0!

PR #1198, opened 2020-10-22
Update status and implementation of SE-0274

dabrahams:
Someone please double-check that my information is correct.

### DIFF
--- a/proposals/0274-magic-file.md
+++ b/proposals/0274-magic-file.md
@@ -4,8 +4,8 @@
 * Authors: [Brent Royal-Gordon](https://github.com/brentdax), [Dave DeLong](https://github.com/davedelong)
 * Review Manager: [Ben Cohen](https://github.com/airspeedswift/)
 * Original review: [Returned for revision](https://forums.swift.org/t/se-0274-concise-magic-file-names/32373/50) 
-* Status: **Accepted**
-* Implementation: Prototype in master behind `-Xfrontend -enable-experimental-concise-pound-file`; revisions in [apple/swift#29412](https://github.com/apple/swift/pull/29412)
+* Status: Implemented (Swift 5.3)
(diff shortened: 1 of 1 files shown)


proposals/0274-magic-file.md
```diff
@@ -4,8 +4,8 @@
 * Authors: [Brent Royal-Gordon](https://github.com/brentdax), [Dave DeLong](https://github.com/davedelong)
 * Review Manager: [Ben Cohen](https://github.com/airspeedswift/)
 * Original review: [Returned for revision](https://forums.swift.org/t/se-0274-concise-magic-file-names/32373/50) 
-* Status: **Accepted**
-* Implementation: Prototype in master behind `-Xfrontend -enable-experimental-concise-pound-file`; revisions in [apple/swift#29412](https://github.com/apple/swift/pull/29412)
+* Status: Implemented (Swift 5.3)
```
benrimmington, line 7: This field needs double asterisks, to avoid a "status not found" warning in the [proposals data](https://data.swift.org/swift-evolution/proposals).

```suggestion
* Status: **Implemented (Swift 5.3)**
```

I'm not sure of the actual status, because [SE-0285](https://github.com/apple/swift-evolution/blob/main/proposals/0285-ease-pound-file-transition.md) extended the transition until Swift 6.

Can the [*Implementation status*](https://github.com/apple/swift-evolution/blob/510866545390364f4faea652aca255ee44b14567/proposals/0274-magic-file.md#implementation-status) section in this proposal be removed?

brentdax: It's kind of an interesting case, actually:

* SE-0274 was implemented during the Swift 5.3 cycle, but...
* SE-0285, which changed it significantly, was also implemented in Swift 5.3, but...
* The SE-0285 implementation was actually reverted from *Apple* Swift 5.3 to avoid integration issues; it's in the Apple Swift 5.3.1 beta.

So, where does that leave SE-0274? The unmodified design shipped in Apple Swift 5.3, but not in any official open-source release. Was it superseded? Was it implemented?

PR #1207, opened 2020-11-18
Amendment to SE-0291 Package Collections

yim-lee:
Motivation:
Multiple members of the Swift community have inquired about the data
(JSON) format and questioned why it is left out of the proposal.

Modifications:
Adding a "Data format" section to the proposal to explain that the data
format is not considered stable API and thus is not included. Also
mentioning that tools will be provide to help generate and consume
package collection JSON.


tomerd (approved):
lgtm

PR #1210, opened 2020-11-21
[Proposal] Codable synthesis for enums with associated values

drexin:


proposals/NNNN-codable-synthesis-for-enums-with-associated-values
```diff
@@ -0,0 +1,274 @@
...
+```swift
+enum Command: Codable {
+  case load(key: String)
+
+  enum CodingKeys: String, CodingKey {
+    case load = "lade"
+  }
+
+  enum CodingKeys_load: String, CodingKey {
+    case key = "schluessel"
```
ktoso, line 240: 😄 fun example use-case

ktoso (approved):

proposals/NNNN-codable-synthesis-for-enums-with-associated-values
```diff
@@ -0,0 +1,274 @@
...
+}
+```
+
+The compiler will now only synthesize the code for the `load` and `store`
+cases. An attempt to en- or decode a `dumpToDisk` value will cause an error
+to be thrown.
+
+Customizing which values will be included follows the same rules as the
+existing functionality. Values that are excluded must have a default value
+defined if a `Decodable` conformance should be synthesized. If only `Enodable`
```
cristik, line 188:
```suggestion
defined if a `Decodable` conformance should be synthesized. If only `Encodable`
```

drexin: Thanks!

PR #1216, opened 2020-12-02
Enable syntax highlighting in SE-0294

MaxDesiatov:
This improves readability of Swift code given in the proposal.


benrimmington (approved):
- Should the `@available(_PackageDescription, introduced: 999.0)` be removed?
- Should the new API be wrapped in `extension Target { ... }` for context?
- There's some inconsistent indentation, but it matches the implementation.

PR #1223, opened 2020-12-05
Update 0296-async-await.md

haikusw:
seems likely "asynchronous" was intended here (?)


haikusw:
Or maybe it's the reverse (given the section title) and "even when they can block" is the part that's incorrect?

> This is hard to quantify, but the authors believe that the awkwardness of defining and using asynchronous APIs (using completion handlers) has led to many APIs being defined with apparently synchronous behavior, even when they [do not] block.

Maybe I'm just confused; if an "API is defined with apparently synchronous behavior" then shouldn't it block and not execute code after it until it has completed its work?

PR #1225, opened 2020-12-07
[SE-0296] Try to clarify potential suspension points

DougGregor:


JensAyton:
Thanks, I think this is a lot clearer. Sorry for being a pain about this!

PR #1230, opened 2020-12-09
Concurrency Interoperability with Objective-C

DougGregor:


proposals/NNNN-concurrency-objc.md
```diff
@@ -0,0 +1,341 @@
...
+  ```swift
+  @objc
+  func urlSession(
+      _ session: URLSession,
+      dataTask: URLSessionDataTask,
+      didReceive response: URLResponse
+  ) async -> URLSession.ResponseDisposition { ... }
+  ```
+
+  Implementing both requirements would produce an error (due to two Swift methods having the same selector), but under the normal Swift rules implementing only one of the requirements will also produce an error (because the other requirement is unsatisfied). Swift’s checking of protocol conformances will be extended to handle the case where multiple (imported) requirements have the same Objective-C selector: in that case, only one of them will be required to be implemented.
```
benrimmington, line 314: Should this paragraph be changed, to account for the `@optional` requirement?

And/or should a different example with `@required` be used?

PR #1235, opened 2020-12-15
[SE-0297] Clarify the index values for the swift_async[_error] attributes

DougGregor:


proposals/0297-concurrency-objc.md
```diff
@@ -246,13 +246,13 @@ The transformation of Objective-C completion-handler-based APIs to async Swift A
...
   * `__attribute__((swift_async(none)))`. Disables the translation to `async`.  
-  * `__attribute__((swift_async(not_swift_private, C)))`. Specifies that the method should be translated into an `async` method, using the parameter at index `C` as the completion handler parameter.
-  * `__attribute__((swift_async(swift_private, C)))`. Specifies that the method should be translated into an `async` method that is "Swift private" (only for use when wrapping), using the parameter at index `C` as the completion handler parameter.
+  * `__attribute__((swift_async(not_swift_private, C)))`. Specifies that the method should be translated into an `async` method, using the parameter at index `C` as the completion handler parameter. The first (non-`self`) parameter has index 1.
+  * `__attribute__((swift_async(swift_private, C)))`. Specifies that the method should be translated into an `async` method that is "Swift private" (only for use when wrapping), using the parameter at index `C` as the completion handler parameter. The first (non-`self`) parameter has index 1.
```
varungandhi-apple, line 250: Is there ever a reason for this to be applied to class methods instead of instance methods? If so, this will be incorrect because the first non-`self` parameter will have index 0 for class methods.

DougGregor: Class methods also have a `self` parameter; it's of type `Class`

varungandhi-apple (approved):

PR #1244, opened 2021-01-06
Introduce a dedicated proposal for `with*Continuation`

jckarter:


proposals/ABCD-continuation.md
```diff
@@ -0,0 +1,211 @@
...
+
+The operation must follow one of the following invariants:
+
+- Either the resume function must only be called *exactly-once* on each
+  execution path the operation may take (including any error handling paths),
+  or else
+- the resume function must be called exactly at the end of the operation
+  function's execution.
+
+`Unsafe*Continuation` being an unsafe interface, it is undefined behavior if
```
slavapestov, line 80: This sentence is a bit awkward, maybe try to rephase it?

proposals/ABCD-continuation.md
```diff
@@ -0,0 +1,233 @@
...
+
+Asynchronous Swift code needs to be able to work with existing synchronous
+code that uses techniques such as completion callbacks and delegate methods to
+respond to events. Asynchronous tasks can suspend themselves on
+**continuations** which synchronous code can then capture and invoke to
+resume the task in response to an event
+
+Swift-evolution thread:
+
+- [Structured concurrency](https://forums.swift.org/t/concurrency-structured-concurrency/41622)
```
benrimmington, line 18:
```suggestion
- [Pitch](https://forums.swift.org/t/concurrency-continuations-for-interfacing-async-tasks-with-synchronous-code/43619)
```

jckarter: Discussion started in the Structured Concurrency thread, so I think we should link both threads.

proposals/ABCD-continuation.md
```diff
@@ -229,5 +230,67 @@ this here, though:
...
   property.
 
-
 ### Don't expose `UnsafeContinuation`
+
+One could similarly make an argument that `UnsafeContinuation` shouldn't be
+exposed at all, since the `Checked` form can always be used instead. We think
+that being able to avoid the cost of checking when interacting with
+performance-sensitive APIs is valuable, once users have validated that their
+interfaces to those APIs are correct.
```
ktoso, line 239: 👍

proposals/ABCD-continuation.md
```diff
@@ -229,5 +230,67 @@ this here, though:
...
+  `Optional`.
+- By contrast, failing to resume a continuation with `UnsafeContinuation`
+  does not corrupt the task, beyond leaking the suspended task's resources;
+  the rest of the program can continue executing normally. Furthermore,
+  the only way we can currently detect and report such a leak is by using
+  a class `deinit` in its implementation. The precise moment at which such
+  a deinit would execute is not entirely predictable because of refcounting
+  variability from ARC optimization. If `deinit` were made to trap, whether that
+  trap is executed and when could vary with optimization level, which we
+  don't think would lead to a good experience.
```
ktoso, line 262: This convienced me that that'a fine default I guess -- I initially wanted to trap here too, but then again -- if I'm debugging such leak then I can write my own wrapper that does the ugly/unreliable deinit anyway.

proposals/ABCD-continuation.md
```diff
@@ -229,5 +230,67 @@ this here, though:
...
+    cancelation: { c.resume(nil) })
+}
+
+if let result = callbackResult {
+  process(result)
+} else {
+  cancel()
+}
+```
+
```
ktoso, line 296: You may want to add this example too, it showcases that we can respect outside/parent task cancellation within an API implemented using `withUnsafeContinuation` which some people on the forums were not sure if this is supported or not:

--- 

It is also possible for wrappers around callback based APIs to respect their parent/current tasks's cancellation, as follows:

```swift
func fetch(items: Int) async throws -> [Items] {
  let worker = ... 
  return try Task.withCancellationHandler(
    handler: { worker?.cancel() }
  ) { 
    return try await withUnsafeThrowingContinuation { c in 
      worker.work(
        onNext: { value in c.resume(returning: value) },
        onCancelled: { value in c.resume(throwing: CancellationError()) },
      )
    } 
  }
}
```

If tasks were allowed to have instances, which is under discussion in the structured concurrency proposal, it would also be possible to obtain the task in which the `fetch(items:)` function was invoked and call `isCanceled` on it whenever the insides of the withUnsafeThrowingContinuation would deem it worthwhile to do so.

ktoso (approved):
LGTM, thanks @jckarter !

One minor addition at the end if you'd like to provide another example, it's something that had people confused on the forums that might be useful to clarify.

proposals/ABCD-continuation.md
jckarter: Thanks! I'll add this example to the proposal text.

PR #1246, opened 2021-01-08
AsyncSequence

parkera:
Pitch thread: https://forums.swift.org/t/concurrency-asyncsequence/

### DIFF
--- a/proposals/NNNN-asyncsequence.md
+++ b/proposals/NNNN-asyncsequence.md
@@ -99,7 +99,7 @@ The standard library will define the following protocols:
 public protocol AsyncSequence {
   associatedtype AsyncIterator: AsyncIteratorProtocol where AsyncIterator.Element == Element
   associatedtype Element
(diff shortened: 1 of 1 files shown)


DougGregor:
This is in good shape for a review. I'll schedule and manage it.

proposals/NNNN-asyncsequence.md
```diff
@@ -0,0 +1,386 @@
...
+for try await line in myFile.lines() {
+  // Do something with each line
+}
+```
+
+The compiler will emit the equivalent of the following code:
+
+```swift
+var it = myFile.lines().makeAsyncIterator()
+while let value = try await it.next() {
```
stevapple, line 185: It seems too late but... Isn’t this line expected to be:

```swift
while let line = try await it.next() {
```

parkera: Thank you -- I'll fix this in an update.

PR #1251, opened 2021-01-15
Fix code example in proposal SE-0235

jeroennoten:


proposals/0235-add-result.md
```diff
@@ -43,7 +43,7 @@ This can make it quite difficult to elegantly consume the results of these APIs:
 
 ```swift
 URLSession.shared.dataTask(with: url) { (data, response, error) in
-    guard error != nil else { self.handleError(error!) }
+    guard error == nil else { self.handleError(error!) }
```
benrimmington, line 46: The `guard` statement would also need to `return`.

tkremenek:
I am going to merge this fix, and then add the missing `return`.

benrimmington:
@tkremenek I've just noticed that this isn't using the `main` base branch.

It would be great to eventually delete the `master` branch. Existing links to proposals will automatically [redirect](https://github.blog/changelog/2020-07-17-links-to-deleted-branches-now-redirect-to-the-default-branch/).

PR #1257, opened 2021-01-29
Propose the addition of custom executors to Swift concurrency

rjmccall:


proposals/0000-custom-executors.md
```diff
@@ -0,0 +1,464 @@
...
+  For example, if one actor frequently makes requests of another,
+  and the actors rarely benefit from running concurrently,
+  configuring them to use the same executor may decrease the
+  runtime costs of switching between them.
+
+  For another example, if an asynchronous function makes many
+  calls to the same actor without any intervening suspensions,
+  running the function explicitly on that actor's executor may
+  allow Swift to avoid a lot of switching overhead (or may even
+  be necessary to perform those calls "atomically").
```
weissi, line 69: Thanks @rjmccall ! If you want another example: Just as you describe below, many systems do have fixed-width thread pools that don't block as long as there is outstanding work to be done. If there's no outstanding work to be done, these threads do usually block (as you describe below). So without custom executors, implementing one of these systems (typically used for I/O) would also not be possible without prohibitive thread switching (because we're obviously can't block say a thread of the default executor in `kqueue`, `epoll_wait`, iouring, or `mach_msg`.

Or in other words: Any system that implements its custom I/O and also wants to run `async` code will need to use custom executors for that.

proposals/0000-custom-executors.md
```diff
@@ -0,0 +1,464 @@
...
+  /// must return true for the reference.
+  ///
+  /// Keeping the actor reference valid must be sufficient to keep
+  /// the executor reference valid.
+  var serialExecutor: UnownedExecutorRef { get }
+}
+```
+
+An `actor` may derive its executor implementation in one
+of the following ways. We may add more ways in the future.
```
glbrntt, line 363: These seem a bit brittle to me: if you have a typo in `serialExecutor`, for example, presumably the actor would use the default serial executor and you would be none-the-wiser?

proposals/0000-custom-executors.md
rjmccall: To be clear, and I do remember that the proposal discusses this: while you certainly *could* use custom executors for that, I'm not sure it's a good idea, because having two fixed-width thread pools in a process basically means having one poorly-managed thread pool with way too many threads.  Abstractly, I think we really want to encourage people who are writing scales-to-the-number-of-cores thread pools to either use the default concurrent executor or replace it globally.  But if you think having multiple pools is important, I'd love to understand how you hope to use the second pool, and particularly how you think it should affect thread allocation in the main pool.

proposals/0000-custom-executors.md
rjmccall: I see your point, but I'm not sure what to do about it.  It's not reasonable to require actors to explicitly opt in to using the default serial executor, which means that there's always going to be some series of errors of omission that can lead to using it accidentally.

If we had a way to annotate that a function is supposed to be used for a protocol — a long-standing request — that would work for the `serialExecutor` case specifically.

proposals/0000-custom-executors.md
weissi: The default mode in SwiftNIO is exactly what you describe: One scales-to-the-number-of-cores thread pool that is shared by everything. However, we leave this up to the user because not every application is homogenous. 

For example, let's say your server is supposed to -- at peak times -- `accept` as many connections as possible. How you'd achieve that is to use one thread pool with number-of-cores threads which each bind on the same socket address (with `SO_RESUEPORT`). You could also give those threads elevated priority in the OS and possibly even pin them to an individual hardware CPU core. This will work well but is still quite latency sensitive because the OS's "listen queue" (queue of incoming connections that has yet to be `accept`ed) is bounded. So if at the peak time you can't call `accept` fast enough, the OS will drop connections for you (once the listen queue is exhausted). So far, this works really well with just one replaced default pool. But imagine, in the same program, we also run other code which is also supposed to run but isn't supposed to interfere with the accept-as-fast-as-possible work. Imagine for example, some JSON is supposed to be decoded or some something like that. It could be a pretty big if the JSON decoding would now cause us to return to the scheduler less often. This may cause us to drop connections. Priorities can't solve this issue because they are not preemptive.

As a really simple fix, users can configure a second (fixed-width) thread pool on which they run the other work. They are basically ring fencing the two workloads that are running in the same address space from each other by having separate pools.

Other examples include
- if you were to run two separate systems that perform I/O in one process. Let's say on Linux one system that multiplexes I/O using `io_uring` and another one that multiplexes using `epoll`. They can't share the same thread pool because their whole eventing and I/O mechanism is totally different. They could for example use two thread pools of size `number_of_cores/2` each. 
- in unit testing we usually start up and tear down the thread pool per test (suite) for them to be completely independent, work better when tests are run in parallel, and also to be able to test interesting scheduling scenarios with say 1 thread, or 3 threads, or some other "interesting" number
- say you have a service and an "admin interface". You may choose to have the admin interface run on a separate thread pool to be sure to never disrupt the main service from achieving its goals even if the admin interface does maybe do processing that's a little too heavy (say parse a JSON, protobuf, compress/decompress, ...) and may harm the latency of the main service

proposals/0000-custom-executors.md
```diff
@@ -0,0 +1,464 @@
...
+about the behavior of `execute(currentExecutor:)` and also adds
+several new methods which are used to allow the executor to opt in
+to supporting "switching" serial executors the same way that the
+default serial executor does.
+
+TODO: if we need to support blocking actors, we will need to extend
+this significantly; otherwise, actor executors will need to assume
+that returning from a job should always unblock the executor.
+(Presumably we do not want to block actor executors by blocking
+their thread.)
```
ktoso, line 328: do you here mean "blocking actor" as in "non-reentrant actor"?

I will read up how we execute actors today... in our other design we didn't need anything special from executors; the actor "has" a mailbox ("the lightweight queue") and submits _itself_ to an executor to run; if new tasks arrive at the actor while it is "'async' blocked" (waiting non-reentrantly on some thing), it does not submit itself to the executor, until that waiting thing comes back.

Though not sure if I understood your TODO right here, I can prototype this a bit in the coming week(s) if that'd help

proposals/0000-custom-executors.md
```diff
@@ -0,0 +1,464 @@
...
+
+  /// Given that canGiveUpThread() previously returned true, give up
+  /// the current thread.
+  func giveUpThread()
+
+  /// Attempt to start running a task on the current actor.  Returns
+  /// true if this succeeds.
+  func tryClaimThread() -> Bool
+}
+```
```
ktoso, line 344: Cool, I think this triple of functions along with executor identity and `SerialExecutor` can provide us everything we need to avoid hops for NIO 👍

proposals/0000-custom-executors.md
ktoso: I'm not too worried here to be honest. Same issue as with any other customization hook like that...

If you REALLY wanted to make sure to not make that mistake I think there's a number of ways to address it in plain user code.

a) @glbrntt I think you could make a require the executor type by a protocol:

```

protocol Executor {}
class EventLoop: Executor {}
protocol Actor {
  associatedtype ActorExecutor = Executor
  var ex: ActorExecutor { get }
}
protocol NeedsEL: Actor {
    associatedtype ActorExecutor = EventLoop
}

class Bad: NeedsEL {
  // error: repl.swift:10:7: error: type 'Specific' does not conform to protocol 'Actor'
  // class Specific: NeedsEL {
  //      ^
  //
  // repl.swift:5:7: note: protocol requires property 'ex' with type 'EventLoop'; do you want to add a stub?
  //  var ex: ActorExecutor { get }
}

class Good: NeedsEL {
  let ex: EventLoop
  init(executor eventLoop: EventLoop) {
    self.ex = eventLoop
  }
}
```

This should be possible right? Synthesis would have to be aware or it would fail since the default serial executor would not fulfil this requirement.

proposals/0000-custom-executors.md
```diff
@@ -0,0 +1,464 @@
...
+  ```
+
+  It will be `@inlinable` if `delegateActor` is.
+
+- Otherwise, the actor will use the default serial executor
+  implementation. Note that this implicitly adds storage to the
+  class.
+
+  `serialExecutor` will be synthesized as `public final`.
+  It will be `@inlinable` if the actor class is `frozen`.
```
ktoso, line 389: May be worth calling out that since it's a public let people can `thatActor.serialExecutor`? 

Since it's a let so it's actor independent 🤔

proposals/0000-custom-executors.md
```diff
@@ -0,0 +1,464 @@
...
+
+The following operations are provided in order to perform explicit
+scheduling onto executors:
+
+```swift
+extension Executor {
+  /// Run the given async operation explicitly on this executor.
+  func run<T>(operation: () -> async throws T) async rethrows -> T
+}
+```
```
ktoso, line 401: 👍

proposals/0000-custom-executors.md
```diff
@@ -0,0 +1,464 @@
...
+  `delegateActor` from the actor functions of the delegating actor.
+
+  The `serialExecutor` property will be synthesized as if
+  the following:
+
+  ```swift
+  public final var serialExecutor: UnownedExecutorRef {
+    return delegateActor.serialExecutor
+  }
+  ```
```
ktoso, line 380: I said this before but here again: this is great 👍 

We already have a number of other people looking at APIs that will benefit from this 👍 (Funnily enough, same shapes as we had in our things before), it'll definitely be very useful.

That one might get some naming bikeshed (can't stop thinking about it as a form of `friend` hm...)  but let's do that on forums 😉

proposals/0000-custom-executors.md
```diff
@@ -0,0 +1,464 @@
...
+```swift
+struct UnownedExecutorRef: Equatable {
+  init<T: SerialExecutor>(serialExecutor: T, supportsSwitching: Bool)
+  init<T: Executor>(concurrentExecutor: T)
+
+  static var defaultConcurrent: UnownedExecutorRef { get }
+
+  var isSerial: Bool { get }
+
+  func asOwned() -> Executor
```
ktoso, line 272: So the unowned is to avoid refcounting, and this vends out the "actual" one right?

proposals/0000-custom-executors.md
ktoso: Internally this makes an `UnownedJobRef` containing that closure, and causes an execute right?

Would this inherit priority?

Not a huge use case I guess, but just asking what you think this should do.

ktoso (approved):
This is looking good 👍 Some comments inline but overall looking great, seems this'll enable all we need 🤔

proposals/0000-custom-executors.md
glbrntt: > It's not reasonable to require actors to explicitly opt in to using the default serial executor, which means that there's always going to be some series of errors of omission that can lead to using it accidentally.

I agree. 

I'm also not worried about this, but seemed worth pointing out as a potential pitfall. Calling it out in documentation would probably be sufficient.

DougGregor:
@rjmccall can you pitch this over on the forums so the discussion can move there?

rjmccall:
> @rjmccall can you pitch this over on the forums so the discussion can move there?

Yeah, I'm just tweaking it now; I'll post it tonight.

proposals/0000-custom-executors.md
```diff
@@ -0,0 +1,464 @@
...
+executor. (For example, they may have their own fixed-width
+thread pool, and having two fixed-width pools in use at once
+completely defeats the purpose.) Programs should not attempt
+to do this by specifying a custom executor on absolutely
+everything in the program; that would be both highly invasive
+and doomed to failure. We think it should be possible to directly
+support hooking the default concurrent executor, perhaps by
+overriding a weak symbol, the same way that C programs can
+override `malloc`.  We won't discuss this further in this proposal
+because it doesn't otherwise impact the language and library design.
```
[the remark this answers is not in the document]
rjmccall: Right.

proposals/0000-custom-executors.md
rjmccall: It's an `async` function, so it would actually have to be done as part of a task and would pick up priority and so on from that.

proposals/0000-custom-executors.md
rjmccall: It's a way to convert back to the original executor in Swift code, yeah.

proposals/0000-custom-executors.md
rjmccall: I added some text about how this can be useful to use directly.

proposals/0000-custom-executors.md
rjmccall: I'm just saying that we'd definitely need extra requirements in `SerialExecutor` to support locking and unlocking custom actor executors — you need the jobs that the actor is running to be able to tell it to stop processing other jobs (and then somehow restart it), because there's an abstraction layer there.

ktoso (approved):
Looks great, thanks John :)

proposals/0000-custom-executors.md
ktoso: Ah right, sounds good 👍

proposals/0000-custom-executors.md
ktoso: Ok got it 👍

DougGregor:
This isn't ready for review yet. Please open up a new PR after it's been re-pitched and is ready for review

PR #1260, opened 2021-02-03
[SE-0301] Fix formatting and update to reflect SE-0294

owenv:
- fix formatting of the CLI argument descriptions
- remove the note about differences between library and executable targets now that SE-0294 has been accepted


owenv:
cc @tomerd

tomerd (approved):

owenv:
thanks!

PR #1268, opened 2021-02-15
Concurrent value and @concurrent functions

DougGregor:


proposals/XXXX-ConcurrentValueAndConcurrentClosures.md
```diff
@@ -0,0 +1,720 @@
...
+
+There are many useful reasons why you’d want to send bits of computation between concurrency domains in the form of a function - even trivial algorithms like `parallelMap` need this.  This occurs at larger scale as well -- for example, consider an actor example like this:
+
+
+```
+actor MyContactList { 
+  func filteredElements(_ fn: (ContactElement) -> Bool) async
+     -> [ContactElement] { … }
+}
+```
```
benrimmington, line 63: Could this example use [`AsyncSequence`](https://github.com/apple/swift-evolution/blob/main/proposals/0298-asyncsequence.md) and its `filter(_:)` method?

```swift
actor MyContactList: AsyncSequence { ... }
```

In which case, would the `@concurrent` attribute be added to `filter(_:)` in the protocol itself, or in each of its conformances?

proposals/XXXX-ConcurrentValueAndConcurrentClosures.md
```diff
@@ -0,0 +1,720 @@
...
+
+### Marker Protocols
+
+This proposal introduces the concept of a “marker” protocol, which indicates that the protocol has some semantic property but is entirely a compile-time notion that does not have any impact at runtime.  Marker protocols have the following restrictions:
+
+
+
+*   They cannot have requirements of any kind.
+*   They cannot inherit from non-marker protocols.
+*   A marker protocol cannot be named as the type in an `is` or as? check (e.g., `x as? ConcurrentValue` is an error).
```
benrimmington, line 132:
```suggestion
*   A marker protocol cannot be named as the type in an `is` or `as?` check (e.g., `x as? ConcurrentValue` is an error).
```

proposals/XXXX-ConcurrentValueAndConcurrentClosures.md
```diff
@@ -0,0 +1,720 @@
...
+// ... but they cannot be called by other code not protected
+// by the actor's mailbox:
+func f(a: SomeActor, myString: NSMutableString) async {
+  // error: 'NSMutableString' may not be passed across actors;
+  //        it does not conform to 'ConcurrentValue'
+  await a.doThing(string: myString)
+```
+
+
+}
```
benrimmington, line 173:
```suggestion
```

proposals/XXXX-ConcurrentValueAndConcurrentClosures.md
```diff
@@ -0,0 +1,720 @@
...
+
+The `ConcurrentValue` protocol models types that are allowed to be safely passed across concurrency domains by copying the value.  This includes value-semantic types, references to immutable reference types, internally synchronized reference types, `@concurrent` closures, and potentially other future type system extensions for unique ownership etc.
+
+Note that incorrect conformance to this protocol can introduce bugs in your program (just as an incorrect implementation of `Hashable` can break invariants), which is why the compiler checks conformance (see below).
+
+
+#### Tuple conformance to `ConcurrentValue`
+
+Swift has [hard coded conformances for tuples](https://github.com/apple/swift-evolution/blob/main/proposals/0283-tuples-are-equatable-comparable-hashable.md) to specific protocols, and this should be extended to `ConcurrentValue`, when the tuples elements all conform to `ConcurrentValue`.
```
benrimmington, line 182: (SE-0283 isn't implemented yet, and will expire in 3 months.)

proposals/XXXX-ConcurrentValueAndConcurrentClosures.md
```diff
@@ -0,0 +1,720 @@
...
+
+
+
+1. A function can be marked `@concurrent`. Any captures must also conform to `ConcurrentValue`.
+2. Closures that have `@concurrent` function type can only use by-value captures. Captures of immutable values introduced by `let` are implicitly by-value; any other capture must be specified via a capture list: \
+`let prefix: String = ... \
+var suffix: String = …  \
+strings.parallelMap { [suffix] in prefix + $0 + suffix }` \
+The types of all captured values must conform to `ConcurrentValue`.
+3. Accessors are not currently allowed to participate with the `@concurrent` system as of this proposal.  It would be straight-forward to allow getters to do so in a future proposal if there was demand for this.
```
benrimmington, line 324:
```suggestion
1.  A function can be marked `@concurrent`. Any captures must also conform to `ConcurrentValue`.

2.  Closures that have `@concurrent` function type can only use by-value captures. Captures of immutable values introduced by `let` are implicitly by-value; any other capture must be specified via a capture list:

    ```swift
    let prefix: String = ...
    var suffix: String = ...
    strings.parallelMap { [suffix] in prefix + $0 + suffix }
    ```

    The types of all captured values must conform to `ConcurrentValue`.

3.  Accessors are not currently allowed to participate with the `@concurrent` system as of this proposal.  It would be straight-forward to allow getters to do so in a future proposal if there was demand for this.
```

proposals/XXXX-ConcurrentValueAndConcurrentClosures.md
```diff
@@ -0,0 +1,720 @@
...
+    someHigherOrderFunction {
+      self.oneSyncFunction(x: 7)  // ok or not?
+    }
+  }
+}
+```
+
+
+
+## We need the compiler to know whether there is a possible concurrency domain hop or not - if so, an await is required.  Fortunately, this works out through straight-forward composition of the basic type system rules above: It is perfectly safe to use actor `self` in a non-`@concurrent` closure in an actor method, but using it in a `@concurrent` closure is treated as being from a different concurrency domain, and thus requires an `await`.
```
benrimmington, line 639:
```suggestion
We need the compiler to know whether there is a possible concurrency domain hop or not - if so, an await is required.  Fortunately, this works out through straight-forward composition of the basic type system rules above: It is perfectly safe to use actor `self` in a non-`@concurrent` closure in an actor method, but using it in a `@concurrent` closure is treated as being from a different concurrency domain, and thus requires an `await`.
```

benrimmington (commented):
(All example code could use ` ```swift ` syntax highlighting.)

proposals/XXXX-ConcurrentValueAndConcurrentClosures.md
lattner: I think it is best to keep actors conforming to protocols out of this proposal, just because it isn't nailed down yet.

PR #1269, opened 2021-02-16
[Proposal Revision] Codable synthesis for enums with associated values

drexin:


proposals/0295-codable-synthesis-for-enums-with-associated-values.md
```diff
@@ -227,13 +268,15 @@ Keys can be mapped to other names by conforming to `RawRepresentable`:
...
+  case store(key: String, Int)
 
   enum CodingKeys: String, CodingKey {
     case load = "lade"
   }
 
   enum LoadCodingKeys: String, CodingKey {
     case key = "schluessel"
+    case _1  = "wert"
   }
```
ktoso, line 280: 🎉

ktoso (approved):
Very nice update :)

proposals/0295-codable-synthesis-for-enums-with-associated-values.md
```diff
@@ -64,36 +65,37 @@ The top-level container contains a single key that matches the name of the enum
 which points to another container that contains the values as they would be encoded
 for structs and classes.
 
-Associated values can also be unlabeled, in which case they will be encoded into an
-array instead, or just the raw value if only one parameter is present.
+Associated values can also be unlabeled, in which case an identifier will be generated in the form of `_$N`, where `$N` is the 0-based position of the parameter.
```
compnerd, line 68: I think that it may be nice to add a note on how this solves a problem with changes to the shape of the `enum`.  Is there ay plan on handling something like this:

```swift
enum E {
case c(_1: String, Int)
}
```
This I believe would cause conflicts with `_1` no?  Is this treated as an overloaded case and simply not synthesized?

proposals/0295-codable-synthesis-for-enums-with-associated-values.md
```diff
@@ -64,36 +65,37 @@ The top-level container contains a single key that matches the name of the enum
...
-  ]
+  "store": {
+    "key": "MyKey",
+    "_1": 42
+  }
 }
 ```
 
-An enum case without associated values would be encoded as `true` value.
+An enum case without associated values would be encoded as an empty `KeyedContainer`,
```
compnerd, line 98: A note indicating that this allows an easier time with shape changes would be nice.

proposals/0295-codable-synthesis-for-enums-with-associated-values.md
```diff
@@ -106,23 +108,16 @@ would encode to:
 
 ```json
 {
-  "dumpToDisk": true
+  "dumpToDisk": {}
 }
 ```
 
-With the exception of the last case, this solution is closely following the default behavior of the Rust library [serde](https://serde.rs/container-attrs.html).
```
compnerd, line 113: I think that it still would be reasonable to acknowledge serde (as a foot note) .

proposals/0295-codable-synthesis-for-enums-with-associated-values.md
```diff
@@ -252,17 +295,17 @@ would encode to:
 
 This proposal specifically does not support auto-synthesis for enums with overloaded case identifiers. This decision has been made because there is no clear way to support the feature, while also allowing the model to evolve, without severe restrictions. The separate cases in an enum typically have different semantics associated with them, so it is crucial to be able to properly identify the different cases and reject unknown cases.
 
+#### Evolution with overloaded case identifiers
+
 In this proposal, we are using keys as descriminators. For overloaded case names that would not be sufficient to identify the different overloads. An alternative would be to use the full name, including labels, e.g. `"store(key:,value:)"` for `case store(key: String, value: Int)`. This leads to several problems.
 
 1. Not a valid enum case identifier, so no user customization possible
 2. Not forward/backward compatible because it changes when parameters are added
 3. Very Swift specific, making interop with non-Swift systems awkward
```
compnerd, line 304: I'm not sure I understand the third point - the key is a string value.

compnerd (commented):
Given the extensive focus given to evolution, I think that it makes sense to add a bit more context around L31R calling this out.

How does something like this sound?

There are two models of evolution, the one designated by the language and one that is useful in the regular use of enumerations.  This proposal subsets the supported cases of automatic synthesis of `Codable` where the two models align.  We believe this to be important to retain flexibility for the user to change the shape of the enumeration.

> The recommended model for enumeration cases with associated types containing more than a single value is to use a `struct` as a container for the values.  This will permit the use of the language model of `struct` evolution to alleviate problems with changing the shape of the enum by using the associated resilience model.

aniltaskiran (approved):

proposals/0295-codable-synthesis-for-enums-with-associated-values.md
drexin: I can remove it. My point was `"store(key:,value:)"` is not something commonly used in other languages and would be a bit weird if it appeared in a serialized document.

proposals/0295-codable-synthesis-for-enums-with-associated-values.md
drexin: Good point. This should be fine as long as there are no collisions, but if there are, we should emit a diagnostic. Right now it will just generate invalid code and then choke on that. So at least it rejects it, but without a useful error message. I'll add that.

proposals/0295-codable-synthesis-for-enums-with-associated-values.md
```diff
@@ -248,21 +295,24 @@ would encode to:
 }
 ```
 
+### Evolution and compatibility
+
+Enum cases can evolve in the same way as structs and classes. Adding new fields, or removing existing ones is compatible, as long as the values are optional and the identifiers for the other cases don't change. This is in opposition to the evolution model of the language, where adding or removing associated values is a source and binary breaking change. We believe that a lot of use cases benefit from the ability to evolve the model, where source and binary compatibility are not an issue, e.g. in applications, services, or for internal types. If binary compatibility is important, evolvability can still be achieved by having a single associated value of a type that acts as container for the values.
```
compnerd, line 300:
```suggestion
Enum cases can evolve in the same way as structs and classes. Adding new fields, or removing existing ones is compatible, as long as the values are optional and the identifiers for the other cases don't change. This is in opposition to the evolution model of the language, where adding or removing associated values is a source and binary breaking change. We believe that a lot of use cases benefit from the ability to evolve the model, where source and binary compatibility are not an issue, e.g. in applications, services, or for internal types. If binary compatibility is important, evolution can be supported by having a single container with all the parameters as the associated value.
```

I'm not too fond of using "evolution" since it has a specific use within the documentation at the moment for the ABI evolution and that tends to be difficult to tease apart.  Unfortunately, I do not have a good solution at the moment :-(.

compnerd (approved):
Minor worth-smithing suggestion, but otherwise, looks good to me.

proposals/0295-codable-synthesis-for-enums-with-associated-values.md
```diff
@@ -64,36 +67,37 @@ The top-level container contains a single key that matches the name of the enum
 which points to another container that contains the values as they would be encoded
 for structs and classes.
 
-Associated values can also be unlabeled, in which case they will be encoded into an
-array instead, or just the raw value if only one parameter is present.
+Associated values can also be unlabeled, in which case an identifier will be generated in the form of `_$N`, where `$N` is the 0-based position of the parameter. Using generated identifiers allows more flexbility in evolution of models than using an `UnkeyedContainer` would. If a user defined parameter has an identifier that conflicts with a generated identifier, the compiler will produce a diagnostic message.
```
benrimmington, line 70: `flexbility` → `flexibility`

```suggestion
Associated values can also be unlabeled, in which case an identifier will be generated in the form of `_$N`, where `$N` is the 0-based position of the parameter. Using generated identifiers allows more flexibility in evolution of models than using an `UnkeyedContainer` would. If a user defined parameter has an identifier that conflicts with a generated identifier, the compiler will produce a diagnostic message.
```

proposals/0295-codable-synthesis-for-enums-with-associated-values.md
```diff
@@ -106,23 +110,18 @@ would encode to:
 
 ```json
 {
-  "dumpToDisk": true
+  "dumpToDisk": {}
 }
 ```
 
-With the exception of the last case, this solution is closely following the default behavior of the Rust library [serde](https://serde.rs/container-attrs.html).
+This allows these cases to evovle in the same manner as cases with associated values, without breaking compatibility.
```
benrimmington, line 117: `evovle` → `evolve`

```suggestion
This allows these cases to evolve in the same manner as cases with associated values, without breaking compatibility.
```

proposals/0295-codable-synthesis-for-enums-with-associated-values.md
```diff
@@ -152,8 +151,54 @@ enum StoreCodingKeys: CodingKey {
 }
 ```
 
-Since cases with unlabeled parameters encode into unkeyed containers,
-no `CodingKeys` enum will be generated for them.
+The `encode(to:)` implementation would looks as follows:
```
benrimmington, line 154: `looks` → `look`

```suggestion
The `encode(to:)` implementation would look as follows:
```

proposals/0295-codable-synthesis-for-enums-with-associated-values.md
```diff
@@ -152,8 +151,54 @@ enum StoreCodingKeys: CodingKey {
...
+    try nestedContainer.encode(key, forKey: .key)
+  case let .store(key, value):
+    var nestedContainer = container.nestedContainer(keyedBy: StoreCodingKeys.self, forKey: .store)
+    try nestedContainer.encode(key, forKey: .key)
+    try nestedContainer.encode(value, forKey: .value)
+  }
+}
+```
+
+and `init(from decoder: Decoder)`
```
benrimmington, line 171:
```suggestion
and the `init(from:)` implementation would look as follows:
```

proposals/0295-codable-synthesis-for-enums-with-associated-values.md
```diff
@@ -152,8 +151,54 @@ enum StoreCodingKeys: CodingKey {
...
+
+```swift
+public init(from decoder: Decoder) throws {
+  let container = try decoder.container(keyedBy: CodingKeys.self)
+  if container.allKeys.count != 1 {
+    let context = DecodingError.Context(
+      codingPath: container.codingPath,
+      debugDescription: "Invalid number of keys found, expected one.")
+    throw DecodingError.typeMismatch(Command.self, context)
+  }
```
benrimmington, line 181: Synthesized code for structs seems to ignore any extra keys. Can you do the same for enum cases?

```json
{
  "extra_key": "extra_value",
  "load": {
    "key": "MyKey"
  }
}
```

```swift
public init(from decoder: Decoder) throws {
  let container = try decoder.container(keyedBy: CodingKeys.self)
  let allKeys = container.allKeys
  if allKeys.contains(.load) {
    /* ... */
  } else if allKeys.contains(.store) {
    /* ... */
  } else {
    let context = /* ... */
    throw DecodingError.typeMismatch(Command.self, context)
  }
}
```

proposals/0295-codable-synthesis-for-enums-with-associated-values.md
```diff
@@ -227,13 +272,15 @@ Keys can be mapped to other names by conforming to `RawRepresentable`:
...
+  case store(key: String, Int)
 
   enum CodingKeys: String, CodingKey {
     case load = "lade"
   }
 
   enum LoadCodingKeys: String, CodingKey {
     case key = "schluessel"
+    case _1  = "wert"
   }
```
benrimmington, line 284: Wouldn't the `case _1  = "wert"` cause a compiler diagnostic?
(The `.load` case only has a single associated value.)

proposals/0295-codable-synthesis-for-enums-with-associated-values.md
```diff
@@ -248,21 +295,24 @@ would encode to:
 }
 ```
 
+### Evolution and compatibility
+
+Enum cases can evolve in the same way as structs and classes. Adding new fields, or removing existing ones is compatible, as long as the values are optional and the identifiers for the other cases don't change. This is in opposition to the evolution model of the language, where adding or removing associated values is a source and binary breaking change. We believe that a lot of use cases benefit from the ability to evolve the model, where source and binary compatibility are not an issue, e.g. in applications, services, or for internal types. If binary compatibility is important, evolution can be supported by having a single container with all the parameters as the associated value.
```
benrimmington, line 300: `single container` → `single struct or class`

```suggestion
Enum cases can evolve in the same way as structs and classes. Adding new fields, or removing existing ones is compatible, as long as the values are optional and the identifiers for the other cases don't change. This is in opposition to the evolution model of the language, where adding or removing associated values is a source and binary breaking change. We believe that a lot of use cases benefit from the ability to evolve the model, where source and binary compatibility are not an issue, e.g. in applications, services, or for internal types. If binary compatibility is important, evolution can be supported by having a single struct or class with all the parameters as the associated value.
```

proposals/0295-codable-synthesis-for-enums-with-associated-values.md
drexin: Ah, that should be `StoreCodingKeys` instead. Adding keys that don't match any parameters would cause a diagnostic, yes.

proposals/0295-codable-synthesis-for-enums-with-associated-values.md
drexin: I don't think that would be a good idea. If the container contains more than one key, it is malformed. Checking for every possible key in there would be very bad for performance.

proposals/0295-codable-synthesis-for-enums-with-associated-values.md
```diff
@@ -227,13 +272,15 @@ Keys can be mapped to other names by conforming to `RawRepresentable`:
 ```swift
 enum Command: Codable {
   case load(key: String)
+  case store(key: String, Int)
 
   enum CodingKeys: String, CodingKey {
     case load = "lade"
   }
 
   enum LoadCodingKeys: String, CodingKey {
```
drexin, line 281:
```suggestion
  enum StoreCodingKeys: String, CodingKey {
```

compnerd (approved):
Thank you for all the improvements to the proposal!

PR #1270, opened 2021-02-16
[SE-0291] Amendment: package collection signing

yim-lee:
Add a section on package collection signing to describe how it works and
how it impacts SwiftPM's package-collection CLI.

### DIFF
--- a/proposals/0291-package-collections.md
+++ b/proposals/0291-package-collections.md
@@ -4,6 +4,7 @@
 * Authors: [Boris Bügling](https://github.com/neonichu), [Yim Lee](https://github.com/yim-lee), [Tom Doron](https://github.com/tomerd)
 * Review Manager: [Tom Doron](https://github.com/tomerd)
 * Status: **Accepted**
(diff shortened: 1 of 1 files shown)


proposals/0291-package-collections.md
```diff
@@ -247,6 +248,85 @@ Package collections must adhere to a specific JSON format for SwiftPM to be able
...
+
+```bash
+$ swift package-collection add https://www.example.com/packages.json --skip-signature-check
+```
+
+Since there are no trusted root certificates by default on non-Apple platforms, the signature check will always fail. SwiftPM will detect this and instruct user to either set up the `~/.swiftpm/config/trust-root-certs` directory or use `--skip-signature-check`.
+
+#### Add an unsigned package collection
+
+When adding an unsigned package collection, user must confirm their trust by passing the `--trust-signed` flag:
```
xwu, line 310: Belated, but: do you mean `--trust-unsigned`?

yim-lee: Thanks for catching that @xwu. Fixing this in https://github.com/apple/swift-evolution/pull/1281

PR #1271, opened 2021-02-17
New proposal: Fixing `rethrows` checking and introducing `rethrows(unsafe)`

slavapestov:


lattner:
Random question now that async/await are accepted: do we need `reasync(unsafe)`?

slavapestov:
For anyone following, I relied to @lattner's question in the forum discussion: https://forums.swift.org/t/pitch-fix-rethrows-checking-and-add-rethrows-unsafe/44863/5

PR #1272, opened 2021-02-18
[SE-0253] Minor edits.

dan-zheng:
- Remove grammatically wrong clause: "the naming this API often repeats the type".
  - The intended writing was probably "the naming of this API often repeats the type". Though the clause still doesn't quite make sense in context: repeated calls to `parser.applied(to:)` would be syntactically noisy but would not repeat type info.
- Minor gardening and renaming for clarity and brevity.
  - In code examples: rename `polynomial(2)` to `f(2)`.

### DIFF
--- a/proposals/0253-callable.md
+++ b/proposals/0253-callable.md
@@ -80,8 +80,8 @@ extension Polynomial {
     }
 }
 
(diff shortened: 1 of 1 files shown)


lattner:
Thx dan!

PR #1289, opened 2021-03-08
Create ReceiverFunction proposal

hfhbd:
Add syntactic sugar to a function with a receiver in a source compatible way.

````swift
func closureCurrentDesign() {
        call(action: { (foo: Foo) -> (Int) -> Void in
            return { (i: Int) in // this second closure is needed because it is returning (Int) -> Void 
                
                self.printing(i) // although the receiver of `action` is Foo, self is Bar
                foo.printing(i) // thats why this correct call with explicit receiver is necessary
            }
        })
    }

func closureNewDesign() {
        call(action: { Foo.(i: Int) -> Void in 
                self.printing(i) // self is implicit switched to Foo, the receiver of this closure
                self:Bar.printing(i) // switch explicite to the outer receiver Bar
            }
        })
    }
````
 
Swift-evolution thread: [[Pitch] Functional - Clearer Design for Functions with Receivers](https://forums.swift.org/t/pitch-functional-clearer-design-for-functions-with-receivers)

### DIFF
--- a/proposals/NNNN-receiversFunctions.md
+++ b/proposals/NNNN-receiversFunctions.md
@@ -1,0 +1,650 @@
+# Improve Syntax for Function Calls with Receivers
+
+* Proposal: [SE-NNNN](NNNN-receiversFunctions.md)
+* Authors: [Philip Wedemann](https://github.com/hfhbd)
+* Review Manager: TBD
+* Status: **Awaiting implementation**
+
+## Introduction
+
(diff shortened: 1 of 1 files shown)


DougGregor:
The [pitch](https://forums.swift.org/t/pitch-functional-clearer-design-for-functions-with-receivers/40807) for this proposal didn't gather enough interest for us to run a review. There is a [similar pitch here](https://forums.swift.org/t/pitch-scoped-functions/45486) that you might want to check out.

PR #1294, opened 2021-03-11
[Proposal] Allow interchangeable use of `CGFloat` and `Double` types

xedin:
This is a full proposal for the Swift Forums pitch [Allow interchangeable use of `CGFloat` and `Double` types](https://forums.swift.org/t/pitch-allow-interchangeable-use-of-cgfloat-and-double-types/45324)


tkremenek:
Scheduling for next Monday.

PR #1300, opened 2021-03-16
[SE-0307] Standardize arrow (`<->` and `->`) formatting in the propos…

xedin:
…al text


xedin:
@DougGregor Thank you!

PR #1310, opened 2021-03-26
[SE-0292] Update links in package registry service

benrimmington:


mattt:
@benrimmington Thanks, Ben!

PR #1311, opened 2021-03-26
Structured concurrency: Updates from review #1

ktoso:
Updates form first review

### DIFF
--- a/proposals/0304-structured-concurrency.md
+++ b/proposals/0304-structured-concurrency.md
@@ -401,7 +401,7 @@ The `UnsafeCurrentTask` is purposefully named unsafe as it *may* expose APIs whi
 
 `Task` defines APIs which are *always safe* to invoke, regardless if from the same task or from another task. Such safe APIs include `isCancelled`, `priority`.
 
-It is possible to get a `Task` out of an `UnsafeCurrentTask`
+It is possible to get a `Task` out of an `UnsafeCurrentTask`.
 
 ## Detailed design
 
@@ -540,7 +540,7 @@ extension Task.Handle {
 }
 ```
 
-Getting the handle's task allows us to check if the work we're about to wait on perhaps was already cancelled (by calling `handle.task.isCancelled`), or query at what priority the task is executing.
+Getting the handle's task allows us to check if the work we're about to wait on perhaps was already cancelled (by calling `handle.isCancelled`), or query at what priority the task is executing.
 
 #### Detached tasks
 
@@ -627,6 +627,8 @@ extension Task {
 }
 ```
 
+This function does not, by itself, spawn a new task, but rather executes the `operation` immediately, and once the `operation` returns the `withCancellationHandler` returns as well (similarily with throwing behaviors).
+
 Note that the `handler` runs `@concurrent` with the rest of the task, because it
 is executed immediately when the task is cancelled, which can happen at any
 point. If the task has already been cancelled at the point `withCancellationHandler` is called, the cancellation handler is invoked immediately, before the
@@ -677,11 +679,13 @@ extension Task {
 }
 ```
 
-#### Querying Task status from synchronous functions
+#### Synchronous functions/properties for Task status querying
 
-While all APIs defined on `Task` so far have been *instance* functions and properties, this implies that in order to query them one first would have to obtain a Task object. Obtaining a task object safely is perfomed by invoking the `Task.current()` asynchronous function, leading to the question: how do we check if a task we're running in, has been cancelled if we are in the in a synchronous function which was invoked from an asynchronous function?
+So far, a number of `Task` APIs have only been defined as instance functions on `Task`, this implies that in order to query them one first would have to obtain a `Task` object. Specifically, functions querying the state of the Task such as `isCancelled`, `checkCancellation` and `priority` have only been defined as instance functions so far.  Obtaining a task object safely is perfomed by invoking the `Task.current()` asynchronous function, leading to the question: how do we check if a task we're running in, has been cancelled if we are in the in a synchronous function which was invoked from an asynchronous function?
 
 Thankfully, thanks to the previously discussed `UnsafeCurrentTask` Swift also can offer safe, and usable from synchronous contexts, static versions of almost all functions defined in the previous sections.
+
+> These static versions of task status querying properties and functions should be preferred in normal day-to-day use, as they work equally well in async and sync contexts.
 
 ##### Cancellation
 
@@ -859,7 +863,30 @@ for try await result in group {
 }
 ```
 
-With this pattern, if a single task throws an error, the error will be propagated out of the `body` function and the task group itself. To handle errors from individual tasks, one can use a do-catch block or the `nextResult()` method.
+With this pattern, if a single task throws an error, the error will be propagated out of the `body` function and the task group itself. 
+
+To handle errors from individual tasks, one can use a do-catch block or the `nextResult()` method. For example, one might want to implement a function which starts `N` tasks, and reports back the first `m` successful results. This is simple to implement with a task group, by means of collecting results until the `results` array have accumulated `m` results, at which point we can cancel all remaining tasks and return from the group:
+
+```swift
+func gather(first m: Int, of work: [Work]) async throws -> [WorkResult] { 
+  assert(m <= work.count) 
+  
+  return Task.withGroup(resultType: WorkResult.self) { group in 
+    for w in work { 
+      await group.add { await w.doIt() } // spawn child tasks to perform the work
+    }  
+    
+    var results: [WorkResult] = []
+    while results.count <= m { 
+      switch try await group.nextResult() { 
+      case nil:             return results
+      case .success(let r): results.append(r)
+      case .failure(let e): print("Ignore error: \(e)")
+      }
+    }
+  }
+}
+```
 
 ##### Task group cancellation
 
--- a/proposals/0304-structured-concurrency.md
+++ b/proposals/0304-structured-concurrency.md
@@ -13,9 +13,11 @@
 However, the `async`/`await` proposal does not introduce concurrency *per se*: ignoring the suspension points within an asynchronous function, it will execute in essentially the same manner as a synchronous function. This proposal introduces support for [structured concurrency](https://en.wikipedia.org/wiki/Structured_concurrency) in Swift, enabling concurrent execution of asynchronous code with a model that is ergonomic, predictable, and admits efficient implementation.
 
 Swift-evolution threads:
-  [Pitch #1](https://forums.swift.org/t/concurrency-structured-concurrency/41622),
-  [Pitch #2](https://forums.swift.org/t/pitch-2-structured-concurrency/43452),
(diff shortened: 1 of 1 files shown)


proposals/0304-structured-concurrency.md
```diff
@@ -535,41 +548,37 @@ It is possible to obtain a task that the handle refers to by using `handle.task`
...
-  @discardableResult
-  static func runDetached<T: ConcurrentValue>(
-    priority: Priority = .default,
-    startingOn executor: UnownedExecutorRef? = nil,
-    operation: @escaping @concurrent () async throws -> T
-  ) -> Handle<T, Error>
-}
+/// Create a new, detached task that produces a value of type `T`.
+@discardableResult
+static func runDetached<T: Sendable>(
```
jckarter, line 565: Looks like these still need to be changed to `spawnDetached`.

proposals/0304-structured-concurrency.md
```diff
@@ -935,13 +1071,33 @@ All of the changes described in this document are additive to the language and a
...
+
+* `Task.current` now returns an optional `Task`: `var current: Task? { get }`, which depending on context it is called from might be `nil`.
+  * This API is not intended to be used "a lot", and if sure a task will be available one can always force unwrap it.
+  * Most usages of tasks are rather intended to go through the static functions/properties on Task which implicitly works on the current task.
+* `Task.unsafeCurrent` becomes a top-level `withUnsafeCurrentTask { maybeUnsafeTask in }`
+  * This better explains the intended semantics of not escaping storing the unsafe task reference.
+* Adopt `spawn` terminology for "spawning tasks" 
+  * `spawnDetached` becomes `spawnDetached`
+  * `Task.Group`'s `group.add` becomes `group.spawn`
+  * Creating a child task will eventually be `spawn <something>`
```
jckarter, line 1086: I don't think we should bullet point changes that haven't been made or decided yet.

proposals/0304-structured-concurrency.md
```diff
@@ -935,13 +1071,33 @@ All of the changes described in this document are additive to the language and a
...
+### Pitch changes
+
 * Changes in the third pitch:
   * Factored `with*Continuation` into [its own proposal](https://github.com/apple/swift-evolution/pull/1244).
   * Factored `async let` into [its own proposal](https://github.com/DougGregor/swift-evolution/pull/50).
   * `Task` becomes a `struct` with instance functions, introduction of `Task.current`, `Task.unsafeCurrent` and the `UnsafeCurrentTask` APIs
   * `Task.Group` now conforms to [the `AsyncSequence` protocol](https://github.com/apple/swift-evolution/blob/main/proposals/0298-asyncsequence.md).
-  * `Task.runDetached` and `Task.Group.add` now accept [executor](https://github.com/apple/swift-evolution/pull/1257) arguments to specify where the newly-spawned tasks are initially scheduled.
-
+  * `spawnDetached` and `Task.Group.add` now accept [executor](https://github.com/apple/swift-evolution/pull/1257) arguments to specify where the newly-spawned tasks are initially scheduled.
```
jckarter, line 1100: should be TaskGroup here

proposals/0304-structured-concurrency.md
ktoso: Okey

proposals/0304-structured-concurrency.md
ktoso: Whoops, okey going through again now

proposals/0304-structured-concurrency.md
ktoso: Thanks, and fixed a few more spots

ktoso:
Only remaining pending question is:

-  what do we do with: `Task.withCancellationHandler` -- should we move it to top level as well?  We'd spell it as `withTaskCancellationHandler` WDYT? @airspeedswift @jckarter

proposals/0304-structured-concurrency.md
```diff
@@ -155,30 +159,46 @@ Our approach follows the principles of *structured concurrency* described above.
...
-  try await Task.withGroup(resultType: Void.self) { group in
-    await group.add {
-      veggies = try await chopVegetables()
+  await withTaskGroup(of: CookingStep.self) { group in
+    group.spawn {
+      try await .vegetables(chopVegetables())
     }
-    await group.add {
-      meat = await marinateMeat()
+    group.spawn {
```
ktoso, line 185: I discuss the removal of await here a bit below. 

The short version is:
- we never implemented this "suspend as a form of back-pressure" nor are we even confident it is the right thing to do, rather than some more holistic take on the concept.
- that await was tremendously confusing to reviewers -- "it is suspending until the operation completed?" was very frequently asked.

proposals/0304-structured-concurrency.md
```diff
@@ -155,30 +159,46 @@ Our approach follows the principles of *structured concurrency* described above.
...
+      await .oven(preheatOven(temperature: 350))
+    }
+                                             
+    for await finishedStep in group {
+      switch step {
+        case .veggies(let v): veggies = v
+        case .meat(let m): meat = m
+        case .oven(let o): oven = o
+      }
```
ktoso, line 197: this had to use the enum dance due to the strong @concurrent closure checking on main now

proposals/0304-structured-concurrency.md
```diff
@@ -717,119 +765,222 @@ The isCancelled function is implemented as:
...
+///   - the group will await any not yet complete tasks,
+///     - if any of those tasks throws, the remaining tasks will be cancelled,
+///   - once the `withTaskGroup` returns the group is guaranteed to be empty.
+/// - if the body throws:
+///   - all tasks remaining in the group will be automatically cancelled.
+static func withThrowingTaskGroup<ChildTaskResult: Sendable, GroupResult>(
+  of childTaskResult: ChildTaskResult.Type,
+  returning groupResultType: GroupResult.Type = GroupResult.self,
+  body: (inout ThrowingTaskGroup<ChildTaskResult, Error>) async throws -> GroupResult
+) async rethrows -> GroupResult { ... }
```
ktoso, line 907: note that implementation wise this is a bit of hell --- we have to duplicate the `TaskGroup` struct impl twice, once normal and once `Throwing` but there is no way around it.

It is the "if any closure submitted to spawn throws, then next() must be throwing" so we can't just rely on rethrows.
It is even more complex with the AsyncSequence conformance, so sadly we have to duplicate implementation a bit here.

ktoso:
I would also suggest doing:

```swift
group.spawn -> Task.Spawned<T, Failure>

extension TaskGroup { 
  Spawned { 
    let successfully: Bool
    let handle: Task.Handle<T, Failure>
  }
}
```

since then it is just a single API to handle both:
- did we add?
- cancel this specific task
- wait for this specific task

proposals/0304-structured-concurrency.md
```diff
@@ -717,119 +765,238 @@ The isCancelled function is implemented as:
...
+  public struct Spawned: Sendable {
+    /// Returns `true` if the task was successfully spawned in the task group,
+    /// `false` otherwise which means that the group was already cancelled and
+    /// refused to accept spawn a new child task.
+    public var successfully: Bool 
+
+    /// Task handle for the spawned task group child task,
+    /// or `nil` if it was not spawned successfully.
+    public let handle: Task.Handle<ChildTaskResult, Never>?
+  }
```
ktoso, line 957: Another alternative design for this is with throwing `CancellationError` however then ONLY the `withThrowingGroup` could do so, and we'd need some new name for `trySpawn` (try spawn is terrible since it becomes `try group.trySpawn { }`)

```swift
withTaskGroup(of: Int.self) { group in 
  // always spawns:
  group.spawn {}
}

withThrowingTaskGroup(of: Int.self) { group in 
  // always spawns:
  group.spawn {} 

  // spawns, unless the group/owning-task were cancelled, then it throws:
  try group.spawnUnlessCancelled {} 
}
```

So that'd be another option; and the spawn would be Void returning OR returning `Task.Handle` and `Task.Handle?` in the "may fail" case.

FYI @airspeedswift for when you have the time to review; plenty of various tradeoffs here, I hope I explained them well though, it is mostly: 
- ability to not spawn in a cancelled group
- ability to always spawn (as async lets would do, if we express them as "implicitly in a group" which I think we should, but I did not manage to do this with my beginner SIL skill...)
- ability to get a handle for a specific spawned task

The Spawned addressess all in one API call though one has to search for it. The throwing versions are very annoying, and can only be available in `withThrowingTaskGroup`. So there's some tradeoffs to make here I'd need input on. Thanks in advance!

proposals/0304-structured-concurrency.md
```diff
@@ -717,119 +765,238 @@ The isCancelled function is implemented as:
...
-      operation: @concurrent @escaping () async throws -> TaskResult
-  ) async -> Bool
+      operation: @concurrent @escaping () async -> TaskResult
+  ) -> Self.Spawned
+  
+  public struct Spawned: Sendable {
+    /// Returns `true` if the task was successfully spawned in the task group,
+    /// `false` otherwise which means that the group was already cancelled and
+    /// refused to accept spawn a new child task.
+    public var successfully: Bool
```
ktoso, line 952: (This of course is just `var sucessfully: Bool { handle != nil }`  so does not need storage)

proposals/0304-structured-concurrency.md
jckarter: You can still `spawnUnlessCancelled` inside a non-throwing group, you'd just need to handle the `CancellationError` locally and not let it propagate out of the group block.

jckarter:
I don't think that having `spawn` return a more elaborate return value is much better than returning `Bool`—you still have the problem of wanting to ignore the return value most of the time, and it then being easy to forget to pay attention to it when you need. Being able to get a `Handle` to child tasks also strikes me as an anti-feature, since it creates the "two classes of Handle" problem where some handles implicitly blow up and start trapping if you access them out of scope, and it also enables unstructured use of the handle within the group.

proposals/0304-structured-concurrency.md
jckarter: If the error handling ceremony is too onerous, another option is to have `spawnUnlessCancelled` return a `Bool` like `add` was previously proposed to, but remove the `@discardableResult` from the declaration, so that you have to look at the result. I don't think it needs to return more information than a `Bool`, though.

ktoso:
> I don't think that having spawn return a more elaborate return value is much better than returning Bool—you still have the problem of wanting to ignore the return value most of the time, and it then being easy to forget to pay attention to it when you need. 

Yeah I think you're right here... sounds like a not ignorable bool return sounds great here, thanks!

> Being able to get a Handle to child tasks also strikes me as an anti-feature, since it creates the "two classes of Handle" problem where some handles implicitly blow up and start trapping if you access them out of scope, and it also enables unstructured use of the handle within the group.

That's a great point, thanks that convienced me that indeed is an anti feature.

proposals/0304-structured-concurrency.md
ktoso: Great points, thank you! I'll adjust this right away so it is part of the review 👍

PR #1314, opened 2021-04-01
More SE-0304 Structured Concurrency review updates

jckarter:
- Clarify that the `@Sendable` capture analysis for child tasks is not fundamentally unsafe, but
  would require special compiler understanding to handle correctly
- Clear up lingering references to `add`, and clarify some fine points of the behavior of
  `spawn` and `spawnUnlessCancelled` into an already-cancelled group


ktoso (approved):
LGTM! Thanks a lot, and very cool if we'd some day implement that special sugar :)

PR #1317, opened 2021-04-03
Add placeholder types proposal

Jumhyn:


proposals/nnnn-placeholder-types.md
```diff
@@ -0,0 +1,349 @@
...
+    func frobnicate7() -> Bar<_, _> {
+        return Bar<_, U>(t: 42, u: 42)
+    }
+    func frobnicate8() -> Bar<_, U> {
+        return Bar<_, _>(t: 42, u: 42)
+    }
+}
+```
+
+Under this proposal, only `frobnicate1`, `frobnicate3` and `frobnicate6` would compile without error (`frobnicate1`, of course, compiles without this proposal as well), since all others have placeholders appearing in at least one position in the function signature.
```
xedin, line 233: Nit: `frobnicate1` doesn't exist, it's just `frobnicate`.

proposals/nnnn-placeholder-types.md
```diff
@@ -0,0 +1,349 @@
...
+
+```
+let losslessStringConverter = Double.init as (String) -> _?
+```
+
+Swift-evolution threads: [Partial type annotations ](https://forums.swift.org/t/partial-type-annotations/41239), [Placeholder types](https://forums.swift.org/t/placeholder-types/41329)
+
+## Motivation
+
+Swift's type inference system is quite powerful, but there are many situations where it is impossible (or simply infeasible) for the compiler to work out the type of an expression, or where the user needs to override the default types worked out by the compiler. The example above of an overloaded function `foo` is one such example.
```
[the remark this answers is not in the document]
Jumhyn: Yes, thank you!

```suggestion
Swift's type inference system is quite powerful, but there are many situations where it is impossible (or simply infeasible) for the compiler to work out the type of an expression, or where the user needs to override the default types worked out by the compiler. Directly referencing the heavily-overloaded `Double.init` initializer, as seen above, is one such situation where the compiler does not have the necessary context to determine the type of the expression without additional context.
```

proposals/nnnn-placeholder-types.md
Jumhyn:
```suggestion
Under this proposal, only `frobnicate`, `frobnicate3` and `frobnicate6` would compile without error (`frobnicate1`, of course, compiles without this proposal as well), since all others have placeholders appearing in at least one position in the function signature.
```

proposals/nnnn-placeholder-types.md
```diff
@@ -0,0 +1,349 @@
...
+## Motivation
+
+Swift's type inference system is quite powerful, but there are many situations where it is impossible (or simply infeasible) for the compiler to work out the type of an expression, or where the user needs to override the default types worked out by the compiler. Directly referencing the heavily-overloaded `Double.init` initializer, as seen above, is one such situation where the compiler does not have the necessary context to determine the type of the expression without additional context.
+
+Fortunately, Swift provides several ways for the user to provide type information explicitly. Common forms are:
+
+* Variable type annotations:
+
+```swift
+let stringTransform: (String) -> Double = foo
```
xedin, line 37: There are a couple more places of `foo` that remain.

proposals/nnnn-placeholder-types.md
```diff
@@ -0,0 +1,349 @@
...
+```
+
+* Type coercion via `as` (seen above):
+
+```swift
+let stringTransform = foo as (String) -> Double
```
xedin, line 43: And here.

PR #1318, opened 2021-04-04
Update index.html

ricksentam:
Updated the index to suit for html


lattner:
What does this patch do?  Please keep the license header at the top of the file.

benrimmington:
This patch is invalid HTML. The [`title`](https://html.spec.whatwg.org/multipage/semantics.html#the-title-element) element:

* must contain only text nodes (i.e. not child elements);

* must be defined zero or one times (i.e. not twice) per document.

PR #1321, opened 2021-04-07
Actors revision 2

DougGregor:


proposals/0306-actors.md
```diff
@@ -268,42 +269,10 @@ extension BankAccount {
...
+* The closure passed to `Rask.runDetached` is non-isolated because that function requires a `@Sendable` function to be passed to it.
```
ktoso, line 274:
```suggestion
* The closure passed to `detach` is non-isolated because that function requires a `@Sendable` function to be passed to it.
```

proposals/0306-actors.md
```diff
@@ -542,15 +507,12 @@ extension BankAccount {
 }  
 ```
 
-Actors are similar to classes in all respects independent of isolation: actor types can have `static` and `class` methods, properties, and subscripts. All of the attributes that apply to classes apply to actors in much the same way, except where those semantics conflict with actor isolation.
```
ktoso, line 545: `static` stay though I hope? Just removal of the test here to compare with classes; no reason not to do static IMHO, very useful for IDs and stuff

proposals/0306-actors.md
```diff
@@ -621,7 +576,7 @@ actor A {
   
   func useAF(array: [Int]) {
     array.map(self.f)                     // okay
-    Task.runDetached(operation: self.g)   // error: self.g has non-sendable type () -> Double that cannot be converted to a @Sendable function type
+    detach(operation: self.g)   // error: self.g has non-sendable type () -> Double that cannot be converted to a @Sendable function type
```
ktoso, line 579:
```suggestion
    detach(operation: self.g)             // error: self.g has non-sendable type () -> Double that cannot be converted to a @Sendable function type
```

proposals/0306-actors.md
```diff
@@ -832,32 +787,70 @@ If we can address the above, task-chain reentrancy can be introduced into the ac
...
 
-* Actor inheritance makes it easier to port existing class hierarchies to get the benefits of actors. Without actor inheritance, such porting will also have to contend with (e.g.) replacing superclasses with protocols and explicitly-specified stored properties at the same time.
-* The lack of inheritance in actors won't prevent users from having to understand the complexities of inheritance, because inheritance will still be used pervasively with classes.
-* The design and implementation of actors naturally admits inheritance. Actors are fundamentally class-like, the semantics of inheritance follow directly from the need to maintain actor isolation. The implementation of actors is essentially as "special classes", so it supports all of these features out of the box. There is little benefit to the implementation from eliminating the possibility of inheritance of actors.
+### Cross-actor lets
```
ktoso, line 799: I'm confused what this section is claiming; 

Are we removing this "very permissiveness" from local actors or not by default or keeping it and this future work explains... I'm not sure what?
 
--- 

The permissive synchronous let access is definitely a bit "weird" and "not pure actor model" etc. but I also don't think it's a huge deal to do so. As long as the type allowed for such access is required to be `Sendable`. 

Distributed actors obviously will not allow such synchronous access either way, _even if_ `nonisolated let x: X` so I don't think it matters either way for distributed actors and the difference will exist anyway. Though one could reach into `distributed actor DA { nonisolated let x: X }` with `da.whenLocal { $0.x }` without having to do any extra "with unsafe actor state" dance etc. Long story short: distributed actors don't matter much for allowing cross-actor lets or not, because in all shapes (even non isolated) it's banned thing.

proposals/0306-actors.md
```diff
@@ -634,7 +589,7 @@ These restrictions follow from the actor isolation rules for the "desugaring" of
 extension A {
   func useAFDesugared(a: A, array: [Int]) {
     array.map { f($0) } )      // okay
-    Task.runDetached { g() }   // error: self is non-isolated, so call to `g` cannot be synchronous
+    detach { g() }   // error: self is non-isolated, so call to `g` cannot be synchronous
```
ktoso, line 592:
```suggestion
    detach { g() }             // error: self is non-isolated, so call to `g` cannot be synchronous
```

proposals/0306-actors.md
```diff
@@ -832,32 +787,70 @@ If we can address the above, task-chain reentrancy can be introduced into the ac
...
 
-Like classes, actors as proposed allow inheritance. However, actors and classes cannot be co-mingled in an inheritance hierarchy, so there are essentially two different kinds of type hierarchies. It has been [proposed](https://docs.google.com/document/d/14e3p6yBt1kPrakLcEHV4C9mqNBkNibXIZsozdZ6E71c/edit#) that actors should not permit inheritance at all, because doing so would simplify actors: features such as method overriding, initializer inheritance, required and convenience initializers, and inheritance of protocol conformances would not need to be specified, and users would not need to consider them. The [discussion thread](https://forums.swift.org/t/actors-are-reference-types-but-why-classes/42281) on the proposal to eliminate inheritance provides several reasons to keep actor inheritance:
+Subsequent review discussion determined that the conceptual cost of actor inheritance outweighed its usefulness, so it has been removed from this proposal. The form that actor inheritance would take in the language is well-understand from prior iterations of this proposal and its implementation, so this feature could be re-introduced at a later time.
```
ktoso, line 797: 👍 I warmed up to this as well once I realized the hell that class style initializers are 🤣+😭

ktoso:
GREAT changes, I love where we're heading 🥳 

Just some typos; one question (about `static`)

ktoso:
Should it become illegal to say `open actor` and `final actor`?

ktoso (approved):

proposals/0306-actors.md
DougGregor: static is fine, I'll comment on it

lattner: Might be good to clarify the isolation behavior (or not) of `static var`'s if they are allowed.  It might be worth banning them from actors?

DougGregor: Sure, I can clarify that as well.

proposals/0306-actors.md
DougGregor: We will call this out for discussion in the second review

ktoso:
Thanks for additions, nothing more to nitpick on from my side :-)

proposals/0306-actors.md
ktoso: Sounds good such properties to ban even, agreed :-)

PR #1329, opened 2021-04-16
SE-0311 typos

0xTim:
A couple of small typo fixes after an initial review of SE-0311

### DIFF
--- a/proposals/0311-task-locals.md
+++ b/proposals/0311-task-locals.md
@@ -1,6 +1,6 @@
 # Task Local Values
 
-* Proposal: [SE-0311](0311-task-local.md)
+* Proposal: [SE-0311](0311-task-locals.md)
 * Authors: [Konrad 'ktoso' Malawski](https://github.com/ktoso)
 * Review Manager: [John McCall](https://github.com/rjmccall)
 * Status: **Active Review (April 16...26, 2021)**
@@ -145,7 +145,7 @@ await Task.withLocal(\.requestID, boundTo: "1234-5678") {
 func nested() async {
   await nestedAgain() // "1234-5678"
   
-  Task.withLocal(\.requestID, boundTo: "xxxx-zzzz") { 
+  await Task.withLocal(\.requestID, boundTo: "xxxx-zzzz") { 
     await nestedAgain() // "xxxx-zzzz"
   }
 } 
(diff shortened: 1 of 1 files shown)


proposals/0311-task-locals.md
```diff
@@ -1,7 +1,7 @@
 # Task Local Values
 
-* Proposal: [SE-0311](0311-task-locals.md)
-* Author: [Konrad 'ktoso' Malawski](https://github.com/ktoso)
+* Proposal: [SE-0311](0311-task-local.md)
```
ktoso, line 3: hm? `proposals/0311-task-locals.md`

proposals/0311-task-locals.md
```diff
@@ -196,7 +196,7 @@ public protocol TaskLocalKey {
 }
 ```
 
-If the [ConcurrentValue` proposal](https://forums.swift.org/t/pitch-3-concurrentvalue-and-concurrent-closures/43947) is accepted, it would be an excellent choice to limit the values stored within task locals to only ConcurrentValues. As access to them may be performed by the task which set the value, and any of its children, therefore it should be safe to use in such concurrent access scenarios. Practically speaking, task local values should most often be simple value types, such as identifiers, counters or similar.
+If the [`ConcurrentValue` proposal](https://forums.swift.org/t/pitch-3-concurrentvalue-and-concurrent-closures/43947) is accepted, it would be an excellent choice to limit the values stored within task locals to only ConcurrentValues. As access to them may be performed by the task which set the value, and any of its children, therefore it should be safe to use in such concurrent access scenarios. Practically speaking, task local values should most often be simple value types, such as identifiers, counters or similar.
```
ktoso, line 199: Might right away rename this to Sendable

```suggestion
If the [`Sendable` proposal](https://forums.swift.org/t/pitch-3-concurrentvalue-and-concurrent-closures/43947) is accepted, it would be an excellent choice to limit the values stored within task locals to only Sendable. As access to them may be performed by the task which set the value, and any of its children, therefore it should be safe to use in such concurrent access scenarios. Practically speaking, task local values should most often be simple value types, such as identifiers, counters or similar.
```

proposals/0311-task-locals.md
```diff
@@ -145,13 +145,13 @@ await Task.withLocal(\.requestID, boundTo: "1234-5678") {
 func nested() async {
   await nestedAgain() // "1234-5678"
   
-  await Task.withLocal(\.requestID, boundTo: "xxxx-zzzz") { 
+  Task.withLocal(\.requestID, boundTo: "xxxx-zzzz") {
```
ktoso, line 148: Binding operations must be async -- notice that inside the closure we call an async function, so the withLocal must be async

```suggestion
  await Task.withLocal(\.requestID, boundTo: "xxxx-zzzz") { 
```

ktoso:
Thank you Tim!

PR #1330, opened 2021-04-16
[SE-0311] Update the table of contents

benrimmington:
Alternatively:

<https://github.blog/changelog/2021-04-13-table-of-contents-support-in-markdown-files/>


ktoso:
Thanks, it's nice to have the ToC in the source so let's get that in even with the GH feature present. Thanks!

PR #1335, opened 2021-04-20
0311 print fixes

kirilltitov:
Plain text print to actual call


kirilltitov:
cc @ktoso

ktoso:
Sure, might as well -- thank you.

PR #1348, opened 2021-04-29
fix: Adding arrow functions and removing redundancy

D-Freitas:
I believe that the changes made will help in the readability of the code. I removed some redundancies and added arrow functions.


krilnon:
We originally avoided using any ES6 features since the browser compatibility wasn't broadly there (in ~2016/2017); did you do any background research on how support has changed since then?

PR #1355, opened 2021-05-06
Add AsyncStream proposal

natecook1000:
Based on the pitch here: https://forums.swift.org/t/pitch-asyncstream-and-asyncthrowingstream/47820


DougGregor:
I'll manage this review.

PR #1368, opened 2021-05-21
Structured concurrency revisions for third review

DougGregor:


ktoso (approved):
LGTM, I believe this has everything now. Thank you Doug!

I checked priorities, APIs, function names, the inheritance semantics and where things are defined etc. I don't think we missed anything 👍

proposals/0304-structured-concurrency.md
```diff
@@ -6,6 +6,54 @@
...
+   * [Child tasks](#child-tasks)
+   * [Partial tasks](#partial-tasks)
+   * [Executors](#executors)
+   * [Task priorities](#task-priorities)
+   * [Priority Escalation](#priority-escalation)
+* [Proposed solution](#proposed-solution)
+   * [Task groups and child tasks](#task-groups-and-child-tasks)
+   * [Asynchronous programs](#asynchronous-programs)
+   * [Cancellation](#cancellation)
+   * [Un-structured tasks](#un-structured-tasks)
```
benrimmington, line 24:
```suggestion
   * [Unstructured tasks](#unstructured-tasks)
```

proposals/0304-structured-concurrency.md
```diff
@@ -6,6 +6,54 @@
...
+      * [Detached tasks](#detached-tasks)
+* [Detailed design](#detailed-design)
+   * [Task API](#task-api)
+      * [The Task type](#the-task-type)
+      * [UnsafeCurrentTask type](#unsafecurrenttask-type)
+      * [Task priorities](#task-priorities-1)
+      * [Un-structured tasks](#un-structured-tasks-1)
+            * [Priority propagation](#priority-propagation)
+            * [Actor context propagation](#actor-context-propagation)
+            * [Implicit "self" in async {} closures](#implicit-self-in-async--closures)
```
benrimmington, line 35:
```suggestion
      * [Unstructured tasks](#unstructured-tasks-1)
         * [Priority propagation](#priority-propagation)
         * [Actor context propagation](#actor-context-propagation)
         * [Implicit "self"](#implicit-self)
```

proposals/0304-structured-concurrency.md
```diff
@@ -406,83 +433,148 @@ func chop(_ vegetable: Vegetable) async throws -> Vegetable {
...
+    throw CancellationError() 
   } 
   // chop some more, chop chop chop ...
 }
 ```
 
 Note also that no information is passed to the task about why it was cancelled.  A task may be cancelled for many reasons, and additional reasons may accrue after the initial cancellation (for example, if the task fails to immediately exit, it may pass a deadline). The goal of cancellation is to allow tasks to be cancelled in a lightweight way, not to be a secondary method of inter-task communication.
 
-### Access from synchronous functions
+### Un-structured tasks
```
benrimmington, line 444:
```suggestion
### Unstructured tasks
```

proposals/0304-structured-concurrency.md
```diff
@@ -406,83 +433,148 @@ func chop(_ vegetable: Vegetable) async throws -> Vegetable {
...
 
-As already shown in the above examples, it is possible to call functions that inspect the "current task" in order to check if it was e.g. cancelled. However, those functions are on purpose not `async` and _can_ be called from synchronous functions as well.
+So far all types of tasks we discussed were child-tasks and respected the primary rule of structured concurrency: that a *child task* cannot live longer than *parent task* (or scope) in which it was created. This is both true for task groups as well as the `async let` proposal which is being reviewed in parallel.
 
-This is possible because of the existence of the `withUnsafeCurrentTask` API:
+Sometimes however, these rigid rules end up being too restrictive. We might need to create new tasks whose lifetime is not bound to the creating task, for example in order to fire-and-forget some operation or to initiate asynchronous work from synchronous code. Un-structured tasks are not able to utilize some of the optimization techniques wrt. allocation and metadata propagation as child-tasks are, however they remain a very important building block especially for more free-form usages and integration with legacy APIs.
```
benrimmington, line 448:
```suggestion
Sometimes however, these rigid rules end up being too restrictive. We might need to create new tasks whose lifetime is not bound to the creating task, for example in order to fire-and-forget some operation or to initiate asynchronous work from synchronous code. Unstructured tasks are not able to utilize some of the optimization techniques wrt. allocation and metadata propagation as child-tasks are, however they remain a very important building block especially for more free-form usages and integration with legacy APIs.
```

proposals/0304-structured-concurrency.md
```diff
@@ -406,83 +433,148 @@ func chop(_ vegetable: Vegetable) async throws -> Vegetable {
...
+
+All un-structured tasks are represented by a task handle, which can be used to retrieve the value (or thrown error) produced by the task, cancel the task, or perform queries of the task's status. A new task can be created with the `Task { ... }` initializer. For example:
```
benrimmington, line 450:
```suggestion
All unstructured tasks are represented by a task handle, which can be used to retrieve the value (or thrown error) produced by the task, cancel the task, or perform queries of the task's status. A new task can be created with the `Task { ... }` initializer. For example:
```

proposals/0304-structured-concurrency.md
```diff
@@ -406,83 +433,148 @@ func chop(_ vegetable: Vegetable) async throws -> Vegetable {
...
-extension Task { 
-  static var current: Task? { get }
-}
+dinnerHandle.cancel()
 ```
 
-The `UnsafeCurrentTask` is purposefully named unsafe as it *may* expose APIs which can *only* be invoked safely from within task itself, and would exhibit undefined behavior if used from another task. It is therefore unsafe to store and "use later" an `UnsafeCurrentTask`. Examples of such unsafe API are interacting with task local values on a task object, which must be equal to the "current" task to be performed safely. This is by design, and offers the runtime optimization opportunities for the normal, and safe, access patterns to task storage.
+#### Context inheritance
+
+Un-structured tasks created with the `Task` initializer inherit important metadata information from the context in which it is created, including priority, task-local values, and actor isolation.
```
benrimmington, line 472:
```suggestion
Unstructured tasks created with the `Task` initializer inherit important metadata information from the context in which it is created, including priority, task-local values, and actor isolation. 
```

proposals/0304-structured-concurrency.md
```diff
@@ -406,83 +433,148 @@ func chop(_ vegetable: Vegetable) async throws -> Vegetable {
...
-The `UnsafeCurrentTask` is purposefully named unsafe as it *may* expose APIs which can *only* be invoked safely from within task itself, and would exhibit undefined behavior if used from another task. It is therefore unsafe to store and "use later" an `UnsafeCurrentTask`. Examples of such unsafe API are interacting with task local values on a task object, which must be equal to the "current" task to be performed safely. This is by design, and offers the runtime optimization opportunities for the normal, and safe, access patterns to task storage.
+#### Context inheritance
+
+Un-structured tasks created with the `Task` initializer inherit important metadata information from the context in which it is created, including priority, task-local values, and actor isolation. 
+
+If called from the context of an existing task:
 
-`Task` defines APIs which are *always safe* to invoke, regardless if from the same task or from another task. Such safe APIs include `isCancelled`, `priority`.
+- inherit the priority of the current task the _synchronous_ function is executing on
+- inherit all task-local values by copying them to the new un-structured task
```
benrimmington, line 477:
```suggestion
- inherit all task-local values by copying them to the new unstructured task
```

proposals/0304-structured-concurrency.md
```diff
@@ -406,83 +433,148 @@ func chop(_ vegetable: Vegetable) async throws -> Vegetable {
...
+
+If called from a context that is _not_ running inside a task:
+
+- consult the runtime and infer the best possible priority to use (e.g. by asking for current thread priority),
+- even though there is no `Task` to inherit task-local values from, check the fallback mechanism for any task-locals stored for the current synchronous context (this is discussed in depth in the SE-Thread Local Values proposal)
+- execute on the global concurrent executor and be non-isolated with regard to any actor.
+
+#### Detached tasks
+
+A *detached task* is an un-structured task that is independent of the context in which it is created, meaning that it does not inherit priority, task-local values, or the actor context. A new detached task can be created with the  `Task.detached` function:
```
benrimmington, line 490:
```suggestion
A *detached task* is an unstructured task that is independent of the context in which it is created, meaning that it does not inherit priority, task-local values, or the actor context. A new detached task can be created with the  `Task.detached` function:
```

proposals/0304-structured-concurrency.md
```diff
@@ -508,154 +607,248 @@ struct UnsafeCurrentTask: Equatable, Hashable {}
...
 
 The `priority` operation queries the priority of the task.
 
-Task priorities are set on task creation (e.g., `detach` or `TaskGroup.spawn`) and can be escalated later, e.g., if a higher-priority task waits on the task handle of a lower-priority task.
-
-The `currentPriority` operation queries the priority of the currently-executing task. Task priorities are set on task creation (e.g., `detach` or `TaskGroup.spawn`) and can be escalated later, e.g., if a higher-priority task waits on the task handle of a lower-priority task.
+The `currentPriority` operation queries the priority of the currently-executing task. Task priorities are set on task creation (e.g., `Task.detached` or `TaskGroup.async`) and can be escalated later, e.g., if a higher-priority task waits on the task handle of a lower-priority task.
 
-#### Task handles
+#### Un-structured tasks
```
benrimmington, line 676:
```suggestion
#### Unstructured tasks
```

proposals/0304-structured-concurrency.md
```diff
@@ -508,154 +607,248 @@ struct UnsafeCurrentTask: Equatable, Hashable {}
...
 
-```swift
-func eat(mealHandle: Task.Handle<Meal, Error>) async throws {
-  let meal = try await mealHandle.get()
-  meal.eat() // yum
-}
-```
+By default, the new task will be initially scheduled on the default global concurrent executor. Once custom executors are introduced in another proposal, these will be able to take an executor parameter to determine on which executor to schedule the new task instead.
+
+###### Priority propagation
```
benrimmington, line 702:
```suggestion
##### Priority propagation
```

proposals/0304-structured-concurrency.md
```diff
@@ -508,154 +607,248 @@ struct UnsafeCurrentTask: Equatable, Hashable {}
...
+The `Task` initializer propagates priority from the point where it is called to the detached task that it creates:
+
+1. If the synchronous code is running on behalf of a task (i.e., `withUnsafeCurrentTask` provides a non-`nil` task), use the priority of that task;
+2. If the synchronous code is running on behalf of the "UI" thread, use `.userInitiated`; otherwise
+3. Query the system to determine the priority of the currently-executing thread and use that.
 
-Task handles also provide the ability to cancel a task programmatically:
+The implementation will also propagate any other important OS-specific information from the synchronous code into the asynchronous task.
+
+###### Actor context propagation
```
benrimmington, line 712:
```suggestion
##### Actor context propagation
```

proposals/0304-structured-concurrency.md
```diff
@@ -508,154 +607,248 @@ struct UnsafeCurrentTask: Equatable, Hashable {}
...
 
-As noted elsewhere, cancellation is cooperative: the task will note that it has been cancelled and can choose to return earlier (either via a normal return or a thrown error, as appropriate). `isCancelled` can be used to determine whether a particular task was ever cancelled.
+In a sense, the `Task` initializer counteracts the normal influence of `@Sendable` on a closure within an actor. Specifically, [SE-0306](https://github.com/apple/swift-evolution/blob/main/proposals/0306-actors.md#closures) states that `@Sendable` closure are not actor-isolated:
 
-It is possible to obtain a task that the handle refers to by using `handle.task`:
+> Actors prevent this data race by specifying that a `@Sendable` closure is always non-isolated. 
+
+Such semantics, where the closure is both `@Sendable` and actor-isolated, are only possible because the closure is also `async`. Effectively, when the closure is called, it will immediately "hop" over to the actor's context so that it runs within the actor.
+
+###### Implicit "self" in `async {}` closures
```
benrimmington, line 739:
```suggestion
##### Implicit "self"
```

benrimmington (commented):
I noticed a problem with the rendered table of contents, and I think "unstructured" doesn't need to be hyphenated.

(I haven't read the entire proposal yet.)

proposals/0304-structured-concurrency.md
```diff
@@ -6,6 +6,54 @@
...
 
+## Table of contents
+
+* [Introduction](#introduction)
+* [Motivation](#motivation)
+* [Structured concurrency](#structured-concurrency-1)
+   * [Tasks](#tasks)
+   * [Child tasks](#child-tasks)
+   * [Partial tasks](#partial-tasks)
+   * [Executors](#executors)
```
benrimmington, line 17: Should "partial task" be renamed to "job" in these two sections?

```suggestion
   * [Jobs](#jobs)
   * [Executors](#executors)
```

DougGregor: Hmm. That's the term we use in the implementation nowadays. Sure, that makes sense. Thanks!

DougGregor:
Thank you @benrimmington

proposals/0304-structured-concurrency.md
```diff
@@ -719,79 +910,34 @@ func download(url: URL) async throws -> Data? {
...
-Similarly, a static `currentPriority` property is available to check the priority of the currently executing task:
-
-```swift
-extension Task { 
-  static var currentPriority: Task.Priority { 
-    Task.current?.priority ?? Task.Priority.default
-  }
-}
-```
+The sleep function accepts a plain integer as nanoseconds to sleep for which mirrors known top-level functions performing the same action in the synchronous world. Because use-sites look quite explicit in the way they have to previous this call with an `await` keyword (`await Task.sleep(nanos)`), we prefer to use the well-known `sleep` word rather than introduce new words for this functionality.
```
benrimmington, line 938: [`Darwin.sleep(_:)`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/sleep.html) has a `seconds: CUnsignedInt` parameter. Could the proposal specify which other known top-level functions accept nanoseconds?

Using `public static func sleep(forTimeInterval seconds: Double) async` would fit in with the [existing](https://developer.apple.com/documentation/foundation/thread/1413673-sleep) Foundation API.

Typo: "previous" → "prefix"

```suggestion
The sleep function accepts a plain integer as nanoseconds to sleep for which mirrors known top-level functions performing the same action in the synchronous world. Because use-sites look quite explicit in the way they have to prefix this call with an `await` keyword (`await Task.sleep(nanos)`), we prefer to use the well-known `sleep` word rather than introduce new words for this functionality.
```

benrimmington:
apple/swift#36905 was recently merged into `release/5.5`, but the main branch and *§Detailed design* appear to have outdated documentation comments.

proposals/0304-structured-concurrency.md
DougGregor: Let's fix the typo here and discuss the `sleep` API as part of the next review.

benrimmington:
Pull request at DougGregor/swift-evolution#65 has some more suggestions.

PR #1375, opened 2021-05-27
Update SE-0302's status to implemented

li-bei:


lattner:
Thanks!

lattner:
really thanks :)

DougGregor:
I'm going to revert this, because we have not sufficiently aligned the implementation with the proposal to call it "done". For example, we aren't yet parsing `@unchecked Sendable`.

PR #1388, opened 2021-06-12
[NFC] Rename file for SE-0302 to match accepted names.

varungandhi-apple:
Noticed the filename was out of date.


benrimmington:
It would be nice to rename this file (my suggestion is `0302-sendable.md`). However, doing so will break links in the forums, on [WWDC][] video pages, and on external sites. A symlink with the old name could be added, but that will probably break the [status page][] and other tools.

[WWDC]: <https://developer.apple.com/videos/play/wwdc2021/10133/>
[status page]: <https://apple.github.io/swift-evolution/>

rjmccall:
Yeah, that's why we didn't rename it when the proposal was retitled.  If all the tooling gets updated, there are several different proposals we could do this to, but I'm not sure it's really worth the trouble.

PR #1390, opened 2021-06-16
[SE-0316] Revise proposal for a second review

DougGregor:
* Changes for the second review:
    * Added the `GlobalActor` protocol, to which all global actors implictly conform.
    * Remove the requirement that all global and static variables be annotated with a global actor.
    * Added a grammar for closure attributes.
    * Clarified the interaction between the main actor and the main thread.


ktoso (approved):
Changes look good 👍

PR #1394, opened 2021-06-21
Add section on backward-deployability

AquaGeek:
The goal is to make it very clear from the outset what proposals can be used in apps that target previous OS versions.

Here is the corresponding Swift forum thread: https://forums.swift.org/t/pitch-meta-call-out-backwards-deployability-in-evolution-proposals/49534


DougGregor:
Per the discussion thread, what we are looking for here is to talk about the runtime impact (which is an aspect of the implementation that can affect design) and not backward deployment (which is a vendor concern, not a Swift language concern).

PR #1396, opened 2021-06-23
[DRAFT]  Add shared storage for property wrappers

iillx:
Opening this for collaboration purposes only, no intention to merge!


proposals/0000-shared-storage-property-wrapper.md
```diff
@@ -0,0 +1,419 @@
...
+
+// not okay
+struct ContainerB {
+  var minDay: Int
+  var maxDay: Int
+  @Clampped(minDay...maxDay3) var weekday = 3
+}
+```
+
+As mentioned before, the compiler will enforce that the shared property initialization only happens once. For that reason, the `shared` property is required to be initialized from a `shared` parameter.
```
hborla, line 180: We briefly discussed this, but I think it makes more sense to not require the line `self.storage = storage`, because it enforces the model that the shared storage is already initialized. What we need to do instead is figure out a way to "match up" the shared parameter with the shared storage, but I think it's better to have new syntax or a new rule for that instead of writing it in a way that looks like stored property initialization.

proposals/0000-shared-storage-property-wrapper.md
```diff
@@ -0,0 +1,419 @@
...
+Inside the wrapper's initializer, assigning the shared value to the `shared` property is handled by the compiler, so there's not need to explicitly do it. If the initializer itself needs access to storage properties, it should be done through the `shared` parameter. 
+
+```swift
+shared let storage: RangeStorage<Value>
+
+init(wrappedValue: Value, shared: RangeStorage<Value>) {
+  self.value = shared.clamp(wrappedValue)
+
+  // referencing `storage` inside the initializer is not okay since 
+  // it might not have been initialized yet?
```
iillx, line 127: @hborla I think we briefly touched on this but I'm not sure if it would really be a problem to access `storage` directly. Since it would be initialized already. 🤔

hborla: I agree - accessing the shared property directly is fine here

proposals/0000-shared-storage-property-wrapper.md
```diff
@@ -0,0 +1,403 @@
...
+    set { _value[shared: Hud._value$shared] = newValue }
+  }
+}
+```
+
+Readability is one of the main disadvantages of this approach, as it would require passing the storage around through subscripts. 
+
+### Related Work
+
+The Future Directions sections on both SE-0258 and SE-0293, and the threads discussing this feature, especially this [post](https://www.notion.so/Copy-of-Shared-storage-for-Property-Wrappers-GSoC-Proposal-c6ac301b7e6a401e960a5e2a06adf962), were essential to this proposal.
```
hborla, line 401: This notion link looks wrong. Is this supposed to be linking to https://forums.swift.org/t/future-directions-of-property-wrappers/27934 (or this specific comment: https://forums.swift.org/t/future-directions-of-property-wrappers/27934/11) or something else?

DougGregor:
We'd prefer to iterate the design on the forums, and raise a pull request when the proposal is ready for review.

PR #1397, opened 2021-06-27
[SE-0313] Require nonisolated declarations to involve Sendable types.

DougGregor:
Eliminate a potential source of data races by requiring that a
non-isolated declaration only involve `Sendable` types. Otherwise,
the `Sendable` type could be passed across actor boundaries through
nonisolated declarations.


lattner:
This makes sense to me, it seems like a bugfix.

PR #1409, opened 2021-07-19
Async let second review

DougGregor:


proposals/0317-async-let.md
```diff
@@ -635,56 +640,233 @@ As discussed in the [structured concurrency proposal](0304-structured-concurrenc
...
 
-In initial versions of this proposal, we considered a rule to force an `async let` declaration to be awaited on each control-flow path that the execution of a function might take. This was subsequently relaxed to allow never awaiting an `async let` at all, and implicitly awaiting them at the end of the scope in which the `async let` was declared in.
+In initial versions of this proposal, we considered a rule to force an `async let` declaration to be awaited on each control-flow path that the execution of a function might take. This rule turned out to be too simplistic, because it doesn't account for the places in which implicit suspension points actually occur. For example, consider a function with an `async let` in a loop:
+
+```swift
+func runLoop() async {
+  for e in list {
+    async let a = f(e)
+    guard <condition> else {
+      break // cancels and implicitly awaits the task that produces a
```
kavon, line 652:
```suggestion
      break // cancels and implicitly awaits the task that produces "a"
```

proposals/0317-async-let.md
```diff
@@ -635,56 +640,233 @@ As discussed in the [structured concurrency proposal](0304-structured-concurrenc
...
-func hello(guest name: String) async -> String {
-  async let registered = register(name: name, delayInSeconds: 3)
-  // ... 
-  return "Hello \(name)!"
-  // implicitly cancels the 'registered' child-task
-  // implicitly awaits the 'registered' child-task
+func runException() async {
+  do {
+    async let a = f()
+    try mayFail() // cancels and implicitly awaits the task that produces a
```
kavon, line 666:
```suggestion
    try mayFail() // cancels and implicitly awaits the task that produces "a"
```

proposals/0317-async-let.md
```diff
@@ -635,56 +640,233 @@ As discussed in the [structured concurrency proposal](0304-structured-concurrenc
...
+    ... await a ...
+  } catch {
+    ...
+  }
 }
 ```
 
-Under the current proposal, this function will execute the `registered` task, and before it returns the "Hello …!" it will cancel the still ongoing task `registered`, and await on it. If the task `registered` were to throw an error, that error would be discarded! This may be surprising.
+In the second example, the `do` block will exit when `mayFail()` throws an error, at which point the child task that computes `a` will need to be cancelled and awaited. However, there is no place to write `await a` to make that suspension point implicit. Therefore, requiring every potential suspension point due to an `async let` to be explicitly marked will require an extension to the Swift grammar.
```
kavon, line 674:
```suggestion
In the second example, the `do` block will exit when `mayFail()` throws an error, at which point the child task that computes `a` will need to be cancelled and awaited. However, there is no place to write `await a` to make that suspension point explicit. Therefore, requiring every potential suspension point due to an `async let` to be explicitly marked will require an extension to the Swift grammar.
```

proposals/0317-async-let.md
```diff
@@ -635,56 +640,233 @@ As discussed in the [structured concurrency proposal](0304-structured-concurrenc
...
 ```swift
-func hello(guest name: String) async -> String {
-  async let registered = register(name: name, delayInSeconds: 3) // error: `registered` was not awaited on
-  // ... 
-  return "Hello \(name)!"
+func runLoop() async {
+  for e in list {
+    async let a = f(e)
+    guard <condition> else {
+      await break   // awaits the child task that produces the value a
```
kavon, line 683:
```suggestion
      await break   // awaits the child task that produces the value "a"
```

proposals/0317-async-let.md
```diff
@@ -635,56 +640,233 @@ As discussed in the [structured concurrency proposal](0304-structured-concurrenc
...
+    await
+
+  default:
+    ... code ...
+  }
+  ... code ...
+}
+```
+
+The above is a significant expansion of the grammar: introducing the `await` keyword in front of `break`, `continue`, `throw`, and `fallthrough`; requiring `await` on certain throwing expressions; and adding the freestanding `await`statement. It would also need to be coupled with rules that only require the new `await` when it is semantically meaningful. For example, the additional `await` shouldn't be required if all of the `async let` child tasks have already been explicitly awaited in some other manner, e.g.,
```
kavon, line 752:
```suggestion
The above is a significant expansion of the grammar: introducing the `await` keyword in front of `break`, `continue`, `throw`, and `fallthrough`; requiring `await` on certain throwing expressions; and adding the freestanding `await` statement. It would also need to be coupled with rules that only require the new `await` when it is semantically meaningful. For example, the additional `await` shouldn't be required if all of the `async let` child tasks have already been explicitly awaited in some other manner, e.g.,
```

proposals/0317-async-let.md
```diff
@@ -635,56 +640,233 @@ As discussed in the [structured concurrency proposal](0304-structured-concurrenc
...
+async let veggies = { try await chopVegetables() }
+```
+
+The problem with requiring braces is that it breaks the equivalence between the type of the entity being declared (`veggies` is of type `[Vegetable]`) and the value it is initialized with (which now appears to be `@Sendable () async throws -> [Vegetable]`). This equivalence holds throughout nearly all of the language; the only real exception is the `if let` syntax, which which strips a level of optionality and is often considered a design mistake in Swift. For `async let`, requiring the braces would become particularly awkward if one were defining a value of closure type:
+
+```swift
+async let closure = { { try await getClosure() } }
+```
+
+Requiring braces on the right-hand side of `async let` would be a departure from Swift's existing precedent with `let` declarations. In the cases where one is defining a longer child task, it is reasonable to create and immediately call a closure, which is common practice with `lazy` variables:
```
kavon, line 853:
```suggestion
Requiring braces on the right-hand side of `async let` would be a departure from Swift's existing precedent with `let` declarations. In the cases where one is defining a syntactically larger child task, it is reasonable to create and immediately call a closure, which is common practice with `lazy` variables:
```

kavon (approved):
This looks like a good summary to me. I added a batch of typo-fix suggestions.

ktoso (approved):

ktoso:
Good summaries, thanks!

airspeedswift (approved):

PR #1417, opened 2021-07-30
Mark SE-0303 and SE-0305 as Implemented (Swift 5.6)

tomerd:


tomerd:
@abertelrud lmk when the feature flag is removed and marked as available for 5.6 and we can merge this PR

abertelrud:
> @abertelrud lmk when the feature flag is removed and marked as available for 5.6 and we can merge this PR

Sounds good.  As far as I know there isn't yet a Swift 5.6 that we use for marking it available, so that will need to be added first.

tomerd:
> Sounds good. As far as I know there isn't yet a Swift 5.6 that we use for marking it available, so that will need to be added first.

@abertelrud my understanding from @DougGregor was that main was marked as 5.6 already

PR #1423, opened 2021-08-13
[SE-0304] Update the conformances of TaskPriority

benrimmington:
* Add `Codable` (single-value container by [default][]) to match the [implementation][].
* Add `RawRepresentable` (instead of using the raw-value-enum syntax).

[default]: <https://github.com/apple/swift/blob/392ba002c88346bc8d5cf8e19cb233d17794e844/stdlib/public/core/Codable.swift#L5151-L5186>
[implementation]: <https://github.com/apple/swift/blob/793b492811d37a8ed2b1f816a1146bbfc4a3661f/stdlib/public/Concurrency/Task.swift#L266-L267>

### DIFF
--- a/proposals/0304-structured-concurrency.md
+++ b/proposals/0304-structured-concurrency.md
@@ -621,7 +621,8 @@ However, the Darwin specific names exist as aliases and may be used interchangea
 
 ```swift
 /// Describes the priority of a task.
-struct TaskPriority: UInt8, Comparable, Sendable {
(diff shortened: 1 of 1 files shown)


benrimmington:
Should any of the following be changed before release? The current implementation has:

* an `@_alwaysEmitIntoClient` property for the `medium` priority;
* a `static let` property for the deprecated `default` priority;
* an unused `_downgradeUserInteractive` property.

PR #1425, opened 2021-08-17
[SE-0274] Update status: "Withdrawn"

benrimmington:
SE-0274 is withdrawn, because it was superseded by SE-0285.


benrimmington:
SE-0274 is a [future feature](https://github.com/apple/swift/blob/7a4c36ea2e5e2025cea16078859986efa755555b/include/swift/Basic/Features.def#L92), so I guess this proposal will be implemented in Swift 6?

PR #1450, opened 2021-09-28
[stdlib] Median

benrimmington:
This proposal would add a global `median` function to the standard library.

Swift-evolution thread: [Pitch](https://forums.swift.org/t/median/52279)

### DIFF
--- a/proposals/NNNN-median.md
+++ b/proposals/NNNN-median.md
@@ -1,0 +1,70 @@
+# Median
+
+* Proposal: [SE-NNNN](NNNN-median.md)
+* Author: [Ben Rimmington](https://github.com/benrimmington)
+* Review Manager: TBD
+* Status: **Awaiting review**
+* Implementation: [apple/swift#39335](https://github.com/apple/swift/pull/39335)
+
+## Introduction
+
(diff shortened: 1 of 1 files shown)


tkremenek:
The core team did a preflight discussion of this proposal to determine if this should move to the review stage. That discussion raised some concerns:

* The proposal adds a function that takes precisely three arguments. While the semantics of median is evident in this case (i.e., there always is a "center" value), there was general skepticism that such a function would add much value to the Standard Library.

* A more general API — that takes (say) a variable number of values — might have higher utility, but also begets the question of how to define the median for different types. In a statistical context, the median of an even number of values is the average of the two central values. In many programming contexts, however, it is important that the returned value be one of the original elements. The latter definition can also be applied to collections of non-numerical types. For this to be available on non-numeric types, there would need to be a well-defined answer. For example, what should the result of `["a","b","c","d"].median` be?

* The core team wondered if this function would better fit with a family of numerical or statistical APIs. Such packaging of the API would better capture the use case for median and provide precise semantics for the odd and even number of values cases.

Based on these concerns, the core team does not believe the proposal should move forward with review. Thank you for taking the time to draft it up.

PR #1454, opened 2021-10-02
[SE-0211] Remove `isDefined` on Unicode.Scalar.Properties

Azoy:
This was never implemented is now years old, so remove this portion from the proposal text.

### DIFF
--- a/proposals/0211-unicode-scalar-properties.md
+++ b/proposals/0211-unicode-scalar-properties.md
@@ -176,17 +176,6 @@ extension Unicode.Scalar.Properties {
 }
 ```
 
-We also propose the following Boolean computed property that is generally
-useful, though it does not correspond to a named Unicode property:
-
-```swift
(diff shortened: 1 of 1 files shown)


kylemacomber (approved):

milseman:
What's the protocol here? Do we ask the core team, post to SE, or what?

Azoy:
Apparently the core team put out a statement addressing issues like this here: https://forums.swift.org/t/addressing-unimplemented-evolution-proposals/40322

Specifically for this case:
> Proposals which are partially implemented will have a "shrink to fit" approach applied, where the unimplemented parts will be removed from the proposal. This approach has already been taken with SE-0057.

This is probably worth a post on SE as well just to keep everyone informed.

milseman:
SGTM

glessard:
Public bug: https://bugs.swift.org/browse/SR-14435

PR #1455, opened 2021-10-08
Add generalized keypath-function conversions proposal

Jumhyn:


proposals/NNNN-generalize-keypath-function-conversions.md
```diff
@@ -0,0 +1,118 @@
...
+evil({ kp in { string -> Bool in string[keyPath: kp] } }(\String.isEmpty))
+
+// 'generic'
+evil({ kp in { (string: String) in string[keyPath: kp] } }(\String.isEmpty))
+
+// 'concrete'
+evil({ kp in { $0[keyPath: kp] } }(\String.isEmpty))
+```
+
+Thus, the author opts for a model which maintains the simple "keypath version has the same semantics as the explicit closure version" rule.
```
Jumhyn, line 106: No longer convinced this is suitable...

Jumhyn: Actually, looks like this is even narrower than previously described

Jumhyn:
@rjmccall since the confusion has come up a couple times now, I would love your thoughts on how to phrase/frame this proposal to make it clear that it's *not* a proposal to introduce a conversion to function type *all* `KeyPath`-typed values. Perhaps 'generalization' invites that interpretation... so open to different titles as well

rjmccall:
> @rjmccall since the confusion has come up a couple times now, I would love your thoughts on how to phrase/frame this proposal to make it clear that it's _not_ a proposal to introduce a conversion to function type _all_ `KeyPath`-typed values. Perhaps 'generalization' invites that interpretation... so open to different titles as well

Yeah, I think "generalization" is the problem.  "Subtyping for key path literals as functions"?

proposals/NNNN-keypath-function-subtyping.md
```diff
@@ -0,0 +1,95 @@
...
+// You write this:
+let f: (User) -> String? = \User.email
+
+// The compiler generates something like this:
+let f: (User) -> String? = { kp in { root in root[keyPath: kp] } }(\User.email)
+```
+
+## Source compatibility
+
+This proposal makes previously invalid conversions valid, and should not affect source compatibility. In situations such as:
```
rjmccall, line 74: Unfortunately, these two phrases are contradictory.  Making a previously invalid conversion valid *always* has the potential to affect source compatibility simply because it does make an overload newly viable.  You're right that any previously-viable overload should be preferable *along this dimension*, but overload resolution has to consider multiple dimensions at once, and it's possible that the newly-viable overload is preferable among one or more or those.  As a result, this can introduce ambiguities or even change the interpretation of code.

I think this should be a fairly small problem in practice, but I think you do need to actually do the investigation (in parallel with the review might be okay?), and you *definitely* shouldn't claim in the proposal text that it has no source compatibility implications.

rjmccall: How about this:

```
This proposal allows conversions in some situations that were previously impossible.  This can affect source compatibility because overloaded function calls may gain new viable overload candidates.

In typical scenarios, these new candidates will be strictly worse than previous candidates because the new conversion is strictly less favorable.  For example, consider this code:

<your example>

Without this proposal, Swift will prefer to call the generic function because the conversion necessary for the concrete function is invalid.  With this proposal, Swift will still prefer to call the generic function because the concrete function requires an extra conversion (from `(String) -> Bool` to `(String) -> Bool?`).

However, this is not always true. A newly-viable overload candidate may be disfavored for the key path argument but favored for other reasons.  This should be uncommon, and so we expect this proposal will have a very small impact in practice, but this will need to be demonstrated as part of landing the proposal in a Swift release.
```

Jumhyn: Yeah, you're totally right, this was too aspirational. Updated!

PR #1471, opened 2021-11-02
Update 0328-structural-opaque-result-types.md

glessard:
I read this sentence multiple times before getting it parsed. I _think_ this is clearer. (It's debatable.)


hborla (approved):
Thank you! I agree that this phrasing is more clear 🙂

PR #1483, opened 2021-11-21
[Proposal] Introduce existential `any`

hborla:
Pitch thread: https://forums.swift.org/t/pitch-introduce-existential-any/53520
Implementation: https://github.com/apple/swift/pull/40282


proposals/NNNN-existential-any.md
```diff
@@ -0,0 +1,188 @@
+# [Pitch] Introduce existential `any`
```
benrimmington, line 1:
```suggestion
# Introduce existential `any`
```

proposals/NNNN-existential-any.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Use `Any<P>` instead of `any P`
+
+A common suggestion is to spell existential types with angle brackets on `Any`, e.g. `Any<Hashable>`. However, this syntax is misleading because it appears that `Any` is a generic type, which is confusing to the mental model for 2 reasons:
+
+1. A generic type is something programmers can implement themselves. In reality, existential types are a built-in language feature that would be _very_ difficult to replicate with regular Swift code.
+2. This syntax creates the misconception that the underlying concrete type is a generic argument to `Any` that is preserved statically in the existential type. The `P` in `Any<P>` looks like an implicit type parameter with a conformance requirement, but it's not; the underlying type confoming to `P` is erased at compile-time.
```
benrimmington, line 142: `confoming` → `conforming`

```suggestion
2. This syntax creates the misconception that the underlying concrete type is a generic argument to `Any` that is preserved statically in the existential type. The `P` in `Any<P>` looks like an implicit type parameter with a conformance requirement, but it's not; the underlying type conforming to `P` is erased at compile-time.
```

proposals/NNNN-existential-any.md
```diff
@@ -0,0 +1,281 @@
...
+}
+```
+
+## Source compatibility
+
+Enforcing that existential types use the `any` keyword will require a source change. To ease the migration, I propose to start allowing existential types to be spelled with `any` with the Swift 5.6 compiler, and require existential types to be spelled with `any` under the Swift 6 language mode. The old existential type syntax will continue to be supported under the Swift 5 language mode, and the transition to the new syntax is mechanical, so it can be performed automatically by a migrator.
+
+[SE-0306 Unlock existentials for all protocols](https://github.com/apple/swift-evolution/blob/main/proposals/0309-unlock-existential-types-for-all-protocols.md) enables more code to be written using existential types. To minimize the amount of new code written that will become invalid in Swift 6, I propose requiring `any` immediately for protocols with `Self` and associated type requirements. This introduces an inconsistency for protocols under the Swift 5 language mode, but this inconsistency already exists today (because you cannot use certain protocols as existential types at all), and the syntax difference serves two purposes:
```
filip-sakel, line 203: I believe that’s SE 309 not 306.

hborla: You are totally right, it even says the number right there in the link 🤦🏻‍♀️ fixed, thank you!

PR #1486, opened 2021-11-29
[Proposal ]support lamdba 

MarcSteven:
In most programming language , it both support lambda , but in swift it never make this happen.
In Python, it looks like that:
`my_lamdba = lambda a : a + 20
   print(my_lamdba(20))`


DougGregor:
This is an unrelated commit, not a proposal. Please see https://github.com/apple/swift-evolution/blob/main/process.md for information on the Swift Evolution process.

PR #1488, opened 2021-11-29
proposal for buffer and pointer initialization improvements

glessard:
- as pitched in https://forums.swift.org/t/53168

### DIFF
--- a/proposals/NNNN-buffer-pointer-initialization.md
+++ b/proposals/NNNN-buffer-pointer-initialization.md
@@ -1,0 +1,613 @@
+# Initialization improvements for UnsafePointer and UnsafeBufferPointer family
+
+* Proposal: [SE-NNNN Initialization improvements for UnsafePointer and UnsafeBufferPointer family][proposal]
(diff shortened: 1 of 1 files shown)


glessard:
I wonder if this should be merged with the "partial initialization" pitch ([thread](https://forums.swift.org/t/pitch-buffer-partial-initialization-better-buffer-slices/53795/4)). Most of the extensions in the `Slice` pitch overlap with this, as it reproduces the functions on buffers.

atrick:
> I wonder if this should be merged with the "partial initialization" pitch ([thread](https://forums.swift.org/t/pitch-buffer-partial-initialization-better-buffer-slices/53795/4)). Most of the extensions in the `Slice` pitch overlap with this, as it reproduces the functions on buffers.

FWIW, I actually didn't realize there were two separate initialization API proposals. I've been confusing them with each other all along.

glessard:
Closing this PR in favor of https://github.com/apple/swift-evolution/pull/1561

PR #1489, opened 2021-12-08
[SE-0333] clarifications and rewordings

glessard:
Add paragraphs to explain the conditions under which rebinding is legal/safe.
Various rewordings and doc-comment updates (under detailed design).

### DIFF
--- a/proposals/0333-with-memory-rebound.md
+++ b/proposals/0333-with-memory-rebound.md
@@ -136,10 +136,9 @@ Two types A and B are layout equivalent when they are, for example:
 - identical types;
 - one is a typealias for the other;
 - trivial scalar types with the same size and alignment, including floating-point and integer types;
-- one is a class type, and the other is an `AnyObject` existential;
+- one is a class type, and the other is one of its superclass types (including `AnyObject`);
 - pointer types, such as `UnsafePointer` and `OpaquePointer`;
 - one is a frozen struct with a single stored property, the other is the type of its stored property;
-- one is a frozen enum with a single case, the other is the associated value of the enum's single case;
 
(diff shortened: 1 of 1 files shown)


proposals/0333-with-memory-rebound.md
```diff
@@ -128,15 +114,43 @@ var points = Array<CGPoint>(unsafeUninitializedCapacity: data.count/MemoryLayout
...
+To this we add that, as an alternative, `T` can be a homogeneous aggregate of `Pointee`, or `Pointee` can be a homogeneous aggregate of `T`.
+
+Two types A and B are layout equivalent when they are, for example:
+- identical types;
+- one is a typealias for the other;
+- trivial scalar types with the same size and alignment, including floating-point and integer types;
+- one is a class type, and the other is an `AnyObject` existential;
+- pointer types, such as `UnsafePointer` and `OpaquePointer`;
+- one is a frozen struct with a single stored property, the other is the type of its stored property;
+- one is a frozen enum with a single case, the other is the associated value of the enum's single case;
```
atrick, line 142: I hope people don't think "enum with a single case" means "enum with a single payload case". e.g. `Optional<T>` is not layout equivalent with `T` (unless `T` is a reference). I'm not really sure where single-case enums come up to be honest. So maybe we should just remove  this rule.

atrick (approved):
LGTM. One comment below on the enum-single-case rule.

proposals/0333-with-memory-rebound.md
glessard: I've made single-case enums for specific `Error` implementations, but that's not very interesting.

glessard: Removing that example sounds fine to me.

PR #1492, opened 2021-12-10
[SE-0335] Proposal clarifications based on review feedback.

hborla:
* Strengthen the justification for `any` instead of `Any<P>`.
* Add an example of `any` with protocol compositions to the proposed solution section.


DougGregor (approved):
Clarifications to the proposal under review are fine. I'll go ahead and merge this now.

PR #1495, opened 2021-12-13
Proposal for 'Module aliasing for disambiguation'

elsh:


proposals/NNNN-module-aliasing-for-disambiguation.md
```diff
@@ -3,7 +3,7 @@
 * Proposal: [SE-NNNN](NNNN-module-aliasing-for-disambiguation.md)
 * Authors: [Ellie Shin](https://github.com/elsh)
 * Review Manager: TBD
-* Status: **Implemented**
+* Status: **Implementation WIP**
```
tomerd, line 6: This should be "Awaiting review" at the point

proposals/NNNN-module-aliasing-for-disambiguation.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Introduction
+
+Currently when there are duplicate module names, it fails to compile. The issue has become more widespread as libraries written in Swift have become more increasingly distributed. There is no submodule support in Swift, and introducing structural namespacing is a non-trivial task and would most likely be source breaking. This proposal introduces a way to disambiguate conflicting modules without requiring source breaking changes, by means of aliasing module names.
```
rjmccall, line 13:
```suggestion
Swift does not allow multiple modules in a program to share the same name, and attempts to do so will fail to build. These name collisions can happen in a reasonable program when using multiple packages developed independently from each other. This proposal introduces a way to resolve these conflicts without making major, invasive changes to a package's source by turning a module name in source into an alias to a different, unique name.
```

proposals/NNNN-module-aliasing-for-disambiguation.md
```diff
@@ -0,0 +1,226 @@
...
+```
+
+```
+[Module Game] // swift-game
+
+import Utils  // swift-game
+public func start(level: Utils.Level) { ... }
+```
+
+As with the first scenario, the Utils modules are conflicting, so we need to perform the steps 1-2 above. Then we need to build module Game by applying `-module-alias Utils=GameUtils`, so that the references to Utils in source files of module Game are compiled as GameUtils without making any source changes. The compiler invocation command to build Game then is `swiftc -module-name Game -module-alias Utils=GameUtils ...`. App can then import module Game and any of the renamed Utils as needed.
```
rjmccall, line 79: Is it possible in this scenario to not rename the `Utils` from `swift-draw`?  If this conflict was introduced by updating `swift-game`, I imagine the developers of `App` would prefer to leave the `Utils` from `swift-draw` alone, since otherwise they will need to make a bunch of source changes.

proposals/NNNN-module-aliasing-for-disambiguation.md
```diff
@@ -0,0 +1,226 @@
...
+```
+[Module Game] // swift-game
+
+import Utils  // swift-game
+public func start(level: Utils.Level) { ... }
+```
+
+As with the first scenario, the Utils modules are conflicting, so we need to perform the steps 1-2 above. Then we need to build module Game by applying `-module-alias Utils=GameUtils`, so that the references to Utils in source files of module Game are compiled as GameUtils without making any source changes. The compiler invocation command to build Game then is `swiftc -module-name Game -module-alias Utils=GameUtils ...`. App can then import module Game and any of the renamed Utils as needed. 
+
+As seen above, module aliasing can be done directly via a compiler invocation command, but most users do not interact with the command directly. Thus we plan to provide an easier access via new build configs which can be adopted by any build systems; in particular, we will focus on how it can be adopted via SwiftPM, described in a later section.
```
rjmccall, line 81:
```suggestion
As seen above, module aliasing can be done directly via compiler invocation flags. However, most users use high-level build systems rather than interacting with the command line directly. We will describe how this feature can be adopted in SwiftPM in a later section. 
```

I'd leave the part about "adopted by any build system" out; I don't see anything like that in the proposal, and I don't know how it would work.  You're adding command line flags that any build system can use, and you're adding features to SwiftPM that will translate to these flags.

I'm not sure I agree with this ordering of the presentation, though.  If you really think that the SwiftPM features are the way most users will interact with this, then I think you should describe those as the Proposed Solution, and the frontend flags should be part of the Detailed Design.  Or maybe Proposed Solution first describes it at a high level without talking about either, then talk about the SwiftPM flags, then Detailed Design goes into more detail about both.  That is, the flow would be:

> **Proposed Solution**
>
> We think module aliases solve the problem.  Swift already lets you set the name of a module.  Module aliases let you also tell Swift that a module name is actually an alias for a different module.  Together, these low-level tools allow conflicts to be resolved by giving modules a unique name while using aliases to avoid the need to change any source code.  All of this can be done directly from SwiftPM.
>
> Here is an example: ...
>
> We need to rename these modules when building them.  App will then use them under the renamed names: ...
>
> Here's how to do that in SwiftPM: ...
>
> That example was uncommon.  Here's a more likely one with an indirect dependency: ...
>
> We need to rename one of these modules, both internally and in the modules that use it: ...
>
> Here's how to do that in SwiftPM: ...
>
> **Detailed Design**
>
> Command-line flags
>
> SwiftPM features

proposals/NNNN-module-aliasing-for-disambiguation.md
```diff
@@ -0,0 +1,226 @@
...
+
+* Introducing new import syntax such as `import Utils as GameUtils` has been discussed in forums to improve module disambiguation. The module aliasing infrastructure described in this proposal paves the way towards using such syntax that could allow more explicit (in source code) aliasing.  
+
+* Visibility change to import decl access level (from public to internal) pitched [here](https://forums.swift.org/t/pre-pitch-import-access-control-a-modest-proposal/50087) could help address the extension leaks issues mentioned in **Requirements / Limitations** section. 
+
+* Swift modules that have C target dependencies could, in a limited capacity, be supported by changing visibility to C symbols.
+
+* C++ interop support could potentially allow C++ modules to be aliased besides pure Swift modules.  
+
+* Swift currently does not support nested namespacing / submodules, but it would be an interesting future direction to explore. Such feature would allow Swift users to better organize their code, resolve naming conflicts, and set more fine grained access controls.
```
rjmccall, line 219: Yeah, I don't think you need any of this.  You can just say that submodules would remove some of the need for this, but not all, and that it might also be necessary to make aliases for submodules when they're introduced.

proposals/NNNN-module-aliasing-for-disambiguation.md
elsh: Yes it's possible; if a conflict is resolved by renaming one module, there's no need to alias the other module.

proposals/NNNN-module-aliasing-for-disambiguation.md
elsh: sounds good; will update

proposals/NNNN-module-aliasing-for-disambiguation.md
rjmccall: Okay, makes sense.

proposals/NNNN-module-aliasing-for-disambiguation.md
```diff
@@ -0,0 +1,231 @@
...
+    name: "App",
+    dependencies: [
+     .product(name: "GameProduct", *moduleAliases*: ["Utils": "GameUtils"], package: "swift-game"), 
+     .product(name: "UtilsProduct", *moduleAliases*: ["Utils": "DrawUtils"], package: "swift-draw"), 
+   ])
+ ]
+}
+```
+
+SwiftPM will check validations on `moduleAliases`; for each entry in `moduleAliases`, check if the target indeed opts in for `allowModuleAliasing`, and trigger a build with `-module-alias`. An error will be thrown if conflicting targets do not opt in or no `moduleAliases` are declared. The aliasing option should be supported with a minor version upgrade.
```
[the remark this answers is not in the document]
elsh: Updated with "Only one new name can be given to a conflicting module and it should be unique."

proposals/NNNN-module-aliasing-for-disambiguation.md
```diff
@@ -0,0 +1,231 @@
...
+### Changes to SwiftPM
+
+To allow module aliasing more accessible, we will introduce new build configs which can map to the compiler flags for aliasing described above. Let’s go over how they can be adopted by SwiftPM with the second scenario (copied below). 
+```
+App
+|— Module Game (from package ‘swift-game’)
+|— Module Utils (from package ‘swift-game’)
+|— Module Utils (from package ‘swift-draw’)
+```
+Manifest swift-game: the Utils module needs to opt in to allow module aliasing; we will introduce a new parameter called `allowModuleAliasing`. The Utils target has to meet the requirements described in the **Requirements/Limitations** section below. A package can vend multiple targets which might not meet the requirements, so this new parameter needs to be specified per target. If opted in, SwiftPM will perform validations such as whether the target is a pure Swift module.
```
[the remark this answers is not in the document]
elsh: This is debatable; I agree that requiring an opt-in can be an extra work and very few will do so. It might be better to just run validations and throw an error if the module doesn't meet the requirements. Will update this section accordingly.

PR #1499, opened 2021-12-16
Proposal for statically linking the swift runtime libraries by default on supported platforms

tomerd:


proposals/NNNN-static-link-runtime-libraries-by-default-on-supported-platforms.md
```diff
@@ -207,3 +209,13 @@ In practice, this proposal does not add new features, it only changes default be
 That said, we believe that changing the default will make using Swift on non-Darwin platforms easier, saving time and costs to Swift users on such platforms.
 
 The spelling of the new flag `--disable-static-swift-runtime` is open to alternative ideas, e.g. `--disable-static-swift-runtime-libraries`.
+
+### Policy for system shipping with the Swift runtime libraries
```
tkremenek, line 213: I wonder if this section should come before "Proposed solution".  Essentially, this section lays out the way we should think about static versus dynamic linking.  It's what justifies the static linking approach.

The proposed solution section should also clearly state that the static linking approach could change for a platform if the state of available Swift runtimes changes.  That's part of the proposal, and folks should be clear about the level of guarantees the linking policy is meant to provide now and on the future.  In other words, whether or not a platform uses static linking is an byproduct of the overall runtime policy we want to pursue, not the other way around.

tkremenek (changes requested):

proposals/NNNN-static-link-runtime-libraries-by-default-on-supported-platforms.md
tomerd: thanks @tkremenek how about https://github.com/apple/swift-evolution/pull/1499/commits/072b984882189b0c83bbc82f9472bb03cf84e77b?

tkremenek (approved):

PR #1502, opened 2021-12-17
[SE-0336: Distributed Isolation] Minor additions and rewordings

ktoso:
Minor rewordings or clarifications etc that we could add; notes from https://forums.swift.org/t/se-0336-distributed-actor-isolation/53939


ktoso:
Discussed if all cleanups look good and added a bunch of clarifications as followups

PR #1520, opened 2022-01-20
[SE-0329] Update links and revision history

benrimmington:
The v3 proposal's filename was changed, so there will be some broken links now.


rjmccall:
Thanks!

PR #1523, opened 2022-01-21
[SE-0335] Update the proposal to reflect the acceptance modifications.

hborla:
Incorporate modifications from the acceptance post: https://forums.swift.org/t/accepted-with-modifications-se-0335-introduce-existential-any/54504


DougGregor (approved):
Thanks!

PR #1528, opened 2022-01-27
[SE-0330] Conditionals in Collections: Update for revision

rintaro:
Clarified the design details

### DIFF
--- a/proposals/0330-collection-conditionals.md
+++ b/proposals/0330-collection-conditionals.md
@@ -1,16 +1,16 @@
 # Conditionals in Collections
 
 * Proposal: [SE-0330](0330-collection-conditionals.md)
-* Authors: [John Holdsworth](https://github.com/johnno1962)
+* Authors: [John Holdsworth](https://github.com/johnno1962), [Rintaro Ishizaki](https://github.com/rintaro)
 * Review Manager: [Ted Kremenek](https://github.com/tkremenek)
-* Status: **Returned for revision**
-* Decision notes: [Rationale](https://forums.swift.org/t/se-0330-conditionals-in-collections/53375/22)
-* Implementation: [apple/swift#19347](https://github.com/apple/swift/pull/19347)
+* Status: **Awaiting review**
+* Previous revision: [1](https://github.com/apple/swift-evolution/blob/973b08fd550e63cafb039c3de7e9a85ad3e7e7d3/proposals/0330-collection-conditionals.md)
+* Implementation: [apple/swift#40316](https://github.com/apple/swift/pull/40316)
 * Bugs: [SR-8743](https://bugs.swift.org/browse/SR-8743)
 
 ## Introduction
 
-This is a lightning proposal to extend the existing Swift language slightly to allow `#if` conditional inclusion of elements in array and dictionary literals. For example:
+This is a proposal to extend the existing Swift language to allow `#if` conditional inclusion of elements in array and dictionary literals. For example:
 
 ```swift
 let array = [
@@ -18,7 +18,9 @@ let array = [
 	#if os(Linux)
 	2,
 	#endif
-	3]
+	3
+]
+
 let dictionary = [
 	#if DEBUG
 	"a": 1,
@@ -26,8 +28,10 @@ let dictionary = [
 	"b": 2,
 	#endif
 	#endif
-	"c": 3]
+	"c": 3
+]
 ```
+
 Swift-evolution thread: [Allow conditional inclusion of elements in array/dictionary literals?](https://forums.swift.org/t/allow-conditional-inclusion-of-elements-in-array-dictionary-literals/16171)
 
 ## Motivation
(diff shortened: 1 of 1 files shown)


proposals/0330-collection-conditionals.md
```diff
@@ -36,13 +40,168 @@ The most notable use case for this is conditional inclusion of tests for the Swi
...
+
+```swift
+let literal = [
+#if CONDITION
+  "foo": "bar",
+#endif
+]
+```
+
+You cannot and don't need to write `:` in a block  to emulate a empty dictionally litral  `[:]`.
```
airspeedswift, line 87:
```suggestion
You cannot, and do not need to, write `:` in a block  to emulate an empty dictionary literal  `[:]`.
```

proposals/0330-collection-conditionals.md
```diff
@@ -58,4 +217,6 @@ N/A. This is not an API.
 
 ## Alternatives considered
 
-It was decided to tackle this limited scope for the introduction of conditional syntax first, as specific use cases can be thought of and to be honest this has always seemed like a bit of an omission. Other areas where conditionals could be introduced abound but can be discussed with reference to their own particular subtleties of implementation separately at a later date.
+### Lexer based preprocessing
+
+Currently, including this proposal, processing of conditional compilation directives is based on parse trees. There is an discussion of doing this with Lexer-based preprocessing. Since Lexer based solution has a significant impact on the current syntax tooling (e.g. swift-format), this proposal doens't consider it, and focuses on expanding the current model to collection literals.
```
airspeedswift, line 222:
```suggestion
Currently, processing of conditional compilation directives in Swift is based on parse trees. This proposal continues that approach.

There has been discussion of achieving the goals of this proposal as well as unlocking many other possible uses of `#if` by switching Swift to a Lexer-based preprocessing model. 

A lexer-based solution would have significant impact on the current syntax tooling (e.g. `swift-format`). Most notably, a lexer-based model would require source tooling to have knowledge of the build arguments to be used to perform syntax analysis. Therefore, this proposal does not consider it, and focuses on expanding the current model to collection literals. It also does not close off the possibility of a lexer-based approach from being proposed at a later time, but that should be done as a separate initiative to what is proposed here.
```

proposals/0330-collection-conditionals.md
```diff
@@ -36,13 +40,168 @@ The most notable use case for this is conditional inclusion of tests for the Swi
...
+  value4,
+#endif
+  value5,
+#endif
+]
+```
+
+### Trailing commas
+
+A trailing comma for the last element in conditional compilation diretive blocks is required, regadless of where the directive appears in the literal.
```
airspeedswift, line 169:
```suggestion
A trailing comma for the last element in conditional compilation directive blocks is required, regardless of where the directive appears in the literal. 
```

proposals/0330-collection-conditionals.md
```diff
@@ -36,13 +40,168 @@ The most notable use case for this is conditional inclusion of tests for the Swi
...
+
+let array = [
+  Dog(),
+#if CONDITION_1
+  Cat(),
+#endif
+]
+```
+
+In this example, if `CONDITION_1` evaluates to `true`, this array is inferred `Array<Base>` because `Animal` the best common type between `Dog()` and `Cat()`, otherwise it's `Array<Dog>` because `Dog()` is the only active element.
```
airspeedswift, line 132:
```suggestion
In this example, if `CONDITION_1` evaluates to `true`, this array is inferred `Array<Animal>` because `Animal` the best common type between `Dog()` and `Cat()`, otherwise it's `Array<Dog>` because `Dog()` is the only active element.
```

proposals/0330-collection-conditionals.md
```diff
@@ -36,13 +40,168 @@ The most notable use case for this is conditional inclusion of tests for the Swi
...
+  key: value,
+#else
+  value, // error: expected ':' and a dictionary literal element
+#endif
+]
+```
+
+### Empty dictionary literal
+
+A dictionary literal can be empty depending on the conditional compilation directives. In the following case, if the `CONDITION`evaluates to `false`, this is considered an empty dictionary literal.
```
airspeedswift, line 77:
```suggestion
A dictionary literal can be empty depending on the conditional compilation directives. In the following case, if the `CONDITION` evaluates to `false`, this is considered an empty dictionary literal. 
```

airspeedswift (approved):

proposals/0330-collection-conditionals.md
```diff
@@ -1,33 +1,37 @@
...
+This is a proposal to extend the existing Swift language to allow `#if` conditional inclusion of elements in array and dictionary literals. For example:
 
 ```swift
 let array = [
 	1,
 	#if os(Linux)
 	2,
 	#endif
-	3]
+	3
```
velairi, line 21: I think this can be prettier :)

proposals/0330-collection-conditionals.md
```diff
@@ -36,13 +40,168 @@ The most notable use case for this is conditional inclusion of tests for the Swi
...
+### Grammar changes
+
+This proposal changes `array-literal-items` and `dictionary-literal-items` to accept `conditional-compilation-block` in it.
+
+```diff
+- array-literal-items → array-literal-item ','(opt) | array-literal-item ',' array-literal-items
++ array-literal-items → array-literal-item ','(opt) | array-literal-item ',' array-literal-items | conditional-compilation-block array-literal-items
+
+- dictionary-literal-items → dictionary-literal-item ','(opt) | dictionary-literal-item ',' dictionary-literal-items
++ dictionary-literal-items → dictionary-literal-item ','(opt) | dictionary-literal-item ',' dictionary-literal-items | conditional-compilation-block dictionary-literal-items
```
beccadax, line 56: There don't *have* to be any items after the conditional compilation block, right? So we probably want this to be:

```suggestion
- array-literal-items → array-literal-item ','(opt) | array-literal-item ',' array-literal-items
+ array-literal-items → array-literal-item ','(opt) | array-literal-item ',' array-literal-items | conditional-compilation-block array-literal-items(opt)

- dictionary-literal-items → dictionary-literal-item ','(opt) | dictionary-literal-item ',' dictionary-literal-items
+ dictionary-literal-items → dictionary-literal-item ','(opt) | dictionary-literal-item ',' dictionary-literal-items | conditional-compilation-block dictionary-literal-items(opt)
```

proposals/0330-collection-conditionals.md
```diff
@@ -58,4 +217,10 @@ N/A. This is not an API.
...
+### Lexer based preprocessing
+
+Currently, processing of conditional compilation directives in Swift is based on parse trees. This proposal continues that approach.
+
+There has been discussion of achieving the goals of this proposal as well as unlocking many other possible uses of `#if` by switching Swift to a Lexer-based preprocessing model.
+
+A lexer-based solution would have significant impact on the current syntax tooling (e.g. `swift-format`). Most notably, a lexer-based model would require source tooling to have knowledge of the build arguments to be used to perform syntax analysis. Therefore, this proposal does not consider it, and focuses on expanding the current model to collection literals. It also does not close off the possibility of a lexer-based approach from being proposed at a later time, but that should be done as a separate initiative to what is proposed here.
```
beccadax, line 226: Looking over the review feedback, several people presented a related argument along the lines of “If we switch to a lexer-based model later, this proposal will create technical debt that will need to be cleaned up”. I think it may be a good idea to add a sentence or two discussing this—perhaps simply pointing out that the updated implementation is pretty small so it seems like there won’t be much tech debt to clean up.

PR #1537, opened 2022-02-02
proposal for unaligned loads and stores

glessard:
Implementation PR: https://github.com/apple/swift/pull/41033


rjmccall:
@glessard Is this still awaiting an implementation?

glessard:
The proposal and implementation are ready for review.

glessard:
I could remove excessive references to drafts, I guess.

rjmccall:
Ah, I see the link in the PR summary.  Could you link it from the document, please?  And yeah, if there's any draft cleanup you want to do, please do so.

glessard:
@rjmccall Ready to go on my end; let me know if you need any other changes.

rjmccall:
Thanks.  We'll try to talk about it in the Core Team ASAP and figure out how it's going to be scheduled, but I wanted to make sure you felt it was ready before we started that process.

rjmccall:
We'd like to go ahead and run this.  I'll be review manager.  I think two weeks, running from March 30th until April 12th, is probably appropriate; is that okay with you as the author?

glessard:
Sounds great. Thanks!

rjmccall:
> `loadUnaligned` has a `debugPrecondition` for POD-ness. I'll clarify this in the document and in the doc-comment.

Thank you.

glessard:
Regarding `storeBytes` and `storeBytesUnaligned`: one issue here is that the destination is always, by definition, byte-aligned. At the builtins level, there is also not a different way to spell the two. If we had both functions, they would be identical in release mode, and would otherwise only differ by their set of `debugPrecondition` calls.

rjmccall:
> Regarding `storeBytes` and `storeBytesUnaligned`: one issue here is that the destination is always, by definition, byte-aligned. At the builtins level, there is also not a different way to spell the two. If we had both functions, they would be identical in release mode, and differ only by their set of `debugPrecondition` calls.

I mean, we could add a builtin that does an aligned store.  But in fact we don't need to: since you're requiring the type to be POD, you can simply use the existing initialize or assign builtin.

rjmccall:
Do you want me to put this into review?  I can just make these same comments in the review thread, and we can see what the community thinks.  Worst case, the Core Team asks you to revise the proposal.

glessard:
Yes, go ahead. Thanks!

PR #1545, opened 2022-02-11
[SE-0341] Restrict opaque type parameters in consuming positions.

DougGregor:
Per discussion on the review thread, restrict the use of opaque type
parameters in "consuming" positions. They are not useful here (because
it's hard to create values to pass when you don't have access to the
type), and this follows the reasoning in SE-0328 for the same
restriction. Add a Future Directions section to talk about "flipping"
the generics for opaque types in consuming position.

### DIFF
--- a/proposals/0341-opaque-parameters.md
+++ b/proposals/0341-opaque-parameters.md
@@ -93,14 +93,18 @@ Each instance of `some` within the declaration represents a different implicit g
 
 ## Detailed design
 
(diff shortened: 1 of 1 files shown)


IRISZL:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

PR #1549, opened 2022-02-17
Implicitly Opened Existentials

DougGregor:


proposals/nnnn-implicit-open-existentials.md
```diff
@@ -0,0 +1,379 @@
...
+
+## Introduction
+
+Existential types in Swift allow one to store and reason about a value whose specific type is unknown and may change at runtime. The actual value of that stored value, which we refer to as the existential's *underlying type*, is known only by the set of protocols it conforms to and, potentially its superclass. While existential types are useful for expressing values of dynamic type, they are necessarily restricted because of their dynamic nature. Recent proposals have made [existential types more explicit](https://github.com/apple/swift-evolution/blob/main/proposals/0335-existential-any.md) to help developers understand this dynamic nature, as well as [making existential types more expressive](https://github.com/apple/swift-evolution/blob/main/proposals/0309-unlock-existential-types-for-all-protocols.md). However, a fundamental issue with existential type remains, that once you have an existential type it is *very* hard to then use any generic operations a value of that type. Developers usually encounter this via the error message "protocol 'P' as a type cannot conform to itself":
```
airspeedswift, line 12: > The actual value of that stored value, which we refer to as the existential's *underlying type*

The actual _type_ of that stored value?

airspeedswift, line 12: > a fundamental issue with existential type remains

type**s**

proposals/nnnn-implicit-open-existentials.md
```diff
@@ -0,0 +1,387 @@
...
+To make it easier to move from existentials back to the more strongly-typed generics, we propose to implicitly *open* an existential value when it is passed to a parameter of generic type. In such cases, the generic argument refers to the *underlying* type of the existential value rather than the existential "box". Let's start with a protocol `Costume` that involves `Self` requirements, and write a generic function that checks some property of a costume:
+
+```swift
+protocol Costume {
+  func withBells() -> Self
+  func hasSameAdornments(as other: Self) -> Bool
+}
+
+// Okay: generic function to check whether adding bells changes anything
+func hasBells<C: Costume>(_ costume: Costume) -> Bool {
```
stevapple, line 48: I bet this should be:
```suggestion
func hasBells<C: Costume>(_ costume: C) -> Bool {
```

Still, since SE-0341 is accepted now, does it make sense to simplify the line as:
```suggestion
func hasBells(_ costume: some Costume) -> Bool {
```

proposals/nnnn-implicit-open-existentials.md
```diff
@@ -0,0 +1,387 @@
...
+  
+  return true
+}
+```
+
+This notion of opening an existential into a value of opaque type provides an explicit opening syntax, e.g.,
+
+```swift
+func explicitlyOpen(p: any P) {
+  let openedP: some P = any P  // open the type of 'p' and capture it via the opaque type of 'openedP'
```
stevapple, line 139:
```suggestion
  let openedP: some P = p  // open the type of 'p' and capture it via the opaque type of 'openedP'
```

proposals/nnnn-implicit-open-existentials.md
```diff
@@ -0,0 +1,387 @@
...
+
+func cannotOpen6<T: P>(_ x: T?) { }
+
+func testCannotOpenMultiple(array: [any P], p1: any P, p2: any P, xp: X<any P>, pOpt: (any P)?) {
+  cannotOpen1(array)         // each element in the array can have a different underlying type, so we cannot open
+  cannotOpen2(p1, p2)        // p1 and p2 can have different underlying types, so there is no consistent binding for 'T'
+  cannotOpen3(p1, p2)        // similar to the case above, p1 and p2 have different types, so we cannot open them
+  cannotOpen4(xp)            // cannot open the existential in 'X<any P>' there isn't a specific value there.
+  cannotOpen5(p1, p2.getA()) // cannot open either argument because 'T' is used in both parameters
+  cannotOpen6(pOpt)         // cannot open the existential in '(any P)?' because it might be nil, so there would not be an underlying type
```
stevapple, line 199:
```suggestion
  cannotOpen6(pOpt)          // cannot open the existential in '(any P)?' because it might be nil, so there would not be an underlying type
```

proposals/nnnn-implicit-open-existentials.md
```diff
@@ -0,0 +1,387 @@
...
+
+```swift
+func testDecomposeQ(q: any Q) {
+  let (a, b, c) = decomposeQ(q) // a is any Q, b is any P, c is Any
+}
+```
+
+This is identical to the [covariant erasure of associated types described in SE-0309](https://github.com/apple/swift-evolution/blob/main/proposals/0309-unlock-existential-types-for-all-protocols.md#covariant-erasure-for-associated-types), and the rules specified there apply equally here. We can restate those requirements more generally for an arbitrary generic parameter as:
+
+When binding a generic parameter `T` to an opened existential, `T`, `T` and  `T`-rooted associated types that
```
stevapple, line 246:
```suggestion
When binding a generic parameter `T` to an opened existential, `T`, `T` and `T`-rooted associated types that
```

stevapple (changes requested):

proposals/nnnn-implicit-open-existentials.md
```diff
@@ -0,0 +1,467 @@
...
+
+Our approach is to prevent opening an existential value when the generic type parameter that would bind to its underlying type only has *self-conforming* protocol requirements. A protocol `Q` is *self-conforming* when the corresponding existential type `any Q` conforms to the protocol `Q`. Most protocols are not self-conforming, but there are two exceptions:
+
+* The `Error` protocol is self-conforming, as specified in [SE-0235](https://github.com/apple/swift-evolution/blob/main/proposals/0235-add-result.md#adding-swifterror-self-conformance).
+* `@objc` protocols that contain no `static` requirements are self-conforming.
+
+When all of the protocol requirements on a generic type parameter are self-conforming protocols, an existential type can successfully fulfill the requirements of that generic parameter, so we disable the implicit opening of existential values for that generic type parameter. This maintains the semantics of existing well-formed code that passes existential values to generic functions.
+
+### Subsuming behavior of `type(of:)`
```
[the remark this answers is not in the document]
DougGregor: It shouldn't... I'll go reword things to see if I can make it clearer.

proposals/nnnn-implicit-open-existentials.md
```diff
@@ -0,0 +1,535 @@
...
+
+## Detailed design
+
+Fundamentally, opening an existential means looking into the existential box to find the dynamic type stored within the box, then giving a "name" to that dynamic type. That dynamic type name needs to be captured in a generic parameter somewhere, so it can be reasonable about statically, and the value with that type can be passed along to the generic function being called. The result of such a call might also refer to that dynamic type name, in which case it has to be erased back to an existential type. The After the call, any values described in terms of that dynamic type opened existential type has to be type-erased back to an existential so that the opened type name doesn't escape into the user-visible type system. This both matches the existing language feature (opening an existential value when accessing one of its members) and also prevents this feature from constituting a major extension to the type system itself.
```
calda, line 148: Small typo:

```suggestion
Fundamentally, opening an existential means looking into the existential box to find the dynamic type stored within the box, then giving a "name" to that dynamic type. That dynamic type name needs to be captured in a generic parameter somewhere, so it can be reasoned about statically, and the value with that type can be passed along to the generic function being called. The result of such a call might also refer to that dynamic type name, in which case it has to be erased back to an existential type. The After the call, any values described in terms of that dynamic type opened existential type has to be type-erased back to an existential so that the opened type name doesn't escape into the user-visible type system. This both matches the existing language feature (opening an existential value when accessing one of its members) and also prevents this feature from constituting a major extension to the type system itself.
```

DougGregor: Bah, thanks!

PR #1551, opened 2022-02-18
[SE-0334] bugfix: add essential, but missing operator

glessard:
While touching up the implementation of SE-0334, I realized that I forgot to add a not-equals (!=) operator in the “comparison operators” section. I believe it is an obvious bug fix for the proposal.

The not-equals operator is in the current version of the implementation PR (https://github.com/apple/swift/pull/39639).


glessard:
Thanks @tkremenek!

PR #1558, opened 2022-02-23
[SE-0344] Style and grammatical fixes

stevapple:


proposals/0344-distributed-actor-runtime.md
```diff
@@ -144,80 +143,72 @@ This code snippet showcases what kind of distributed actors one might want to im
...
+  /// the `distributed actor` will gain an automatically synthesized conformance to `Codable` as well.
   associatedtype ActorID: Sendable & Hashable
-  
-  /// The specific type of the invocation encoder that will be created and populated 
+
+  /// The specific type of the invocation encoder that will be created and populated
   /// with details about the invocation when a remote call is about to be made.
   ///
-  /// The populated instance will be passed to the remoteCall from where it can be
+  /// The populated instance will be passed to the remote call from where it can be
```
jckarter: @xedin Was the callout to `remoteCall` intentional here?

jckarter (commented):
LGTM, just one question for the proposal authors.

proposals/0344-distributed-actor-runtime.md
xedin: Yes, remoteCall reference is intentional here.

proposals/0344-distributed-actor-runtime.md
```diff
@@ -1074,37 +1071,37 @@ final struct ClusterSystem: DistributedActorSystem {
...
   }
 }
 ```
 
 Note that `method` property is enough to identify the target of the call, we do not need to carry any extra type information explicitly in the call. The method identifier is sufficient to resolve the target method on the recipient, however in order to support generic distributed methods, we need to carry additional (mangled) type information for any of the generic parameters of this specific method invocation. Thankfully, these are readily provided to us by the Swift runtime, so we'll only need to store and send them over.
 
 > **Note:** An implementation may choose to define any shape of "envelope" (or none at all) that suits its needs. It may choose to transport mangled names of involved types for validation purposes, or choose to not transfer them at all and impose other limitations on the system and its users for the sake of efficiency.
->
-> While advanced implementations may apply compression and other techniques to minimize the overhead of these envelopes - this is a deep topic by itself, and we won't be going in depth on it in this proposal - rest assured though, we have focused on making different kinds of implementations possible with this approach.
+>
```
xedin: Is this considered a trailing whitespace as well? There are 3 more occurrences of this later on.

proposals/0344-distributed-actor-runtime.md
```diff
@@ -895,20 +894,20 @@ public protocol DistributedTargetInvocationEncoder {
   mutating func recordGenericSubstitution<T>(_ type: T.Type) throws
 
   /// Record an argument of `Argument` type in this arguments storage.
-  ///
-  /// Ad-hoc requirement.
+  //
+  // Ad-hoc requirement.
```
xedin, line 894: Why `//` instead of `///` for ad-hoc requirement?

proposals/0344-distributed-actor-runtime.md
```diff
@@ -159,34 +158,30 @@ At the core of everything distributed actors do, is the `DistributedActorSystem`
 Building a solid actor system implementation is not a trivial task, and we only expect a handful of mature implementations to take the stage eventually.
 
 > At the time of writing, we–the proposal authors–have released a work in progress [peer-to-peer cluster actor system implementation](https://www.swift.org/blog/distributed-actors/) that is tracking this evolving language feature. It can be viewed as a reference implementation for the language features and `DistributedActorSystem` protocol discussed in this proposal.
->
+>
```
xedin: Trailing whitespace?

xedin (commented):
Thank you!

proposals/0344-distributed-actor-runtime.md
stevapple: This follows the style of L288-289.
Do you think `///` is better?

xedin: Yeah, I think they all need to be `///`

proposals/0344-distributed-actor-runtime.md
stevapple: 👌 Reverted

stevapple:
@xedin One more thing to consider: Do we want to uniform the case style of titles?

xedin:
@stevapple Yeah, consistency is good.

proposals/0344-distributed-actor-runtime.md
stevapple: I double-checked CommonMark spec and believe this is a valid (and somehow recommended) use case. So I tend to remove the trailing whitespace here.

xedin: Alright then, sounds good!

stevapple:
I’ve got a few questions on this:
1. Is "Distributed Actor" a proper noun that needs to preserve its case?
2. Should we apply upper case to the first letter of every effective word in the title, or only the first word?
3. Any exception on the main title?

xedin:
I believe that "Distributed Actor" is a proper noun. @ktoso what is your preference for 2 & 3?

proposals/0344-distributed-actor-runtime.md
ktoso: They were // because that's a note to the reviews, not actual doc, but either is fine

proposals/0344-distributed-actor-runtime.md
ktoso: That was intentional, thanks for fixing

ktoso:
I try to use "distributed actor" in sentences and "DistributedActor" when we mean the type.

ktoso:
Looks good, thanks for the pr!

ktoso (approved):

proposals/0344-distributed-actor-runtime.md
```diff
@@ -1444,7 +1444,7 @@ As we can see, this thunk is "just" taking care of converting the heterogeneous
 
 The thunk again uses the indirect return, so we can avoid any kind of implicit existential boxing even on those layers. Errors are always returned indirectly, so we do not need to do it explicitly.
 
-#### Recipient: Collecting result/error from the Invocation
+#### Recipient: Collecting result/error from the invocation
```
xedin, line 1447: I think either we have to capitalize `invocation` here or make it `an` invocation instead of `the`.

stevapple: Every title uses the plural form "invocations" now👀

xedin (approved):

PR #1566, opened 2022-03-04
Function Back Deployment

tshortli:


rjmccall:
If this is ready, can you mark it as not a draft?

tshortli:
> If this is ready, can you mark it as not a draft?

There are probably a few things that could use a bit of elaboration after the discussion on the forums (Alternatives Considered) but it is otherwise ready.

tshortli:
I think I've addressed the feedback in the pitch thread now.

PR #1569, opened 2022-03-07
REDACTED

ghost:
The mechanism for ObjC classes to provide generic argument information to Swift was
never implemented.Apple don't be stupid


filip-sakel:
There are actual people running these open-source projects and not some abstract entity, so whatever your issue issue is, please refrain from using offensive language and be respectful.

PR #1571, opened 2022-03-08
Implemented updates

DougGregor:
* SE-0341 is implemented in Swift 5.7
* SE-0302 was removed from Swift 5.6 and is implemented in Swift 5.7


omochi:
@DougGregor Hello.

I noticed that in this patch, the implementation version of `SE-0302` was changed from `5.6` to `5.7`.
Could you explain why that is?

From what I’ve found, it seems this proposal has had a rather complicated history.

It was once listed as implemented in `5.5`, but then rolled back because support for `@unchecked Sendable` was not yet implemented:
https://github.com/swiftlang/swift-evolution/pull/1378

Later, it was marked as implemented in `5.6`:
https://github.com/swiftlang/swift-evolution/pull/1531

It’s also listed as a Swift `5.6` feature in the Swift changelog:
https://github.com/swiftlang/swift/blob/main/CHANGELOG.md#swift-56

So, I’m wondering why it’s now been changed again to `5.7`.
Was there still some part that hadn’t actually been implemented as of `5.6`?

PR #1573, opened 2022-03-09
Adding `-async-top-level` flag to top-level concurrency

etcwilde:
There have been requests to enable the asynchronous context in the
top-level without needing to explicitly `await`. I've added
`-async-top-level` to the proposal to enable this behavior.


etcwilde:
Closing in favour of: https://github.com/apple/swift-evolution/pull/1574

PR #1574, opened 2022-03-09
Add paragraph for  `-warn-concurrency` to 0343

etcwilde:
There was interest in discussions on having the concurrency protections
for top-level variables without an await.


compnerd (approved):

DougGregor (approved):
Thanks!

PR #1584, opened 2022-03-21
[stdlib] StaticBigInt

benrimmington:


benrimmington:
Can this proposal be considered for Swift 5.7, or should it wait for `_const` compile-time constant values?

proposals/NNNN-staticbigint.md
```diff
@@ -0,0 +1,133 @@
...
+// Collection APIs
+extension StaticBigInt {
+    public struct Words: RandomAccessCollection, Sendable {
+        public typealias Element = UInt
+        public typealias Index = Int
+    }
+}
+```
+
+## Alternatives considered
```
benrimmington, line 99: @xwu I neglected to incorporate your feedback into the proposal. Would the following summary be acceptable?

```suggestion
## Alternatives considered

Xiaodi Wu [suggested](https://forums.swift.org/t/staticbigint/54545/23) that a different naming scheme and API design be chosen to accommodate other similar types, such as IEEE 754 interchange formats.
```

xwu: Fantastic.

rjmccall:
We haven't quite fit this into the review schedule yet, but I'll be acting as the review manager, and I need to do the review-manager review of the pitch thread first anyway.

As for whether we can get this into 5.7 (assuming it's accepted), I'm not sure.  It's certainly possible, but I don't want to promise anything.

rjmccall:
Things from the review thread that I think ought to be addressed in the proposal:
- Please discuss in Alternatives Considered about how it's an intentional decision that the model is the mathematical integers and *not* the textual representation of literals in source.  For example, we have explicitly decided not to distinguish literals written in different bases, with different leading zeroes, and so on.  This seems like an obviously justifiable decision to me, but you should take the time to justify it.
- Please discuss in Alternatives Considered the possibility of unifying this with the handling of floating-point literals.  Again, I think it shouldn't be difficult to justify this — there's a lot of expressive complexity for floating-point literals that would be an unnecessary burden on integer types — but it does need to be justified in the document.
- Please discuss somewhere in the document (briefly is fine) why this type cannot conform to any of the numeric protocols.
- Please discuss in Alternatives Considered the prior designs that exposed the words as arrays.
- Please discuss in Future Directions the possibility of diagnosing overflow in literals using static evaluation, if the language ever provides that.

proposals/NNNN-staticbigint.md
```diff
@@ -0,0 +1,136 @@
...
+    public var words: Words { get }
+}
+
+// Collection APIs
+extension StaticBigInt {
+    public struct Words: RandomAccessCollection, Sendable {
+        public typealias Element = UInt
+        public typealias Index = Int
+    }
+}
```
rjmccall, line 97: Should there be a way to construct an arbitrary `ExpressibleByIntegerLiteral` type from a `StaticBigInt`?

proposals/NNNN-staticbigint.md
```diff
@@ -0,0 +1,136 @@
...
+```
+
+## Alternatives considered
+
+Xiaodi Wu [suggested](https://forums.swift.org/t/staticbigint/54545/23) that a different naming scheme and API design be chosen to accommodate other similar types, such as IEEE 754 interchange formats.
+
+A *mutable* `BigInt: SignedInteger` type could (eventually) be implemented in the standard library.
```
rjmccall, line 104: This is a future direction.

proposals/NNNN-staticbigint.md
```diff
@@ -0,0 +1,136 @@
...
+
+// Collection APIs
+extension StaticBigInt {
+    public struct Words: RandomAccessCollection, Sendable {
+        public typealias Element = UInt
+        public typealias Index = Int
+    }
+}
+```
+
```
rjmccall, line 99: Please include the standard ABI Stability section.  This feature adds to the ABI of the standard library (and therefore by default will not back-deploy), and that should be explicit in the proposal.

You should probably also mention that, because the integer literal type has to be selected statically as the associated type, there is currently no way to conditionally use a different integer literal type depending on the execution environment.  Types will not be able to adopt this and use the most flexible possible literal type dynamically available.

benrimmington:
@rjmccall I still need to update the implementation.

> This seems like an obviously justifiable decision to me, but you should take the time to justify it.

I can try to justify the design of `Builtin.IntLiteral`, but the honest answer is that I'm not offering to re-design the existing ABI.

> Would it be more useful if there was a way to get the (infinitely-sign-extended) value of an arbitrary word of the representation?

I've removed the `words` collection, and added `indices` and `subscript(_:)` APIs. This is similar to the design of SIMD types.

Can the infinite sign extension be implemented by the subscript, rather than by the `Builtin.wordAtIndex` function?

> Should there be a way to construct an arbitrary `ExpressibleByIntegerLiteral` type from a `StaticBigInt`?

I don't think this is possible, because a [default implementation][] needs to use `Self`-typed literals, which would trigger infinite recursion.

The proposal now has an `init(exactly:)` example, which shows how third-party types could become convertible from a `StaticBigInt`.

[default implementation]: <https://github.com/apple/swift/blob/f045f14dcdd144292cf458d302e53f4486fbf004/stdlib/public/core/Integers.swift#L3165-L3180>

rjmccall:
Thanks for these changes, this is looking a lot better.

The biggest problem I see now is that Alternatives Considered needs to actually argue against the alternatives that were considered in favor of what's being proposed.  Just citing the pitch thread isn't really enough.

I want to make it clear that I was just asking a question about infinite extension, not specifically requesting the change.  It seems like it makes using `StaticBigInt` easier when building a fixed-width type, and it dodges the question about how many words the literal has.  However, those things together mean you can't have a `Collection` of words, and if you think it would be better to have a `Collection`, then by all means, let's stick with that.  Ultimately, this is your proposal, and you should feel confident to defend the decisions you've made.

I'm happy to write the Alternatives Considered section about different literals, since that's really my design (among others) being reflected here.  But you should write the response to the `words` collection (if you do want to make that change) and the response to the pitch-thread commentary about interchange types.

Here's suggested text for the literals item, combining your first two bullet points:

> Modeling the original source text instead of a mathematical value would allow this type to support a wide range of use cases, such as fractional values, decimal values, and other things such as arbitrary binary strings expressed in hexadecimal.  However, it is not a goal of Swift's integer literals design to support these use cases.  Supporting them would burden integer types with significant code size, dynamic performance, and complexity overheads.  For example, either the emitted code would need to contain both the original source text and a more optimized representation used by ordinary integer types, or ordinary integer types would need to fall back on parsing numeric values from source at runtime.
>
> Along similar lines, it is intentional that `StaticBigInt` cannot represent fractional values.  Integer types should not be constructible with fractional literals, and allowing that simply adds unnecessary costs and introduces a new way for construction to fail.  It is still a language goal for Swift to someday support dynamically flexible floating-point literals the way it does for integer literals, but that is a separable project from introducing `StaticBigInt`.

> Can the infinite sign extension be implemented by the subscript, rather than by the `Builtin.wordAtIndex` function?

Either way is fine by me.

> > Should there be a way to construct an arbitrary `ExpressibleByIntegerLiteral` type from a `StaticBigInt`?
> 
> I don't think this is possible, because a [default implementation](https://github.com/apple/swift/blob/f045f14dcdd144292cf458d302e53f4486fbf004/stdlib/public/core/Integers.swift#L3165-L3180) needs to use `Self`-typed literals, which would trigger infinite recursion.

Does the following not work?

```swift
extension ExpressibleByIntegerLiteral {
  init(dynamicIntegerLiteral value: StaticBigInt) {
    self.init(integerLiteral: IntegerLiteralType(_builtinIntegerLiteral: value.value))
  }
}
```

benrimmington:
> I want to make it clear that I was just asking a question about infinite extension, not specifically requesting the change.

Understood. I added it to the proposal because I like the idea. I wasn't sure of an "infinite" collection, so I chose this simpler design.

> I'm happy to write the Alternatives Considered section about different literals, since that's really my design (among others) being reflected here.

Thank you. I feel guilty about you rewriting so much of the proposal.

> But you should write the response to the `words` collection (if you do want to make that change) and the response to the pitch-thread commentary about interchange types.

I'll update these alternatives. I think a `words` collection can be derived from `x.indices.lazy.map { x[$0] }` if needed. A wrapper type could provide a choice of `UInt8`, `UInt16`, `UInt32`, `UInt64`, or `UInt` elements.

> Does the following not work?

An `init?(exactly: StaticBigInt)` and `init(truncatingIfNeeded: StaticBigInt)` on fixed-width integers may be more versatile (than a trapping initializer).

rjmccall:
> > Does the following not work?
> 
> An `init?(exactly: StaticBigInt)` and `init(truncatingIfNeeded: StaticBigInt)` on fixed-width integers may be more versatile (than a trapping initializer).

But those aren't possible to write in general, because arbitrary literal-constructible types don't offer those operations.  This `init` would simply be a way to invoke the standard literal-initialization path for the given literal on a dynamically-chosen literal-constructible type, which is otherwise unimplementable.

benrimmington:
Your `init(dynamicIntegerLiteral:)` suggestion works, but should the argument label be omitted?

Other conversions described by `BinaryInteger` [documentation][] would require `FixedWidthInteger`.

```swift
extension FixedWidthInteger {

  public init?(exactly source: StaticBigInt) {
    if Self.isSigned {
      guard source.bitWidth <= Self.bitWidth else { return nil }
    } else {
      guard source.bitWidth <= Self.bitWidth + 1 else { return nil }
      guard source.signum() >= 0 else { return nil }
    }
    self.init(source)
  }

  public init(clamping source: StaticBigInt) {
    self = Self(exactly: source) ?? (source.signum() < 0 ? Self.min : Self.max)
  }

  public init(truncatingIfNeeded source: StaticBigInt) {
    self = Self(_truncatingBits: source[0])
    if Self.bitWidth > UInt.bitWidth {
      let endIndex = (Self.bitWidth + UInt.bitWidth - 1) / UInt.bitWidth
      for index in 1..<endIndex {
        self |= Self(_truncatingBits: source[index]) &<< (index * UInt.bitWidth)
      }
    }
  }
}
```

[documentation]: <https://developer.apple.com/documentation/swift/binaryinteger>

rjmccall:
> Actually, it doesn't work as expected, because dynamic overflow results in a zero-initialized value, without a runtime error. Related issue: [Negative integer literal converted to generic unsigned integer type without warning](https://forums.swift.org/t/negative-integer-literal-converted-to-generic-unsigned-integer-type-without-warning/55883/2).

Hmm.  Well, we should definitely fix whatever bug this is, but if we're seeing it when creating literals generically, then it makes sense that we see the same behavior here.

benrimmington:
I've updated the design.

- The subscript now has a generic return type, which can be any fixed-width unsigned integer (from the standard library).

- The `indices` property is removed. *Alternatives considered* mentions the need for a replacement API.

PR #1610, opened 2022-04-06
fix formatting in the ToC

scott-vsi:


rjmccall:
Change was made upstream, but thanks for the PR

PR #1611, opened 2022-04-06
SE-0347: Add clarifying example and supporting text to Introduction section

xedin:


xedin:
@xwu Does it look good to you? (Cannot add you as a reviewer to this PR).

xwu:
I like it. On line 32, do you mean that it's possible to infer the default type from the _second_ argument?

xedin:
Ah yes, thank you!

xedin:
Fixed.

PR #1613, opened 2022-04-07
SE-0356: Swift Snippets

bitjammer:
I propose a convention for writing and referencing single-file
examples in Swift packages.


tomerd:
@bitjammer is this ready to kick off the review, or are there any further changes you would like to make before we do so?

bitjammer:
It’s ready, thank you!

PR #1614, opened 2022-04-08
[SE-nnnn] Regex syntax and run-time construction

milseman:


proposals/nnnn-regex-syntax-run-time-construction.md
```diff
@@ -0,0 +1,1052 @@
...
+
+  public var endIndex: Int { get }
+
+  public var count: Int { get }
+
+  public func index(after i: Int) -> Int
+
+  public func index(before i: Int) -> Int
+
+  public subscript(position: Int) -> Element
```
rxwei, line 140: Should we add accessors to make it clear it's readable but not writable
```suggestion
  public subscript(position: Int) -> Element { get }
```

proposals/nnnn-regex-syntax-run-time-construction.md
```diff
@@ -0,0 +1,1052 @@
...
+extension AnyRegexOutput: RandomAccessCollection {
+  public struct Element {
+    /// The range over which a value was captured. `nil` for no-capture.
+    public var range: Range<String.Index>?
+
+    /// The slice of the input over which a value was captured. `nil` for no-capture.
+    public var substring: Substring?
+
+    /// The captured value. `nil` for no-capture.
+    public var value: Any?
```
rxwei, line 125: Suggest adding accessors.
```suggestion
    /// The range over which a value was captured. `nil` for no-capture.
    public var range: Range<String.Index>? { get }

    /// The slice of the input over which a value was captured. `nil` for no-capture.
    public var substring: Substring? { get }

    /// The captured value. `nil` for no-capture.
    public var value: Any?  { get }
```

proposals/nnnn-regex-syntax-run-time-construction.md
```diff
@@ -0,0 +1,1062 @@
...
+}
+```
+
+Finally, we propose API for creating a regex containing literal string content. This produces an equivalent regex to a string literal embededd in the result builder DSL. As this is much less common than run-time compilation or an embedded literal in the DSL, it has an explicit argument label.
+
+```swift
+extension Regex {
+  /// Produces a regex that matches `literal` exactly, as though every
+  /// metacharacter in it was escaped.
+  public init(quoting literal: String)
```
rxwei, line 203: I don't think we should call this parameter "literal", as it may very likely not be a literal, e.g. `Regex(quoting: str1 + str2)`.

milseman: It's more like it's quoting it literally, making it into a literal from the regex's perspective. It's more than just `\Q\(str)\E` because it will effectively escape any interior `\E`. Not to be confused with a Swift language literal, which includes strings and... also regexes.

Seemed better than `string` but I'm open to alternatives

rxwei: Feels more like `verbatim`

milseman: SGTM

PR #1618, opened 2022-04-11
Amend SE-0329 to add Clock.Duration

lorentey:
As [discussed on the forum](https://forums.swift.org/t/pitch-primary-associated-types-in-the-standard-library/56426/39), we'll likely want to use `Duration` as the primary associated type on `protocol Clock`; however, that protocol currently only has `Instant`.

SE-0329 has not yet shipped in an ABI stable release, so we still have the opportunity to address this.

This PR adds `Duration` to `Clock`, constraining `Instant.Duration` to the same type. This setup is reminiscent of `Sequence.Element` vs. `Sequence.Iterator.Element`.


phausler (approved):

rjmccall (approved):
Approved by the Core Team.

PR #1622, opened 2022-04-14
[Proposal] Build-Time Constant Value Attribute

artemcm:
Proposal for a compile-time constant value attribute.

### DIFF
--- a/proposals/0358-build-time-constant-values.md
+++ b/proposals/0358-build-time-constant-values.md
@@ -25,7 +25,7 @@ The first step towards building out support for compile-time constructs in Swift
 
 The Swift compiler will recognize declarations of properties, local variables, and function parameters declared with a `@const` attribute as having an additional requirement to be known at compile-time. If a `@const` property or variable is initialized with a runtime value, the compiler will emit an error. Similarly, if a runtime value is passed as an argument to a `@const` function parameter, the compiler will emit an error. Aside from participating in name mangling, the attribute has no runtime effect.
 
-For example, an `@const` property can provide the compiler and relevant tooling build-time knowledge of a type-specific value:
+For example, a `@const` property can provide the compiler and relevant tooling build-time knowledge of a type-specific value:
 
 ```swift
 struct DatabaseParams {
@@ -34,17 +34,17 @@ struct DatabaseParams {
 }
 ```
 
-An `@const` parameter provides a guarantee of a build-time-known argument being specified at all function call-sites, allowing future APIs to enforce invariants and provide build-time correctness guarantees:
+A `@const` parameter provides a guarantee of a build-time-known argument being specified at all function call-sites, allowing future APIs to enforce invariants and provide build-time correctness guarantees:
 
 ```swift
 func acceptingURLString(@const _ url: String)
 ```
 
-And an `@const static var` protocol property requirement allows protocol authors to enforce get the benefits of build-time known properties for all conforming types:
+And a `@const static let` protocol property requirement allows protocol authors to enforce get the benefits of build-time known properties for all conforming types:
 
 ```swift
 protocol DatabaseSerializableWithKey {
-    @const static var key: String
+    @const static let key: String
 }
 ```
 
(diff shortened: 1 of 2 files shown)


proposals/0358-build-time-constant-values.md
```diff
@@ -0,0 +1,327 @@
...
+
+In order to allow such implementation in the future, this proposal makes the *value* of `public` `@const` values/properties a part of a module's ABI. That is, a resilient library that vends `@const let x = 11` changing the value of `x` is considered an ABI break. This treatment allows `public` `@const` data to exist in a single read-only location shared by all library clients, without each client having to copy the value or being concerned with possible inconsistency in behavior across library versions.
+
+## Future Directions
+
+### Constant-propogation
+Allow default-initialization of `@const` properties using other `@const` values and allow passing `@const` values to `@const` parameters. The [Future Inference/Propagation Rules](#Future Inference/Propagation Rules) section discusses a direction for enabling inference of the attribute on values. This is a necessary next building-block to generalizing the use of compile-time values.
```
benrimmington, line 299: Link to section.

```suggestion
Allow default-initialization of `@const` properties using other `@const` values and allow passing `@const` values to `@const` parameters. The [Future Inference/Propagation Rules](#future-inferencepropagation-rules) section discusses a direction for enabling inference of the attribute on values. This is a necessary next building-block to generalizing the use of compile-time values.
```

proposals/0358-build-time-constant-values.md
```diff
@@ -0,0 +1,327 @@
...
+
+```swift
+func foo(@const i: Int) {
+     @const let j = i
+}
+```
+
+### Toolchain support for extracting compile-time values at build time.
+The current proposal covers an attribute that allows clients to build an API surface that is capable of carrying semantic build-time information that may be very useful to build-time tooling, such as [SwiftPM plugins](https://github.com/apple/swift-evolution/blob/main/proposals/0303-swiftpm-extensible-build-tools.md). The next step towards this goal would include toolchain support for tooling that extracts such information in a client-agnostic fashion so that it can be adopted equally by use-cases like the manifest example in [Facilitate Compile-time Extraction of Values](#Facilitate Compile-time Extraction of Values) and others.
```
benrimmington, line 308: Link to section.

```suggestion
The current proposal covers an attribute that allows clients to build an API surface that is capable of carrying semantic build-time information that may be very useful to build-time tooling, such as [SwiftPM plugins](https://github.com/apple/swift-evolution/blob/main/proposals/0303-swiftpm-extensible-build-tools.md). The next step towards this goal would include toolchain support for tooling that extracts such information in a client-agnostic fashion so that it can be adopted equally by use-cases like the manifest example in [Facilitate Compile-time Extraction of Values](#facilitate-compile-time-extraction-of-values) and others.
```

proposals/0358-build-time-constant-values.md
```diff
@@ -0,0 +1,327 @@
...
+   #const_assert(input <= 0, "'foo()' expects a positive input")
+}
+```
+Which would require that it is possible to evaluate the `input <= 0` expression at compile-time, which would also require that certain functions can be evaluated at compile-time if their arguments are known at compile-time. For example:
+
+
+```swift
+func <=(@const lhs: Int, @const rhs: Int) -> Bool
+```
+Inference on which functions can or cannot be evaluated at compile time will be defined in a future proposal and can follow similar ideas to those described in [Future Inference/Propagation Rules](#Future Inference/Propagation Rules) section.
```
benrimmington, line 324: Link to section.

```suggestion
Inference on which functions can or cannot be evaluated at compile time will be defined in a future proposal and can follow similar ideas to those described in [Future Inference/Propagation Rules](#future-inferencepropagation-rules) section.
```

artemcm:
Thanks @benrimmington !

benrimmington:
@artemcm There are more suggestions above (see 8 hidden conversations).

DougGregor:
This is in good shape to run. I'll manage it.

PR #1624, opened 2022-04-19
Rename 0110 distinguish single tuple arg

jsoref:
Per https://github.com/apple/swift/pull/42438#discussion_r853288871


CodaFi:
⛵

benrimmington:
Please don't rename proposal files, because it may end up breaking existing links.

(In this case, those links are probably using the master branch; but if that branch is ever deleted then GitHub will now fail to redirect to the main branch's proposal.)

PR #1628, opened 2022-04-20
[SE-0352] Address review feedback

DougGregor:
Add discussion about type erasure losing constraints and the new requirement to introduce an explicit `as` coercion when the upper bound loses information.

### DIFF
--- a/proposals/0352-implicit-open-existentials.md
+++ b/proposals/0352-implicit-open-existentials.md
@@ -14,6 +14,7 @@
    * [Detailed design](#detailed-design)
      * [When can we open an existential?](#when-can-we-open-an-existential)
      * [Type-erasing resulting values](#type-erasing-resulting-values)
+     * ["Losing" constraints when type-erasing resulting values](#losing-constraints-when-type-erasing-resulting-values)
      * [Contravariant erasure for parameters of function type](#contravariant-erasure-for-parameters-of-function-type)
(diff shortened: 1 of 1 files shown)


proposals/0352-implicit-open-existentials.md
```diff
@@ -289,6 +290,76 @@ func covariantReturns(q: any Q){
...
+
+With the less-specific upper bound (`any P`), the call `f(b)` would choose the first overload that returns an `Int`. With the more-specific upper bound (`any P` where `A` is known to be `Int`), the call `f(b)` would choose the second overload that returns a `Double`. 
+
+Due to overloading, the source-compatibility impacts of improving the upper bound cannot be completely eliminated without (for example) holding the upper bound constant until a new major language version. However, we propose to mitigate the effects by requiring a specific type coercion on any call where the upper bound is unable to express some requirements due to limitations on existentials. Specifically, the call `getBFromQ(q)` would need to be written as:
+
+```swift
+getBFromQ(q) as any P
+```
+
+This way, if the upper bound changes due to increased expressiveness of existential types in the language, the overall express will still produce a value of the same type---`any P`---as it always has. A developer would be free to remove the `as any P` at the point where Swift can fully capture all of the information known about the type in an existential.
```
xedin, line 349: express -> expression

PR #1631, opened 2022-04-22
Add the Unicode for String Processing proposal

natecook1000:
This proposal is another component in the larger regex-powered string processing project.

### DIFF
--- a/proposals/0000-unicode-for-string-processing.md
+++ b/proposals/0000-unicode-for-string-processing.md
@@ -1,0 +1,873 @@
+# Unicode for String Processing
+
+* Proposal: [SE-NNNN](NNNN-unicode-for-string-processing.md)
+* Authors: [Nate Cook](https://github.com/natecook1000), [Alejandro Alonso](https://github.com/Azoy)
+* Review Manager: TBD
+* Status: **Draft**
+* Implementation: [apple/swift-experimental-string-processing][repo]  
+
+## Introduction
+
+This proposal describes `Regex`'s rich Unicode support during regex matching, along with the character classes and options that define and modify that behavior.
+
+This proposal is one component of a larger [regex-powered string processing initiative](https://github.com/apple/swift-evolution/blob/main/proposals/0350-regex-type-overview.md). For the status of each proposal, [see this document](https://github.com/apple/swift-experimental-string-processing/blob/main/Documentation/Evolution/ProposalOverview.md) — discussion of other facets of the overall regex design is out of scope of this proposal and better discussed in the most relevant review.
+
+## Motivation
+
+Swift's `String` type provides, by default, a view of `Character`s or [extended grapheme clusters][graphemes] whose comparison honors [Unicode canonical equivalence][canoneq]. Each character in a string can be composed of one or more Unicode scalar values, while still being treated as a single unit, equivalent to other ways of formulating the equivalent character:
+
+```swift
+let str = "Cafe\u{301}" // "Café"
+str == "Café"           // true
+str.dropLast()          // "Caf"
+str.last == "é"         // true (precomposed e with acute accent)
+str.last == "e\u{301}"  // true (e followed by composing acute accent)
+```
+
+This default view is fairly novel. Most languages that support Unicode strings generally operate at the Unicode scalar level, and don't provide the same affordance for operating on a string as a collection of grapheme clusters. In Python, for example, Unicode strings report their length as the number of scalar values, and don't use canonical equivalence in comparisons:
+
+```python
+cafe = u"Cafe\u0301"
+len(cafe)                     # 5
+cafe == u"Café"               # False
+```
+
+Existing regex engines follow this same model of operating at the Unicode scalar level. To match canonically equivalent characters, or have equivalent behavior between equivalent strings, you must normalize your string and regex to the same canonical format.
+
+```python
+# Matches a four-element string
+re.match(u"^.{4}$", cafe)     # None
+# Matches a string ending with 'é'
+re.match(u".+é$", cafe)       # None
+
+cafeComp = unicodedata.normalize("NFC", cafe)
+re.match(u"^.{4}$", cafeComp) # <re.Match object...>
+re.match(u".+é$", cafeComp)   # <re.Match object...>
+```
+
+With Swift's string model, this behavior would surprising and undesirable — Swift's default regex semantics must match the semantics of a `String`.
+
+<details><summary>Other engines</summary>
+
+Other regex engines match character classes (such as `\w` or `.`) at the Unicode scalar value level, or even the code unit level, instead of recognizing grapheme clusters as characters. When matching the `.` character class, other languages will only match the first part of an `"e\u{301}"` grapheme cluster. Some languages, like Perl, Ruby, and Java, support an additional `\X` metacharacter, which explicitly represents a single grapheme cluster.
+
+| Matching  `"Cafe\u{301}"` | Pattern: `^Caf.` | Remaining | Pattern:  `^Caf\X` | Remaining |
+|---|---|---|---|---|
+| C#, Rust, Go, Python | `"Cafe"` | `"´"` | n/a | n/a |
+| NSString, Java, Ruby, Perl | `"Cafe"` | `"´"` | `"Café"` | `""` |
+
+Other than Java's `CANON_EQ` option, the vast majority of other languages and engines are not capable of comparing with canonical equivalence.
+
+</details>
+
+## Proposed solution
+
+In a regex's simplest form, without metacharacters or special features, matching behaves like a test for equality. A string always matches a regex that simply contains the same characters.
+
+```swift
+let str = "Cafe\u{301}"     // "Café"
+str.contains(/Café/)        // true
+```
+
+From that point, small changes continue to comport with the element counting and comparison expectations set by `String`:
+
+```swift
+str.contains(/Caf./)        // true
+str.contains(/.+é/)         // true
+str.contains(/.+e\u{301}/)  // true
+str.contains(/\w+é/)        // true
+```
+
+
+For compatibility with other regex engines and the flexibility to match at both `Character` and Unicode scalar level, you can switch between matching levels for an entire regex or within select portions. This powerful capability provides the expected default behavior when working with strings, while allowing you to drop down for Unicode scalar-specific matching.
+
+By default, literal characters and Unicode scalar values (e.g. `\u{301}`) are coalesced into characters in the same way as a normal string, as shown above. Metacharacters, like `.` and `\w`, and custom character classes each match a single element at the current matching level.
+
+For example, these matches fail, because by the time the parser encounters the "`\u{301}`" Unicode scalar literal, the full `"é"` character has been matched:
+
+```swift
+str.contains(/Caf.\u{301})    // false - `.` matches "é" character
+str.contains(/Caf\w\u{301})   // false - `\w` matches "é" character
+str.contains(/.+\u{301})      // false - `.+` matches each character
+```
+
+Alternatively, we can drop down to use Unicode scalar semantics if we want to match specific Unicode sequences. For example, these regexes matches an `"e"` followed by any modifier with the specified parameters:
+
+```swift
+str.contains(/e[\u{300}-\u{314}]/.matchingSemantics(.unicodeScalar))
+// true - matches an "e" followed by a Unicode scalar in the range U+0300 - U+0314
+str.contains(/e\p{Nonspacing Mark}/.matchingSemantics(.unicodeScalar))
+// true - matches an "e" followed by a Unicode scalar with general category "Nonspacing Mark"
+```
+
+Matching in Unicode scalar mode is analogous to comparing against a string's `UnicodeScalarView` — individual Unicode scalars are matched without combining them into characters or testing for canonical equivalence.
+
+```swift
+str.contains(/Café/.matchingSemantics(.unicodeScalar))
+// false - "e\u{301}" doesn't match with /é/
+str.contains(/Cafe\u{301}/.matchingSemantics(.unicodeScalar))
+// true - "e\u{301}" matches with /e\u{301}/
+```
+
+Swift's `Regex` follows the level 2 guidelines for Unicode support in regular expressions described in [Unicode Technical Standard #18][uts18], with support for Unicode character classes, canonical equivalence, grapheme cluster matching semantics, and level 2 word boundaries enabled by default. In addition to selecting the matching semantics, `Regex` provides options for selecting different matching behaviors, such as ASCII character classes or Unicode scalar semantics, which corresponds more closely with other regex engines.
+
+## Detailed design
+
+First, we'll discuss the options that let you control a regex's behavior, and then explore the character classes that define the your pattern.
+
+### Options
+
+Options can be enabled and disabled in two different ways: as part of [regex internal syntax][internals], or applied as methods when declaring a `Regex`. For example, both of these `Regex`es are declared with case insensitivity:
+
+```swift
+let regex1 = /(?i)banana/
+let regex2 = Regex {
+    "banana"
+}.ignoresCase()`
+```
+
+Note that the `ignoresCase()` is available on any type conforming to `RegexComponent`, which means that you can always use the more readable option-setting interface in conjunction with regex literals or run-time compiled `Regex`es:
+
+```swift
+let regex3 = /banana/.ignoresCase()
+```
+
+Calling an option-setting method like `ignoresCase(_:)` acts like wrapping the callee in an option-setting group `(?:...)`. That is, while it sets the behavior for the callee, it doesn’t override options that are applied to more specific regions. In this example, the middle `"na"` in `"banana"` matches case-sensitively, despite the outer call to `ignoresCase()`:
+
+```swift
+let regex4 = Regex {
+    "ba"
+    "na".ignoresCase(false)
+    "na"
+}
+.ignoresCase()
+
+"banana".contains(regex4)     // true
+"BAnaNA".contains(regex4)     // true
+"BANANA".contains(regex4)     // false
+
+// Equivalent to:
+let regex5 = /(?i)ba(?-i:na)na/
+```
+
+All option APIs are provided on `RegexComponent`, so they can be called on a `Regex` instance, or on any component that you would use inside a `RegexBuilder` block when the `RegexBuilder` module is imported.
+
+The options that `Regex` supports are shown in the table below. Options that affect _matching behavior_ are supported through both regex syntax and APIs, while options that have _structural_ or _syntactic_ effects are only supported through regex syntax.
+
+| **Matching Behavior**        |                |                           |
+|------------------------------|----------------|---------------------------|
+| Case insensitivity           | `(?i)`         | `ignoresCase()`           |
+| Single-line mode             | `(?s)`         | `dotMatchesNewlines()`    |
+| Multi-line mode              | `(?m)`         | `anchorsMatchNewlines()`  |
+| ASCII-only character classes | `(?DSWP)`      | `asciiOnlyDigits()`, etc  |
+| Unicode word boundaries      | `(?w)`         | `wordBoundaryKind(_:)`    |
+| Semantic level               | `(?Xu)`        | `matchingSemantics(_:)`   |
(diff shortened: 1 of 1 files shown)


proposals/0000-unicode-for-string-processing.md
```diff
@@ -0,0 +1,901 @@
...
+
+Unicode property matching is extended to `Character`s with a goal of consistency with other regex character classes. For `\p{Decimal}` and `\p{Hex_Digit}`, only single-scalar `Character`s can match, for the reasons described in that section, above. For all other Unicode property classes, matching `Character`s can comprise multiple scalars, as long as the first scalar matches the property.
+
+To invert a Unicode property character class, use `\P{...}`.
+
+
+#### POSIX character classes: `[:NAME:]`
+
+**POSIX character classes** represent concepts that we'd like to define at all semantic levels. We propose the following definitions, some of which have been described above. When matching with grapheme cluster semantics, Unicode properties are extended to `Character`s as descrived in the rationale above, and as shown in the table below. That is, for POSIX class `[:word:]`, any `Character` that starts with a matching scalar is a match, while for `[:digit:]`, a matching `Character` must only comprise a single Unicode scalar value.
```
PsychoH13, line 702:
```suggestion
**POSIX character classes** represent concepts that we'd like to define at all semantic levels. We propose the following definitions, some of which have been described above. When matching with grapheme cluster semantics, Unicode properties are extended to `Character`s as described in the rationale above, and as shown in the table below. That is, for POSIX class `[:word:]`, any `Character` that starts with a matching scalar is a match, while for `[:digit:]`, a matching `Character` must only comprise a single Unicode scalar value.
```

proposals/0000-unicode-for-string-processing.md
```diff
@@ -0,0 +1,901 @@
...
+
+
+For compatibility with other regex engines and the flexibility to match at both `Character` and Unicode scalar level, you can switch between matching levels for an entire regex or within select portions. This powerful capability provides the expected default behavior when working with strings, while allowing you to drop down for Unicode scalar-specific matching.
+
+By default, literal characters and Unicode scalar values (e.g. `\u{301}`) are coalesced into characters in the same way as a normal string, as shown above. Metacharacters, like `.` and `\w`, and custom character classes each match a single element at the current matching level.
+
+For example, these matches fail, because by the time the parser encounters the "`\u{301}`" Unicode scalar literal, the full `"é"` character has been matched:
+
+```swift
+str.contains(/Caf.\u{301})    // false - `.` matches "é" character
```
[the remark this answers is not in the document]
natecook1000: Thanks, @PsychoH13!

proposals/0000-unicode-for-string-processing.md
```diff
@@ -0,0 +1,901 @@
...
+
+## Future directions
+
+### Expanded options and modifiers
+
+The initial version of `Regex` includes only the options described above. Filling out the remainder of options described in the [Run-time Regex Construction proposal][literals] could be completed as future work, as well as additional improvements, such as adding an option that makes a regex match only at the start of a string.
+
+### Extensions to Character and Unicode Scalar APIs
+
+An earlier version of this pitch described adding standard library APIs to `Character` and `UnicodeScalar` for each of the supported character classes, as well as convenient static members for control characters. In addition, regex literals support Unicode property features that don’t currently exist in the standard library, such as a scalar’s script or extended category, or creating a scalar by its Unicode name instead of its scalar value. These kinds of additions are
```
benrimmington, line 841: Unfinished sentence: "These kinds of additions are"

proposals/0000-unicode-for-string-processing.md
```diff
@@ -0,0 +1,1321 @@
...
+
+Regexes perform case sensitive comparisons by default. The `i` option or the `ignoresCase(_:)` method enables case insensitive comparison.
+
+```swift
+let str = "Café"
+	
+str.firstMatch(of: /CAFÉ/)          // nil
+str.firstMatch(of: /(?i)CAFÉ/)      // "Café"
+str.firstMatch(of: /(?i)cAfÉ/)      // "Café"
+```
```
milseman, line 222: This seems too interior-syntax heavy and we don't get to see the API name in use. What about having an additional line that uses `.ignoresCase()` after the literal?

proposals/0000-unicode-for-string-processing.md
```diff
@@ -0,0 +1,1321 @@
...
+
+```swift
+let queRegexScalar = queRegex.matchingSemantics(.unicodeScalar)
+print(composed.contains(queRegexScalar))
+// Prints "true"
+print(decomposed.contains(queRegexScalar))
+// Prints "false"
+```
+
+With grapheme cluster semantics, a grapheme cluster boundary is naturally enforced at the start and end of the match and every capture group. Matching with Unicode scalar semantics, on the other hand, can yield string indices that aren't aligned to character boundaries. Take care when using indices that aren't aligned with grapheme cluster boundaries, as they may have to be rounded to a boundary if used in a `String` instance.
```
milseman, line 391: The first sentence packs a lot. It seems like its own paragraph if not section to unpack what it implies. Also, what's special about captures that isn't special about anything else in a regex? Why isn't there a boundary around an alternation, character class, etc?

proposals/0000-unicode-for-string-processing.md
```diff
@@ -0,0 +1,1321 @@
...
+```
+
+The second call to `wholeMatch(in:)` fails because at the point the matching engine exits the `(?u:...)` group, the matching position is still in the middle of the `"e\u{301}\u{327}` character. This implicit grapheme cluster boundary assumption maintains the guarantee that capture groups over grapheme cluster semantic sections will have valid character-aligned indices.
+
+If a regex starts or ends with a Unicode scalar semantic section, there is no assertion added at the start or end of the pattern. Consider the following regex, which has Unicode scalars for the entire pattern except for a section in the middle that matches a purple heart emoji. When applied to a string with a multi-scalar character before or after the `"💜"`, the resulting match includes a partial character at its beginning and end.
+
+```swift
+let regex = Regex {
+    CharacterClass.any
+    Regex {  // <-- Implicit grapheme cluster boundary assertion, as above
```
milseman, line 430: Where? Before, after, or both?

proposals/0000-unicode-for-string-processing.md
```diff
@@ -0,0 +1,1321 @@
...
+```swift
+// Example `OneOrMore` initializer
+extension OneOrMore {
+    public init<W, C0, Component: RegexComponent>(
+    _ behavior: RegexRepetitionBehavior? = nil,
+    @RegexComponentBuilder _ component: () -> Component
+  ) where Output == (Substring, C0), Component.Output == (W, C0)
+}
+```
+
```
milseman, line 525: You're missing a definition of what this does to a literal, which has not orthogonal `nil` separate from `eager`.

proposals/0000-unicode-for-string-processing.md
```diff
@@ -0,0 +1,1321 @@
...
+    <<This string
+    uses double-angle-brackets
+    to group text.>>
+    """
+    
+str.firstMatch(of: /<<.+>>/)        // nil
+str.firstMatch(of: /(?s)<<.+>>/)    // "This string\nuses double-angle-brackets\nto group text."
+```
+
+This option applies only to `.` used in regex syntax and does _not_ affect the behavior of `CharacterClass.any`, which always matches any character or Unicode scalar. To get the default `.` behavior when using `RegexBuilder` syntax, use `CharacterClass.anyNonNewline`.
```
milseman, line 541: Does this argue there should be a `dot` character class that _does_ obey this mode? One of the goals of `CharacterClass` is to give an API for what can be expressed in a literal.

proposals/0000-unicode-for-string-processing.md
```diff
@@ -0,0 +1,1321 @@
...
+	
+str.firstMatch(of: /^abc/)          // "abc"
+str.firstMatch(of: /^abc$/)         // nil
+str.firstMatch(of: /(?m)^abc$/)     // "abc"
+	
+str.firstMatch(of: /^def/)          // nil
+str.firstMatch(of: /(?m)^def$/)     // "def"
+```
+
+This option applies only to anchors used in a regex literal. The anchors defined in `RegexBuilder` are specific about matching at the start/end of the input or the line, and therefore are not affected by this option.
```
milseman, line 574: Seems again like we also want equivalents of `$` and `^` as API. Those can just be a `if` check on options before forwarding onwards, and that's how the documentation can read.

proposals/0000-unicode-for-string-processing.md
```diff
@@ -0,0 +1,1321 @@
...
+
+_Grapheme cluster semantics:_ Matches a character made up of a single Unicode scalar that fits the hex digit criteria above.
+
+To invert the hexadecimal digit character class, use `CharacterClass.hexDigit.inverted`.
+
+*<details><summary>Rationale</summary>*
+
+Unicode's recommended definition for `\d` is its [numeric type][numerictype] of "Decimal" in contrast to "Digit". It is specifically restricted to sets of ascending contiguously-encoded scalars in a decimal radix positional numeral system. Thus, it excludes "digits" such as superscript numerals from its [definition][derivednumeric] and is a proper subset of `Character.isWholeNumber`. 
+
+We interpret Unicode's definition of the set of scalars, especially its requirement that scalars be encoded in ascending chains, to imply that this class is restricted to scalars which meaningfully encode base-10 digits. Thus, we choose to make the grapheme cluster interpretation *restrictive*.
```
milseman, line 712: But what does "restrictive" even mean? There's no mention of the character properties proposal or how this aligns with it

proposals/0000-unicode-for-string-processing.md
```diff
@@ -0,0 +1,1321 @@
...
+// Prints:
+// 👩🏾‍🚀
+// 👨🏻‍🎤
+// 🧑🏻‍💻
+```
+
+The table below shows our best effort at choosing the right manner of extending.
+
+| Property                            | Extension                     | Notes                             |
+|-------------------------------------|-------------------------------|-----------------------------------|
```
milseman, line 806: Double check with the character properties proposal. It might have left some things un or under-specified.

proposals/0000-unicode-for-string-processing.md
```diff
@@ -0,0 +1,1321 @@
...
+- `static func lowercaseMapping(_: String) -> CharacterClass`
+- `static func uppercaseMapping(_: String) -> CharacterClass`
+- `static func titlecaseMapping(_: String) -> CharacterClass`
+- `static func canonicalCombiningClass(_: Unicode.CanonicalCombiningClass) -> CharacterClass`
+
+You can see the full `CharacterClass` API with documentation comments in the **Custom Classes** section, below.
+
+#### POSIX character classes: `[:NAME:]` or `\p{NAME}`
+
+**POSIX character classes** represent concepts that we'd like to define at all semantic levels. We propose the following definitions, some of which have been described above. When matching with grapheme cluster semantics, Unicode properties are extended to `Character`s as described in the rationale above, and as shown in the table below. That is, for POSIX class `[:word:]`, any `Character` that starts with a matching scalar is a match, while for `[:digit:]`, a matching `Character` must only comprise a single Unicode scalar value.
```
milseman, line 898: This sounds like it's contrasting with the above. Is `[:word:]` the same as `\w` or not?

proposals/0000-unicode-for-string-processing.md
```diff
@@ -0,0 +1,1321 @@
...
+| `[:print:]`  | `[[:graph:][:blank:]--[:cntrl:]]` | starts-with          | `[[:graph:] ]`                |
+
+\* The Unicode scalar property definition for `[:word:]` is `[\p{Alphanumeric}\p{Mark}\p{Join_Control}\p{Connector_Punctuation}]`.  
+\*\* The Unicode scalar property definition for `[:cntrl:]` is `[^\p{Space}\p{Control}\p{Surrogate}\p{Unassigned}]`.
+
+#### Custom classes
+
+Custom classes function as the set union of their individual components, whether those parts are individual characters, individual Unicode scalar values, ranges, Unicode property classes or POSIX classes, or other custom classes.
+
+- Individual characters and scalars will be tested using the same behavior as if they were listed in an alternation. That is, a custom character class like `[abc]` is equivalent to `(a|b|c)` under the same options and modes.
```
milseman, line 923: But not capturing

proposals/0000-unicode-for-string-processing.md
```diff
@@ -0,0 +1,1321 @@
...
+  ///
+  /// This character class is equivalent to `\d` in regex syntax.
+  public static var digit: CharacterClass { get }
+  
+  /// A character class that matches any hexadecimal digit.
+  public static var hexDigit: CharacterClass { get }
+
+  /// A character class that matches any element that is a "word character".
+  ///
+  /// This character class is equivalent to `\w` in regex syntax.
```
milseman, line 1011: Is it? This is a departure from how `any` doesn't vary based on options.

proposals/0000-unicode-for-string-processing.md
```diff
@@ -0,0 +1,1321 @@
...
+  public static var word: CharacterClass { get }
+
+  /// A character class that matches any element that is classified as
+  /// whitespace.
+  ///
+  /// This character class is equivalent to `\s` in regex syntax.
```
milseman, line 1017: Is this affected by ascii only? Yes/no? Should there be an additional one that is or isn't affected?

milseman (changes requested):

proposals/0000-unicode-for-string-processing.md
natecook1000: Sure, I'll update to include API usage throughout this section. The goal is here is dual — we're explaining the semantics of the option and the two ways of applying it.

proposals/0000-unicode-for-string-processing.md
natecook1000: As with `.`, we don't want or need to emulate the unclear behavior of `^` and `$` in the `RegexBuilder` API. When writing those symbols in a regex, they always have a specific meaning; using `RegexBuilder` API we can give those specific meanings specific names.

The proposal mislabels the options as being a part of the `RegexBuilder` module; they're really part of `_StringProcessing`. Sorry about that! I'll update those labels.

proposals/0000-unicode-for-string-processing.md
```diff
@@ -0,0 +1,1321 @@
...
+let str = "<token>A value.</token>"
+	
+// By default, the '+' quantifier is eager, and consumes as much as possible.
+str.firstMatch(of: /<.+>/)          // "<token>A value.</token>"
+	
+// Adding '?' makes the '+' quantifier reluctant, so that it consumes as little as possible.
+str.firstMatch(of: /<.+?>/)         // "<token>"
+```
+
+The `U` option toggles the "eagerness" of quantifiers, so that quantifiers are reluctant by default, and only become eager when `?` is added to the quantifier. This change only applies within regex syntax. See the `defaultRepetitionBehavior(_:)` method, described above, for control over repetition behavior in `RegexBuilder` syntax.
```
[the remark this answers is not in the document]
natecook1000: Not sure what you mean here. In the builder, you can specify the behavior of individual quantifications or set a default for all quantifications in a particular scope. Is there a different behavior that you would like to bring from the regex syntax?

proposals/0000-unicode-for-string-processing.md
```diff
@@ -0,0 +1,1321 @@
...
+    Repeat(.digit, ...3)
+}
+```
+
+You can build custom character classes by combining regex-defined classes with individual characters or ranges, or by performing common set operations such as subtracting or negating a character class.
+
+
+#### “Any”
+
+The simplest character class, representing **any character**, is written as `.` and is sometimes referred to as the "dot" metacharacter. This class always matches a single `Character` or Unicode scalar value, depending on the matching semantic level. This class excludes newlines, unless "single line mode" is enabled (see section above).
```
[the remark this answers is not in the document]
natecook1000: The paragraph that follows this one describes the difference between `any` and "dot" — did you see that?

proposals/0000-unicode-for-string-processing.md
```diff
@@ -0,0 +1,1321 @@
...
+// Prints:
+//  ̀
+//  ́
+//  ̂
+// ...
+```
+
+#### Any grapheme cluster
+
+`Regex` also provides a way to match a single grapheme cluster, regardless of the current semantic level. The **any grapheme cluster** character class is written as `\X` or `CharacterClass.anyGraphemeCluster`, and matches from the current location up to the next grapheme cluster boundary. This includes matching newlines, regardless of any option settings. This metacharacter is equivalent to the regex syntax `(?Xs:.)`.
```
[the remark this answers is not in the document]
natecook1000: `(?Xs:.)` means "in grapheme cluster, single-line mode, consume a single element". `.+?\y` would also work. `CharacterClass.anyGraphemeCluster` is the builder syntax that this paragraph is describing.

proposals/0000-unicode-for-string-processing.md
```diff
@@ -0,0 +1,1321 @@
...
+
+### Include `\O` and `CharacterClass.anyUnicodeScalar`
+
+An earlier draft of this proposal included a metacharacter and `CharacterClass` API for matching an individual Unicode scalar value, regardless of the current matching level, as a counterpart to `\X`/`.anyGraphemeCluster`. The behavior of this character class, particularly when matching with grapheme cluster semantics, is still unclear at this time, however. For example, when matching the expression `\O*`, does the implict grapheme boundary assertion apply between the `\O` and the quantification operator, or should we treat the two as a single unit and apply the assertion after the `*`?
+
+At the present time, we prefer to allow authors to write regexes that explicitly shift into and out of Unicode scalar mode, where those kinds of decisions are handled by the explicit scope of the setting. If common patterns emerge that indicate some version of `\O` would be useful, we can add it in the future.
```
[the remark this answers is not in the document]
natecook1000: We should be erroring out on `\O` if it isn't supported.

proposals/0000-unicode-for-string-processing.md
natecook1000: Yep! I'll make that change:
```suggestion
- Individual characters and scalars will be tested using the same behavior as if they were listed in an alternation. That is, a custom character class like `[abc]` is equivalent to `(?:a|b|c)` under the same options and modes.
```

proposals/0000-unicode-for-string-processing.md
```diff
@@ -0,0 +1,1321 @@
...
+
+The **hexadecimal digit** character class is matched by  `CharacterClass.hexDigit`.
+
+_Unicode scalar semantics:_ Matches a decimal digit, as described above, or an uppercase or small `A` through `F` from the _Halfwidth and Fullwidth Forms_ Unicode block. Note that this is a broader class than described by the `UnicodeScalar.properties.isHexDigit` property, as that property only include ASCII and fullwidth decimal digits.
+
+_Grapheme cluster semantics:_ Matches a character made up of a single Unicode scalar that fits the hex digit criteria above.
+
+To invert the hexadecimal digit character class, use `CharacterClass.hexDigit.inverted`.
+
+*<details><summary>Rationale</summary>*
```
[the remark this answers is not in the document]
natecook1000: I noticed that too, I'll see if there's something I can do.

proposals/0000-unicode-for-string-processing.md
milseman: Maybe a sub-section about the outputs of a regex and how those indices are aligned.

proposals/0000-unicode-for-string-processing.md
```diff
@@ -0,0 +1,1321 @@
...
+let borahae = "태형💜아미"    // Note: These hangeul characters are decomposed
+if let match = borahae.firstMatch(of: regex) {
+    print(match.0)
+}
+// Prints "ᆼ💜ᄋ"
+```
+
+Boundaries from a grapheme cluster section into a Unicode scalar also imply a grapheme cluster boundary, but in this case no assertion is needed. This boundary is an emergent property of the fact that under grapheme cluster semantics, matching always happens one character at a time.
+
+**Regex syntax:** `(?X)...` or `(?X...)` for grapheme cluster semantics, `(?u)...` or `(?u...)` for Unicode scalar semantics.
```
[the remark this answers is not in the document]
milseman: Or should it be `(?-u)`? Can this be punted until there's a byte mode?

PR #1636, opened 2022-04-26
Add Regex Literal proposal

hamishknight:


proposals/nnnn-regex-literals.md
```diff
@@ -0,0 +1,485 @@
...
+// regex: Regex<(Substring, Substring?)>
+```
+
+The optional wrapping will nest arbitrarily if the capture is nested within multiple zero-quantifiers or alternations:
+
+```swift
+let r = /(.)*|\d/
+// regex: Regex<(Substring, Substring??)>
+``` 
+
```
milseman, line 97: I don't think this behavior makes sense and I don't think we should let temporary (I hope) limitations from `RegexBuilder` spill over into permanently locking us into the wrong type system representation.

milseman (changes requested):
Temporarily marking as request changes so we don't accidentally merge. We'll discuss what to do.

proposals/nnnn-regex-literals.md
milseman: My takeaway from the meeting is that we want to type check this as

```swift
// regex: Regex<(Substring, Substring??)>
```

And we'll be exploring ways to explicitly drop a single layer of optionality for the DSL conversion (turtles all the way down)

hamishknight: Updated to say that we will apply at most one layer of optionality. Do we want to talk about the DSL conversion in this proposal?

proposals/nnnn-regex-literals.md
```diff
@@ -69,44 +69,42 @@ Due to the existing use of `/` in comment syntax and operators, there are some s
...
-// regex: Regex<(Substring, Substring?)>
+let regex1 = /([ab])?/
+// regex1: Regex<(Substring, Substring?)>
+
+let regex2 = /([ab])|\d+/
+// regex2: Regex<(Substring, Substring?)>
 ```
 
-it also occurs when it appears as a child of an alternation:
+This does not affect cases where the capture surrounds the zero quantifier or alternation:
```
milseman, line 87:
```suggestion
A zero quantifier or alternation nested within a capture do not produce an optional capture, unless the capture itself is inside a zero quantifier or alternation:
```

milseman (approved):

proposals/nnnn-regex-literals.md
milseman: Hmm, hard to say. It's not a language/library thing (other than we happen to be using the same library as the parser to achieve it).

hamishknight: Hmm yeah, in that case I think it's probably fine to leave as-is

PR #1639, opened 2022-05-01
Add alternative considered and clarify motivation

milseman:


hamishknight (approved):

proposals/0354-regex-literals.md
```diff
@@ -38,7 +38,9 @@ let regex = /(?<identifier>[[:alpha:]]\w*) = (?<hex>[0-9A-F]+)/
 // regex: Regex<(Substring, identifier: Substring, hex: Substring)>
 ```
 
-Forward slashes are a regex term of art. They are used as the delimiters for regex literals in, e.g., Perl, JavaScript and Ruby. Perl and Ruby additionally allow for [user-selected delimiters](https://perldoc.perl.org/perlop#Quote-and-Quote-like-Operators) to avoid having to escape any slashes inside a regex. For that purpose, we propose the extended literal `#/.../#`.
+Forward slashes are a regex term of art, predating their use in modern languages literals. They are also used in, e.g., Perl, JavaScript and Ruby. They have clear aesthetic value for use cases where there are no contained unescaped `/`s present.
```
milseman, line 41:
```suggestion
Forward slashes are a regex term of art. The association between forward slashes and regexes dates back to 1969's ed, the first Unix editor, and it was inherited by subsequent interactive text tools like less and vim. The syntax was also adopted by the sed language; from there it passed to Perl, and then to Ruby and Javascript. Forward slash is instantly recognizable as a regex; the only common alternative is an ordinary string literal passed to a library API, which usually has extra overhead, requires more escaping, and defers regex syntax errors to runtime. The proposed Swift regex literals do not have these limitations, so forward slash provides the right behavioral cues to developers. There are over fifty years of precedents for forward slash and very little for anything else. 
```

proposals/0354-regex-literals.md
```diff
@@ -487,17 +506,28 @@ Instead of adding a custom regex literal, we could require users to explicitly w
 
 We therefore feel this would be a much less compelling feature without first class literal support.
 
+### Restrict feature set to that of the builder DSL
+
+The regex builder DSL is unable to provide some of the features presented such as named captures as tuble labels. An alternative could be to cut those features from the literal out of concern they may lead to an over-use of the literals. However, to do so would remove the clearest demonstration of the need for better type-level operations including working with labeled tuples.
+
+Similarly, there is no literal equivalent for some of the regex builder features, but that isn't an argument against them. The regex builder DSL has references which serves this role (though not as concisely) and they are useful beyond just naming captures.
+
+Regex literals should not be outright avoided, they should be used well. Artifically hampering their usage doesn't provide any benefit and we wouldn't want to lock these limitations into Swift's ABI.
```
masters3d, line 515: What does Artificially hampering mean in this context?

milseman: Arbitrarily reducing the utility for no technical reason

PR #1640, opened 2022-05-02
[SE-0305] Typo: Use correct checksum computation command.

fabianfett:
I had to go into SE-0272 to find the correct command to compute the checksum. I guess we should fix this in SE-0305.


rjmccall (approved):

rjmccall:
Thanks!

PR #1651, opened 2022-05-13
[SE-0354] Update regex literal proposal

hamishknight:
Update to specify the new parsing rules.

### DIFF
--- a/proposals/0354-regex-literals.md
+++ b/proposals/0354-regex-literals.md
@@ -139,7 +139,7 @@ let regex = #/
   usr/lib/modules/ # Prefix
   (?<subpath> [^/]+)
   /vmlinuz          # The kernel
-#/
+/#
 // regex: Regex<(Substring, subpath: Substring)>
 ```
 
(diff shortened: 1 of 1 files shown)


proposals/0354-regex-literals.md
```diff
@@ -248,132 +246,123 @@ let regex = Regex {
...
+baz(/, /)  // Will become the regex literal '/, /'
+baz(/^, /) // Will become the regex literal '/^, /'
+baz(!/, /) // Will become prefix '!' on the regex literal '/, /'
 
-func baz(_ x: S) -> Int {
-  x[(/)] + x[/]
-}
-```
+// Will become the regex literal '/] /'
+let d = hasSubscript[/] / 2
```
PsychoH13, line 317: This case does not seem to make sense to me, the opening bracket is not a character operator which means the `[` becomes unbalanced. With this rule, the line would be parsed as keyword `let`, identifier `d`, operator identifier `=`, identifier `hasSubscript`, subscript opening bracket `[`, regular expression `/] /`, literal `2`, which leads to a bunch errors:
- Expected ',' separator
- No exact matches in call to subscript
It would be like writing `let d = hasSubscript[regex 2`

Wouldn't it make more sense to apply the same rule to `]` and `}`  that is applied to parentheses?

hamishknight: The issue is that both `/] /` and `/} /` are valid regex patterns. We could ban them, but it's not clear to me that cases such as `hasSubscript[/] / 2` are common enough to warrant that (we haven't encountered any such cases in the source compatibility testing we've done so far). It is however something we could explore if we start encountering cases like this in the wild.

PR #1659, opened 2022-05-19
[SE-0356] Add named snippet slices to future directions

bitjammer:


JCDiana:
> 

More direct knowledge is a must. Spinning in circles and grasping for sort of straight forward help. Being in the dark is not way.

tomerd (approved):

PR #1671, opened 2022-05-30
SE-0344: Fix example code in the proposal

MaxDesiatov:
Multiple fixes to example code of the proposal in this PR:

* The example code makes a reference to `Res.self`, which is not declared anywhere. From context and from sample code provided by the toolchain, it makes sense that `Res` (short for "Result"?) is the third generic argument of the function, which here is named as `Success`. 
* That same example code was referring to `data`, which from the context was meant to be `responseData`. Missing `return` statement was added as well to make this code compile in a broader example context. This example snippet is updated accordingly.
* Another snippet had wrong type of `argument: Argument`, which was updated to `argument: RemoteCallArgument<Argument>` to match the `DistributedTargetInvocationEncoder` protocol requirement.
* Argument labels for `DistributedTargetInvocationEncoder` implementation were updated to match the protocol requirement.
* Example code apparently should have `DistributedTargetInvocationDecoder` as a `struct`, not `class`. It doesn't have any initializers for this to be a valid `class`.
* `ClusterTargetInvocationDecoder` implementation should have its functions marked as `mutating` to match requirements of the `DistributedTargetInvocationDecoder` protocol.
* `RemoteCallTarget` does not have `mangledName` property, apparently it was meant to be `identifier` in the example code.

### DIFF
--- a/proposals/0344-distributed-actor-runtime.md
+++ b/proposals/0344-distributed-actor-runtime.md
@@ -1199,7 +1199,7 @@ Once that is complete, the runtime will pass the constructed `InvocationEncoder`
     // [4] decode the response from the response bytes
     // in our example system, we're using Codable as SerializationRequirement,
     // so we can decode the response like this (and never need to cast `as? Codable` etc.):
-    try self.someDecoder.decode(as: Success.self, from: data)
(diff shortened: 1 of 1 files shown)


ktoso:
Thanks, yep seems all those are right fixups.

Thanks!

PR #1673, opened 2022-05-30
[SE-0354] Update proposal

hamishknight:
- Specify the behavior around `(?-x)` and `\Q...\E`
- Add additional alternatives considered


milseman (approved):
LGTM.

Do you want to address the alternative where Swift's lexer and source tools fully understand option setting/unsetting (including isolated options)?

hamishknight:
> Do you want to address the alternative where Swift's lexer and source tools fully understand option setting/unsetting (including isolated options)?

Added some discussion of this. I think it's only really a problem if it's syntax that is done as part of the delimiter, as then it interferes with the "newline after delimiter means multi-line" check. If we generally supported a semantic whitespace multi-line literal, allowing matching option changing within that should just work I think as that could be handled by the parser.

PR #1693, opened 2022-06-23
Proposal to add a move(x) function for ending local variable lifetimes

jckarter:


proposals/nnnn-move-function.md
```diff
@@ -0,0 +1,455 @@
...
+test.swift:13:17: note: move here
+  consumeFinalY(move(y))
+                ^
+test.swift:19:13: note: use here
+  useYAgain(y)
+            ^
+```
+
+Notice how the compiler gives us all of the information that we need to resolve
+this: it tells us where the move was and gives tells us the later uses that
```
gottesmm, line 153: `gives tells` looks like a thought stream being crossed.

proposals/nnnn-move-function.md
```diff
@@ -0,0 +1,455 @@
...
+  useX(x)
+  let other = x
+  let _ = move(x)
+  useX(move(other))
+  consumeX(other)
+  useX(x)
+}
+```
+
+we get separable nice diagnostics:
```
gottesmm, line 194: I am not sure if nice is needed here. But I am fine with it staying ; ).

gottesmm (commented):
Two small nits.

hborla:
Hi @gottesmm and @atrick, I'll be managing the review for this proposal! I just finished going through the pitch thread and reading the updated proposal. Here are my notes on reviewers' feedback that I think you should consider addressing in the proposal:

* The term "binding" seems to have caused significant confusion in the pitch thread - consider defining it in the proposal.
* The proposal is missing a "Motivation" section. The motivation is outlined in the introduction, but please put the motivation under a "Motivation" heading.
* The alternative syntax suggestions are already covered in the proposal, but it may be helpful to more directly discuss the mentioned "potential impact to the language syntax" that is minimized by choosing to define `move` as a function. 
  * For example, `any` was recently added as a contextual keyword in expression context, and to avoid stomping on the [top-level `any` function](https://developer.apple.com/documentation/swift/any(_:)), you cannot parenthesize the constraint operand to a contextual `any` keyword in expressions.
* Clarification on what types of values can be passed to `move`
    * https://forums.swift.org/t/pitch-move-function-use-after-move-diagnostic/53983/33
    * https://forums.swift.org/t/pitch-move-function-use-after-move-diagnostic/53983/10
* Justification for not annotating `move` with `@discardableResult`.
  * Michael [addressed this](https://forums.swift.org/t/pitch-move-function-use-after-move-diagnostic/53983/19) in the pitch discussion, but I think it's worth adding to the proposal.
* How `move` fits into the overarching ownership story, perhaps by linking the [recent roadmap](https://forums.swift.org/t/a-roadmap-for-improving-swift-performance-predictability-arc-improvements-and-ownership-control/54206) in a "Future Directions" section.

Let me know if you have any questions! I'll schedule this for review once you've had a chance to consider my notes.

proposals/nnnn-move-function.md
```diff
@@ -0,0 +1,455 @@
...
+  // I can still use x here!
+}
+// [PATH1][PATH2] (continuation)
+// Since I reinitialized x along [PATH1] I can reuse the var here.
+```
+
+Notice how in the above, we are able to use `x` both in the true block AND the
+continuation block since over all paths, x now has a valid value.
+
+The value based analysis uses Ownership SSA to determine if values are used
```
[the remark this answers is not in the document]
tbkka: I don't think we should be referring to Ownership SSA in a language-level document.  We should focus on _what_ the compiler does, not _how_ it does it.

proposals/nnnn-move-function.md
```diff
@@ -0,0 +1,455 @@
...
+```
+useX(x) // do some stuff with local variable x
+
+// Ends lifetime of x, y's lifetime begins.
+let y = move(x) // [1]
+
+useY(y) // do some stuff with local variable y
+useX(x) // error, x's lifetime was ended at [1]
+
+// Ends lifetime of y. Since _ is no-op, we perform an actual release here.
```
tbkka, line 35: Whether there is a "release" depends on the type of `x`.  Focus on the language implications:
```
  Ends the lifetime of the variable `y`.  Assigning to `_` avoids creation of a new lifetime.
```
I would also through here be careful to distinguish the lifetime of a "variable" from the lifetime of the "value" that may be temporarily stored in that value.

proposals/nnnn-move-function.md
```diff
@@ -0,0 +1,455 @@
...
+* Bugs: [SR-NNNN](https://bugs.swift.org/browse/SR-NNNN), [SR-MMMM](https://bugs.swift.org/browse/SR-MMMM)
+* Previous Revision: [1](https://github.com/apple/swift-evolution/blob/...commit-ID.../proposals/NNNN-filename.md)
+* Previous Proposal: [SE-XXXX](XXXX-filename.md)
+-->
+
+## Introduction
+
+In this document, we propose adding a new function called `move` to the swift
+standard library, which ends the lifetime of a specific local `let`,
+local `var`, or `consuming` parameter. In order to enforce this, the compiler will
```
tbkka, line 22: Overall suggestion for making markup documents easier to review:  Put a line break after every full stop.

I suggest you add a sentence here after "parameter."
```
This gives developers a way to control retain/release operations and avoid
unnecessary copy-on-write behaviors in performance-critical code.
```
Then this paragraph will provide a good summary:  A sentence saying _what_ you propose, a sentence about _why_, and a sentence summarizing _how_ the compiler will support this operation.

proposals/nnnn-move-function.md
```diff
@@ -0,0 +1,455 @@
...
+useY(y) // do some stuff with local variable y
+useX(x) // error, x's lifetime was ended at [1]
+
+// Ends lifetime of y. Since _ is no-op, we perform an actual release here.
+let _ = move(y) // [2]
+useX(x) // error, x's lifetime was ended at [1]
+useY(y) // error, y's lifetime was ended at [2]
+```
+
+This allows the user to influence the uniqueness of COW data structures
```
tbkka, line 41: I would start the "Motivation" section here, and reword things some to focus
on the problems that you're trying to solve:
```
Performance-critical code can suffer from unexpected copy-on-write operations
and unnecessary retain/release calls.
Consider the following array/uniqueness example:
```

proposals/nnnn-move-function.md
```diff
@@ -0,0 +1,455 @@
...
+and reduce retain/release calls in a way that is future-proof against changes
+to the language implementation or changes to the source code that unexpectedly
+extend the lifetime of moved variables. Consider the following array/uniqueness
+example:
+
+```
+func test() {
```
tbkka, line 48: Good example!

proposals/nnnn-move-function.md
```diff
@@ -0,0 +1,455 @@
...
+  longAlgorithmUsing(&y)
+  consumeFinalY(y)
+
+  // We no longer use y after this point. Ideally, x would be guaranteed
+  // unique so we know we can append again without copying.
+  x.append(7)
+}
+```
+
+in the example above, without `move`, `y`'s formal lifetime extends to the end of
```
tbkka, line 67: Capitalize "In"

proposals/nnnn-move-function.md
```diff
@@ -0,0 +1,455 @@
...
+programmers modifying this code in the future
+may introduce new references to `y` that inadvertently extend its lifetime
+and break our attempt to keep `x` unique.
+
+Swift-evolution pitch thread: [https://forums.swift.org/t/pitch-move-function-use-after-move-diagnostic](https://forums.swift.org/t/pitch-move-function-use-after-move-diagnostic)
+
+
+## Proposed solution: Move Function + "Use After Move" Diagnostic
+
+That is where the `move` function comes into play. The `move` function is a new
```
tbkka, line 80: Strike the sentence "That is where ... "

proposals/nnnn-move-function.md
```diff
@@ -0,0 +1,455 @@
...
+generic stdlib function that when given a local let, local var, or parameter
+argument provides a compiler guarantee to the programmer that the binding will
```
tbkka, line 82: This is a little oddly phrased: "The move function provides a _compiler_ guarantee to the _programmer_?"  I would rather say that the programmer can insert `move` operations in order to explicitly end the lifetime of a particular binding.  This provides guaranteed behavior in the face of future changes to the surrounding code or to the compiler optimization pipeline.

proposals/nnnn-move-function.md
```diff
@@ -0,0 +1,455 @@
...
+  var y = x
+  longAlgorithmUsing(&y)
+  // We think we no longer use y after this point...
+  consumeFinalY(move(y))
+
+  // ...and x will be unique again here...
+  x.append(7)
+
+  // ...but this additional use of y snuck in:
+  useYAgain(y)
```
tbkka, line 134: Be very careful to put an `// Error` comment on any line in any example that is an error.

proposals/nnnn-move-function.md
```diff
@@ -0,0 +1,455 @@
...
+      ^
+test.swift:13:17: note: move here
+  consumeFinalY(move(y))
+                ^
+test.swift:19:13: note: use here
+  useYAgain(y)
+            ^
+```
+
+Note how the compiler gives all of the information that we need to resolve
```
tbkka, line 152: Strike this paragraph.  It only talks about the quality of your prototype implementation, which is not the issue.  This document should focus on the language feature and the need for it.

(Personally, I would cut out most of the previous example as well; only the first error message is truly relevant.)

proposals/nnnn-move-function.md
```diff
@@ -0,0 +1,455 @@
...
+  useX(x)
+  let other = x   // other is a new binding used to extend the lifetime of x
+  _ = move(x) // x's lifetime ends
+  useX(other)     // other is used here... no problem.
+  consumeX(other) // other is used here... no problem.
+}
+```
+
+In fact, each variable's lifetime is tracked independently, and gets a separate
+diagnostic if used after move. If we try to compile this:
```
tbkka, line 177: This is redundant with the paragraphs just above.  I would strike this paragraph, the following example, and the diagnostics after that.

proposals/nnnn-move-function.md
```diff
@@ -0,0 +1,455 @@
...
+  useX(x) // error: no value in x
+  x = getValue()
+  useX(x) // ok, x has a new value here
+}
+```
+
+This follows from move being applied to the binding (`x`), not the value in the
+binding (the value returned from `getValue()`).
+
+We also support applying the move operation to consuming function arguments:
```
tbkka, line 234: "We also support" is a statement about your implementation.  Rather,  "The move operation can also be applied to ...." is a statement about the language feature.

proposals/nnnn-move-function.md
```diff
@@ -0,0 +1,455 @@
...
+
+```
+func f(_ x: __owned SomeClassType) {
+    let _ = move(x)
+    useX(x) // !! Error! Use of x after move
+}
+```
+
+Normal arguments are passed by borrow, meaning that the lifetime of the value
```
tbkka, line 243: Good clarification.

proposals/nnnn-move-function.md
```diff
@@ -0,0 +1,455 @@
...
+being moved from and used again; however, since the final value of an
+`inout` argument is passed back to the caller, an `inout` argument *must* be
+reassigned by the callee before it returns:
+
+```
+func f(_ buffer: inout Buffer) { // error: 'buffer' not reinitialized after move!
+  let b = move(buffer)           // note: move was here
+  b.deinitialize()
+  ... write code ...
+}                                // note: return without reassigning inout argument `buffer`
```
tbkka, line 260: I would put the error on the close brace for purposes of exposition here.  (Even though I understand that's not how the current prototype works.)

proposals/nnnn-move-function.md
```diff
@@ -0,0 +1,455 @@
...
+```
+
+But we can to re-initialize `buffer` by writing the following code:
+
+```
+func f(_ buffer: inout Buffer) {
+  let b = move(buffer)
+  b.deinitialize()
+  // ... write code ...
+  // We re-initialized buffer before end of function so the checker is satisfied
```
tbkka, line 270: Strike "so the checker is satisfied"

proposals/nnnn-move-function.md
```diff
@@ -0,0 +1,455 @@
...
+  buffer = getNewInstance()
+}
+```
+
+Move analysis understands that `defer` statements run before every scope exit,
```
tbkka, line 275:
```
s/Move analysis understands that/Because/
```

proposals/nnnn-move-function.md
```diff
@@ -0,0 +1,455 @@
...
+  let b = move(buffer)
+  // Ensure the buffer is reinitialized before we exit.
+  defer { buffer = getNewInstance() }
+  b.deinitialize()
+  // ... write code ...
+}
+```
+
+
+In the future, we may add support for globals and stored properties, although
```
tbkka, line 290: Again, you're talking about the implementation.  You need to explain why you do not think the feature requires this support today.

Hmmm....  This paragraph might be better off moved to "alternatives considered".

proposals/nnnn-move-function.md
```diff
@@ -0,0 +1,455 @@
...
+```
+/// This function ends the lifetime of the passed in binding.
+@_transparent
+@alwaysEmitIntoClient
+func move<T>(_ t: __owned T) -> T {
+  Builtin.move(t)
+}
+```
+
+Builtin.move is a hook in the compiler to force emission of special SIL "move"
```
tbkka, line 325: I think you should delete the rest of this section from this point.

At most, I would like to see a brief paragraph that says in essence that "Builtin.move() causes the compiler to perform additional data flow analysis to verify that the argument is not in fact used after this point."  The details of how that is implemented in this particular compiler is not relevant; a different compiler may implement it in very different ways.

proposals/nnnn-move-function.md
```diff
@@ -0,0 +1,455 @@
...
+
+Notice how in the above, we are able to use `x` both in the true block AND the
+continuation block since over all paths, x now has a valid value.
+
+The value based analysis uses Ownership SSA to determine if values are used
+after the move and handles non-address only lets. The address based analysis is
+an SSA based analysis that determines if any uses of an address are reachable
+from a move. All of these are already in tree and can be used today by invoking
+the stdlib non-API function `_move` on a local let or move. *NOTE* This function
+is always emit into client and transparent so there isn't an ABI impact so it is
```
tbkka, line 401: This NOTE is useful to keep, since it explains the library implementation shown at the top of this section.

proposals/nnnn-move-function.md
```diff
@@ -0,0 +1,455 @@
...
+
+## Effect on ABI stability
+
+None, move will always be a transparent always emit into client function.
+
+## Effect on API resilience
+
+None, this is additive.
+
+## Alternatives considered
```
tbkka, line 417: This is a reasonable place to explain why you don't think global support is necessary and could be considered separately.

proposals/nnnn-move-function.md
```diff
@@ -0,0 +1,455 @@
...
+- an expression attribute, like `useX(@move x)`
+- a compiler directive, like `useX(#move(x))`
+
+There are also potentially other names besides `move` that we could use. We're
+proposing using the name `move` because it is an established term of art in
+other programming language communities including C++ and Rust, as well as a
+term that has already been used in other Swift standard library APIs such as
+the `UnsafeMutablePointer.move*` family of methods that move a value out of
+memory referenced by a pointer. Declaring it as a function also minimizes the
+potential impact to the language syntax. We are however open to discussing
```
tbkka, line 437: Strike the sentence beginning "We are however open ..."

tbkka (dismissed):
On the whole, the examples are good, but there's far too much discussion of how your particular implementation of the proposed feature works.

Here's a way to think about it:  Pretend that some other group had implemented a from-scratch implementation of a Swift compiler.  That compiler might not use SIL nor OSSA nor any other particular internal technology, but everything in this document should still be relevant to that other implementation.

jckarter:
I'll go ahead and start working on the requested revisions. Thanks @hborla and @tbkka!

jckarter:
@hborla @tbkka How's this look now?

proposals/nnnn-move-function.md
```diff
@@ -0,0 +1,493 @@
...
+
+  // ...but this additional use of y snuck in:
+  useYAgain(y)
+}
+```
+
+In this case, we get the following output from the compiler as expected:
+
+```swift
+test.swift:10:7: error: 'y' used after being moved
```
tbkka, line 151: I would drop this entire example and just put the error message directly into the example above:
```
   useYAgain(y) // Error: 'y' used after being moved
```

proposals/nnnn-move-function.md
tbkka: On further reflection, this paragraph really needs to go.

tbkka:
Definitely improved.  Still some work to do, but it's improving nicely.

proposals/nnnn-move-function.md
jckarter: Done.

hborla (approved):
Thank you!

proposals/0366-move-function.md
```diff
@@ -0,0 +1,481 @@
...
+  // I can still use x here!
+  useX(x) // OK
+}
+// But I can't use x here.
+useX(x) // !! ERROR! Use after move.
+```
+
+If the binding is a `var`, the analysis additionally allows for code to
+conditionally reinitialize the var and thus be able to use it in positions
+that are dominated by the reinitialization.  continuation path. Consider the
```
tbkka, line 330: Extra words?

PR #1698, opened 2022-06-29
[SE-0311] Clarify the diagnosis behavior of @TaskLocal

YuAo:
> we propose to diagnose and fail at compile time if the @TaskLocal property wrapper is defined on a static or global property

Should be: "_we propose to diagnose and fail at compile time if the @TaskLocal property wrapper is_ **not** _defined on a static or global property_"


YuAo:
@ktoso ?

ktoso:
That's right, they must be declared global or static -- thanks for spotting the typo.

PR #1701, opened 2022-06-30
[SE-0363] Minor fixes

ole:
- Fixes a regex literal that doesn't correspond with the regex builder version directly below
- Fixes a broken link
- Deletes a duplicate paragraph

cc @natecook1000 @Azoy


Azoy (approved):
This looks good to me! @airspeedswift ?

PR #1722, opened 2022-07-18
Update retroactive conformances proposal (SE-0364)

harlanhaskins:
This updates the proposal to apply to all modules, not just resilient
libraries. It also updates the warning text to more clearly explain
what the consequences are of such a construct.


harlanhaskins:
The implementation has been updated to reflect this new direction.

PR #1730, opened 2022-07-29
[SE-289] Adjust selection statements examples to match reality

xedin:
Examples made it look like `buildOptional` calls are injected
after enclosing `if` when it doesn't have an else, which is
incorrect - `buildOptional` happens inside the block and `else`
is synthesized by the result builder transform.


DougGregor (approved):

DougGregor:
Thank you!

PR #1739, opened 2022-08-11
`take` and `borrow` parameter ownership modifiers

jckarter:


proposals/NNNN-parameter-ownership-modifiers.md
```diff
@@ -428,6 +449,100 @@ func hackPasswords() throws -> HackedPasswords {
...
+inside of a `taking func`:
+
+```
+extension FileHandle {
+  // Return the file descriptor back to the user for manual management
+  // and disable automatic management with the FileHandle value.
+
+  taking func giveUp() -> Int32 {
+    // `deinit fd` is strawman syntax for consuming a value without running
+    // its initializer. it is only allowed inside of `taking func` methods
```
tbkka, line 526: Should the above be "its deinitializer"?

PR #1748, opened 2022-08-18
[SE-0370] Add syntax highlighting for two code blocks

xwu:


xwu:
@rjmccall @glessard Mind if I make this editorial tweak?

glessard:
Thanks @xwu. These are too easy to miss!

PR #1776, opened 2022-09-11
Add more thorough examples to SE-0365 proposal body

calda:
Based on feedback and discussion from the proposal review, this PR updates the body of SE-0365 to include more thorough examples.


compnerd (approved):
Thanks for the improvements to the text here!

PR #1778, opened 2022-09-13
Removed the “Syntactic sugar for `if let` self-assignment” case.

wadetregaskis:
As of Swift 5.7 there actually is this exact shorthand (`if let foo {` as an abbreviation of `if let foo = foo {`).


Jumhyn:
This has already been [removed](https://github.com/apple/swift-evolution/commit/77cb9392aa550c9313c73aa9b95be3b6508b6c83) on `main`.

wadetregaskis:
Ah!  I didn't realise 'master' had been replaced by 'main'.  Sorry for the noise!

PR #1796, opened 2022-09-27
[SE-0311] Fix typos

97mik:


PsychoH13:
While you're here you can change the `spawn let` to `async let` and replace `group.spawn`to `group.addTask`.

rjmccall (approved):

PR #1811, opened 2022-10-11
SE-0305: add a link to SE-0272 for referenced type

MaxDesiatov:
`binaryTarget` type is mentioned in SE-0305, but it's not clear where it's coming from in this proposal. Adding a link to SE-0272 adds more context for binary targets introduction.


DougGregor:
Thanks!

PR #1816, opened 2022-10-16
[se-0370] Mark as implemented

glessard:
Mark SE-0370 as implemented.


glessard:
The primary associated types issue hasn't been resolved yet, but it doesn't affect the API in use. It's just less readable.

PR #1817, opened 2022-10-17
Add Enum Case KeyPaths

Azoy:
Pending evolution pitch


rjmccall:
This has stalled out; I'm going to close it, but we're open to pursuing it in the future.

PR #1823, opened 2022-10-26
[DNM] Function Values and Sendable (version 1)

kavon:


proposals/nnnn-sendable-function-values.md
```diff
@@ -0,0 +1,525 @@
...
+
+This section provides a summary of the solutions and changes proposed for Swift. For complete details and caveats, see the Detailed design.
+
+### Inferring `@Sendable` for methods
+
+In [SE-302](0302-concurrent-value-and-concurrent-closures.md), the `@Sendable` attribute was introduced for both closures and named functions/methods. Beyond allowing the function value to cross actor boundaries, the attribute primarily influences the kinds of values that can be captured by the function. But methods of a nominal type cannot capture anything but the object instance itself. Furthermore, a non-local (i.e., a global or static) function cannot capture _anything_, because a reference to a global declaration is not considered a capture of that value. Thus, the proposed simplifications are:
+
+1. the inference of `@Sendable` on all methods of a type that conforms to `Sendable`.
+2. the inference of `@Sendable` on all non-local functions.
+3. the prohibition of marking a method `@Sendable` if the object type does not conform to `Sendable`.
```
beccadax, line 69: The indirect phrasing of these rules makes them a little difficult to understand. If I’m interpreting your text correctly, I’d suggest phrasing like this:

```suggestion
1. All methods of a type that conforms to `Sendable` shall be implicitly `@Sendable`.
2. All non-local functions shall be implicitly `@Sendable`.
3. It is not possible to mark a method `@Sendable` if the enclosing type does not conform to `Sendable`.
```

However, I’m still left with a few questions:

* Rule 1 only applies to *instance* methods, right?
* When Rule 2 talks about “non-local functions”, what exactly does it mean? I think it means free functions and static/class methods, but it’s not totally clear.
* Which functions do not fall under Rule 1, Rule 2, or Rule 3? That is, when would you still have a choice about whether a function is `@Sendable` and be able to explicitly mark it as such? (Maybe just local functions and perhaps closures?)

rjmccall:
I'm going to close this PR; we're likely to review this in smaller chunks.

PR #1824, opened 2022-10-26
Add proposal for `AsyncStream` convenience factories

FranzBusch:


c01a651mn7 (approved):

beccadax:
Pitch thread: https://forums.swift.org/t/pitch-convenience-async-throwing-stream-makestream-methods/61030

proposals/NNNN-async-stream-factory.md
```diff
@@ -0,0 +1,166 @@
...
+    return .init(stream: stream, continuation: continuation)
+  }
+}
+```
+
+## Source compatibility, Effect on ABI stability, Effect on API resilience
+
+As this is an additive change, it should not have any compatibility, stability or resilience problems. The only potential problem would be if someone has already run into this shortcoming and decided to define their own `makeStream` methods.
+
+## Alternatives considered
```
beccadax, line 131: I looked through the pitch and identified a couple of alternatives you may want to add:

* [Using an initializer on `NewStream` instead of a static method that returns `NewStream`](https://forums.swift.org/t/pitch-convenience-async-throwing-stream-makestream-methods/61030/20)
* [Separately allocating the continuation and passing it into a stream `init`](https://forums.swift.org/t/pitch-convenience-async-throwing-stream-makestream-methods/61030/27)

Discussing these ideas and preemptively rebutting them helps to keep reviewers from rehashing topics you've already settled.

proposals/NNNN-async-stream-factory.md
```diff
@@ -0,0 +1,166 @@
...
+the closure that is passed to the initialiser.
+Escaping the continuation is slightly inconvenient since it requires a dance
+around an implicitly unwrapped optional. Furthermore, the closure implies
+that the continuation lifetime is scoped to the closure which it isn't.
+
+## Proposed solution
+
+In order to fill this gap, I propose to add a new static method `makeStream` on
+`AsyncStream` and `AsyncThrowingStream` that returns both the stream
+and the continuation.
```
beccadax, line 40: I would recommend you add a simple example of how this API would be used. This doesn't have to be anything elaborate—you can just call the new method and pass the two components off to code in different tasks. I think you can just give an example for `AsyncStream` and let people infer that it'll work the same for `AsyncThrowingStream`, or vice versa.

(To further strengthen your argument, you could add a matching example to the Motivation section showing how clumsy the closure-based API is.)

beccadax (changes requested):
I have some suggestions to improve the proposal, but this is generally looking pretty good!

proposals/NNNN-async-stream-factory.md
```diff
@@ -0,0 +1,166 @@
...
+```
+
+## Source compatibility, Effect on ABI stability, Effect on API resilience
+
+As this is an additive change, it should not have any compatibility, stability or resilience problems. The only potential problem would be if someone has already run into this shortcoming and decided to define their own `makeStream` methods.
+
+## Alternatives considered
+
+### Return a tuple instead of a concrete type
+My initial pitch was using a tuple as the return paramter of the factory;
```
beccadax, line 134: There's a misspelling here, but the terminology is a little off anyway:
```suggestion
My initial pitch was using a tuple as the result type of the factory;
```

proposals/NNNN-async-stream-factory.md
```diff
@@ -0,0 +1,166 @@
...
+however, I walked back on it since I think we can provide better documentation on
+the concrete type. Furthermore, it makes it more discoverable as well.
+
+The upside of using a tuple based approach is that we can backdeploy it. 
+
+An implementation returning a tuple would look like this;
+
+```swift
+extension AsyncStream
```
beccadax, line 143: Nit: Missing `{`

beccadax (approved):
LGTM. Thanks for the improvements!

beccadax:
Scheduled for February 15th–26th.

PR #1831, opened 2022-10-27
SE-0337: fix broken `@available` declaration

MaxDesiatov:
`@available(unavailable, *)` as currently stated in the proposal won't compile, the correct order of arguments to this attribute is `@available(*, unavailable)`.


Jumhyn:
Thanks Max!

PR #1833, opened 2022-10-27
[SE-0366] Fix "missing or invalid dates" warning

benrimmington:
* Use full month names (for the [status page][]).
* Add a link to the second review.

[status page]: <https://apple.github.io/swift-evolution/#?proposal=SE-0366>


hborla (approved):
Thank you! Note that I actually copy-pasted the dates from https://github.com/apple/swift-evolution/blob/main/proposals/0377-parameter-ownership-modifiers.md, so it's wrong there too 🙂 I can fix that though unless you want to include that in this PR

PR #1834, opened 2022-10-27
[SE-0376] Fix "missing or invalid dates" warning

benrimmington:
* Use full month names (for the [status page][]).

[status page]: <https://apple.github.io/swift-evolution/#?proposal=SE-0376>


Jumhyn:
Ah thanks Ben!

PR #1836, opened 2022-11-01
Mark SE-0365 as implemented in Swift 5.8

calda:
The implementation for SE-0365 in Swift 5 mode has been merged (https://github.com/apple/swift/pull/61520), so we can mark it as implemented in SE-0365

@compnerd @xwu


xwu:
Thanks!

PR #1846, opened 2022-11-17
Proposal for `@noImplicitCopy` attribute for local variable and parameter bindings

jckarter:


proposals/NNNN-no-implicit-copy-bindings.md
```diff
@@ -0,0 +1,487 @@
...
+
+    ```
+    @noImplicitCopy let x: Optional = getValue()
+
+    if borrow y = x {}
+
+    // We can still use x because we only borrowed its payload
+
+    switch x {
+    case .some(borrow y):
```
atrick, line 477: Was shadowing 'y' intentional here?

jckarter: There isn't any shadowing, really,  since `if borrow y` would be out of scope by the time we reach the switch. But I can still change the name to be clear.

jckarter:
We're going a different direction with this.

PR #1847, opened 2022-11-18
[DRAFT] borrow-operator

gottesmm:
Very initial draft. Don't take it too seriously.

### DIFF
--- a/proposals/0000-borrow-operator.md
+++ b/proposals/0000-borrow-operator.md
@@ -1,0 +1,189 @@
+# Borrow Operator, Pass-by-Borrow, and Pass-by-Value
+
+* Proposal: [SE-NNNN](NNNN-filename.md)
+* Authors: ...
+* Review Manager: TBD
+* Status: ...
+
+<!--
+
+*During the review process, add the following fields as needed:*
+
+* Implementation: [apple/swift#NNNNN](https://github.com/apple/swift/pull/NNNNN) or [apple/swift-evolution-staging#NNNNN](https://github.com/apple/swift-evolution-staging/pull/NNNNN)
+* Decision Notes: [Rationale](https://forums.swift.org/), [Additional Commentary](https://forums.swift.org/)
+* Bugs: [SR-NNNN](https://bugs.swift.org/browse/SR-NNNN), [SR-MMMM](https://bugs.swift.org/browse/SR-MMMM)
+* Previous Revision: [1](https://github.com/apple/swift-evolution/blob/...commit-ID.../proposals/NNNN-filename.md)
+* Previous Proposal: [SE-XXXX](XXXX-filename.md)
+
+-->
+
+## Introduction
+
+In this document, we propose adding a new operator to the language, marked by
+the context sensitive keyword `borrow`, that causes the compiler to pass
+arguments using a `pass-by-borrow` convention instead of a `pass-by-value`
+convention.
+
+Swift-evolution thread: [Discussion thread topic for that proposal](https://forums.swift.org/)
+
+## Motivation
+
+Swift normally passes arguments using a `pass-by-value` convention. A
+`pass-by-value` argument convention implies that each argument is passed as a
+separate value to the callee. This has the effect that the compiler is allowed
+(and sometimes required) to introduce temporary copies of arguments into the
+program in caller callsites.
+
+As an elucidating example, consider the following code:
+
+```
+func useValueTwice(_ x: Type, _ y: Type) {}
+
+var a: Type = ...
+useValueTwice(a, a)
+```
+
+Since `useValueTwice` is using default Swift conventions, we pass `a` as two
+separate guaranteed parameters to `useValueTwice`. Inside `useValueTwice`, `x`
+and `y` both appears to be separate truly borrowed values in the sense that
+`useValueTwice` has immutable access to two different values and does not take
+ownership of those values. In order for the compiler to achieve this, it must
+necessarily copy `a` to ensure that it can pass two different values to
+`useValueTwice`. These copies are observable program behavior and are in fact a
+desired default behavior since otherwise in general cases we couldn't write code
+like the above that takes the same argument multiple times. We instead would
+necessarily require the compiler to insert an error similar to an exclusivity
+violation which is undesireable:
+
+```
+useValueTwiceWithoutPassByValue(a, a) // Error! Cannot pass a value twice!
+```
+
+In certain cases though, this `pass-by-value` behavior and the implicit copies
+it requires the compiler to insert may be undesireable due to the author needing
+strict performance guarantees that implicit copies like the above are never
+generated. For most programs this guarantee is not needed since the optimizer
+generally does a good enough job eliminating such copies. But in critical system
+code, good enough is not an acceptable constraint and a true semantic guarantee
+is needed.
+
+## Proposed solution
+
+We propose introducing a new context sensitive keyword called `borrow` that when
+applied to a function argument forces Swift to use a `pass-by-borrow` convention
+on a specific argument instead of a `pass-by-value` convention. This would thus
+forbid the compiler from inserting such copies and instead cause the compiler to
+emit an error in such a case:
+
+```
+var a: Type = ...
+useValueTwice(borrow a, borrow a) // Error! Cannot overlapping borrows of 'a'!
+```
+
+The result of the usage of borrow is that a performance minded programmer would
+be able to sleep well at night knowing that their code doesn't have copies in
+this code.
+
+## Detailed design
+
+The `borrow` keyword can be applied in argument positions to lets, vars, and
+arguments. The application of the keyword would cause the compiler to use a
+`pass-by-borrow` convention for that parameter instead of a `pass-by-value`
+convention. First we define the `pass-by-value` convention as follows:
+
+1. `let` and non-`inout` parameters: Formally, the compiler always copies the
+   `let` or non-`inout` parameter into a new independent value and passes the
+   value as the parameter. To the callee this value appears as an independent
+   immutable value that is different from all other parameters.
+
+2. `var` and `inout` parameters: Formally the compiler after formal evaluations
+   have completed, emits an access to the lvalue and then copies the value from
+   the lvalue into a new rvalue. The rvalue is then passed as an independent
+   value to the guaranteed parameter. Similar to the `let` case, this rvalue in
+   the callee appears as an independent immutable value that is different from
+   all other parameters.
+
+We define the `pass-by-borrow` convention as follows:
+
+1. `let` and non-inout parameters: This would cause the compiler to not emit
+   defensive copies when calling guaranteed parameters. Since a let is immutable
+   and a guaranteed parameter is immutable, we would allow for a let to be
+   passed multiple times as a `pass-by-borrow` parameter.
+
+2. `var` and `inout` parameters: This would cause the compiler to formally
+   evaluate the `var`/`inout` using a read exclusivity scope and then use a
+   borrowed formal access to pass the value to the function argument. This is
+   the same mechanism used to pass vars as an inout parameter except we have
+   exclusive read access instead of exclusive write access to the given memory.
+
+Since the `var`/`inout` case takes exclusive read access to the underlying
+memory, once cannot pass such a binding multiple times to the same function:
+
+```
+var a: Type = ...
+// Error! Cannot take exclusive access to the same variable twice
+useValueTwice(borrow a, borrow a)
+```
+
+necessarily this implies that we must also error if one passes a `var`/`inout`
+once as a `by-borrow` parameter and once as a `by-value` parameter, one will
+also achieve an exclusivity violation since one will attempt to access the
+`var`/`inout` for the lvalue-to-rvalue conversion /after/ the formal evaluation
+of the borrowed argument has begun. Example:
+
+```
+var a: Type = ...
+// Error! Cannot take exclusive access to the same variable twice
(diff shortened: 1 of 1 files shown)


atrick:
Since this is a language proposal, the ABI is irrelevant. The only copies we care about are "semantic" copies, not physical copies.

`foo(x, x)` may or may not require a physical copy depending on the ABI. But each argument *can* semantically borrow the same instance of 'x' of the compiler knows it's safe.

Passing-by-borrow means the that the Callee does not need its own separate copy and there's no need to retain the passed-in value..
The following needs a temporary copy/retain for pass-by-copy. But for pass-by-borrow, it's illegal.
```
var x ...
foo(x, { x = ...})
```

proposals/0000-borrow-operator.md
```diff
@@ -0,0 +1,189 @@
...
+```
+var a: Type = ...
+useValueTwice(borrow a, borrow a) // Error! Cannot overlapping borrows of 'a'!
+```
+
+The result of the usage of borrow is that a performance minded programmer would
+be able to sleep well at night knowing that their code doesn't have copies in
+this code.
+
+## Detailed design
```
kavon, line 87: Overall, there's a lot of swift-compiler-specific jargon that needs to either be explained or revised to use terms that an average compiler person who has not worked on the Swift compiler can understand. like:

- "guaranteed parameter"
- "a borrowed formal access"

>one will also achieve an exclusivity violation since one will attempt to access the `var`/`inout` for the lvalue-to-rvalue conversion /after/ the formal evaluation of the borrowed argument has begun. Example:

none of these words are in the bible :) .... more seriously, consider whether details like this are actually important for the community to understand when evaluating this _language_ proposal. Esoteric details that require an understanding of concepts specific to the _implementation_ are irrelevant unless if your proposal requires a major overhaul of the compiler.

proposals/0000-borrow-operator.md
```diff
@@ -0,0 +1,189 @@
...
+// Call function
+useValueTwice(firstArgAddr, rvalue)
+
+// Tear down
+end_access firstArgAddr
+```
+
+## Source compatibility
+
+`borrow` will only be allowed to be applied to lvalues (similar to `take`) and
```
kavon, line 163: "lvalue" doesn't mean much here or in most places in the document, as a lot of things can appear on the left-side of an assignment that are probably not allowed in a `borrow`.

kavon (commented):
Good start!

proposals/0000-borrow-operator.md
```diff
@@ -0,0 +1,189 @@
...
+* Previous Revision: [1](https://github.com/apple/swift-evolution/blob/...commit-ID.../proposals/NNNN-filename.md)
+* Previous Proposal: [SE-XXXX](XXXX-filename.md)
+
+-->
+
+## Introduction
+
+In this document, we propose adding a new operator to the language, marked by
+the context sensitive keyword `borrow`, that causes the compiler to pass
+arguments using a `pass-by-borrow` convention instead of a `pass-by-value`
```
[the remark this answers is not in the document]
gottesmm: I was purposely using this since I want it to be viewed as a key word for the reader to pay attention to. Similar to the way that JoeG highlighted borrow and take in the following proposal: https://github.com/apple/swift-evolution/blob/main/proposals/0377-parameter-ownership-modifiers.md

proposals/0000-borrow-operator.md
```diff
@@ -0,0 +1,189 @@
...
+
+Swift-evolution thread: [Discussion thread topic for that proposal](https://forums.swift.org/)
+
+## Motivation
+
+Swift normally passes arguments using a `pass-by-value` convention. A
+`pass-by-value` argument convention implies that each argument is passed as a
+separate value to the callee. This has the effect that the compiler is allowed
+(and sometimes required) to introduce temporary copies of arguments into the
+program in caller callsites.
```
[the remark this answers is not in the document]
gottesmm: I think it is important to mention this here. But I agree the wording is unfortunate... let me try rephrasing it and tell me what you think. I am willing to be convinced that it isn't the right place for it though if that doesn't work.

proposals/0000-borrow-operator.md
```diff
@@ -0,0 +1,189 @@
...
+necessarily copy `a` to ensure that it can pass two different values to
+`useValueTwice`. These copies are observable program behavior and are in fact a
+desired default behavior since otherwise in general cases we couldn't write code
+like the above that takes the same argument multiple times. We instead would
+necessarily require the compiler to insert an error similar to an exclusivity
+violation which is undesireable:
+
+```
+useValueTwiceWithoutPassByValue(a, a) // Error! Cannot pass a value twice!
+```
```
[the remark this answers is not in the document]
gottesmm: I think the appropriate phrase to use is a borrowed argument convention like JoeG did here: https://github.com/apple/swift-evolution/blob/main/proposals/0377-parameter-ownership-modifiers.md. I think parse of the purpose of this document is to make it clear that there is a difference in between borrowing in the callee vs borrowing in the caller. I think that is the way potentially to make this clearer to user. Borrowing from the perspective of the callee is what JoeG is talking about here: https://github.com/apple/swift-evolution/blob/main/proposals/0377-parameter-ownership-modifiers.md. I am trying to talk about borrowing from the perspective of the caller and hopefully being clear here.

proposals/0000-borrow-operator.md
```diff
@@ -0,0 +1,189 @@
...
+
+In certain cases though, this `pass-by-value` behavior and the implicit copies
+it requires the compiler to insert may be undesireable due to the author needing
+strict performance guarantees that implicit copies like the above are never
+generated. For most programs this guarantee is not needed since the optimizer
+generally does a good enough job eliminating such copies. But in critical system
+code, good enough is not an acceptable constraint and a true semantic guarantee
+is needed.
```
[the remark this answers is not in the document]
gottesmm: I agree. Let me see what I can come up with.

proposals/0000-borrow-operator.md
```diff
@@ -0,0 +1,189 @@
...
+arguments. The application of the keyword would cause the compiler to use a
+`pass-by-borrow` convention for that parameter instead of a `pass-by-value`
+convention. First we define the `pass-by-value` convention as follows:
+
+1. `let` and non-`inout` parameters: Formally, the compiler always copies the
+   `let` or non-`inout` parameter into a new independent value and passes the
+   value as the parameter. To the callee this value appears as an independent
+   immutable value that is different from all other parameters.
+
+2. `var` and `inout` parameters: Formally the compiler after formal evaluations
```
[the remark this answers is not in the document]
gottesmm: Maybe the way to do this is to flip it around?

`inout` parameters and vars:

proposals/0000-borrow-operator.md
gottesmm: I need to clean up the language here based off of https://github.com/apple/swift-evolution/blob/main/proposals/0366-move-function.md. That was my intention.

proposals/0000-borrow-operator.md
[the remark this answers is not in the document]
gottesmm: I made this no longer relevant.

proposals/0000-borrow-operator.md
```diff
@@ -0,0 +1,189 @@
...
+emit an error in such a case:
+
+```
+var a: Type = ...
+useValueTwice(borrow a, borrow a) // Error! Cannot overlapping borrows of 'a'!
+```
+
+The result of the usage of borrow is that a performance minded programmer would
+be able to sleep well at night knowing that their code doesn't have copies in
+this code.
```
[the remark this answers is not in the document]
gottesmm: I answered these questions. Only thing I need to look at is the borrow a.x. It is going to have the same behavior as we have for inout, except with no write back given a computed property and using immutable accesses.

proposals/0000-borrow-operator.md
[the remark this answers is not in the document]
gottesmm: I am not sure if putting in an example where the optimizer fails is useful since the optimizer can change and what may be a failure today may not be a failure in the future. I think it is good enough to say that the optimizer /does/ fail sometimes.

proposals/0000-borrow-operator.md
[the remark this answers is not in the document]
gottesmm: I actually did this differently than I described here. Tell me what you think of my "redoing" here.

proposals/0000-borrow-operator.md
```diff
@@ -0,0 +1,195 @@
...
+
+1. Potentially trigger exclusivity violations. For instance, if we formally
+accessed the inout parameters memory before performing the `pass-by-value`, we
+would have an exclusivity violation when attempting to exclusively access the
+`var` memory to perform our copy.
+
+2. If we did not copy the parameter, we would need to either perform a `take` of
+the non-`inout` parameter meaning that said memory would now be
+invalidated. This would mean that we couldn't pass in var storage directly as an
+`inout` parameter since the memory would be invalidated.
```
tbkka, line 58: The above is awkward.  `either` is missing an `or`.   Rather than "perform a `take`" (an undefined term at this point), I would more conversationally explain that passing a reference for the first argument would result in two references within the function to the same storage and that this "aliasing" is a well-known problem in programming languages such as C that permit unchecked references to be passed to functions.

proposals/0000-borrow-operator.md
```diff
@@ -0,0 +1,195 @@
...
+For this and other reasons, we pass these arguments as `pass-by-value`
+formally. These copies are observable program behavior and as shown above are
+desireable in our formal model of callsite evaluation. In certain cases though,
+this `pass-by-value` behavior and the implicit copies it requires the compiler
+to insert may be undesireable due to the author needing strict performance
+guarantees that implicit copies like the above are never generated. For most
+programs this guarantee is not needed since the optimizer generally does a good
+enough job eliminating such copies. But in critical system code, "good enough"
+is not an acceptable constraint and a true semantic language level guarantee is
+needed.
```
tbkka, line 69: Most of the discussion above isn't really necessary.  An SE proposal doesn't need to rationalize historical decisions.  I think your example above is good, but you might only need to explain that the first argument is copied, the second is (internally) a reference, and that this makes the behavior safe without the function implementation needing to account for the possibility that these were originally the same value.

proposals/0000-borrow-operator.md
```diff
@@ -0,0 +1,195 @@
...
+
+We propose introducing a new context sensitive keyword called `borrow` that when
+applied to a local var, computed var, or global var passed as a function
+argument forces Swift to use a `pass-by-immutable-borrow` convention on a
+specific argument instead of a `pass-by-value` convention. This would forbid the
+compiler from inserting such copies and instead cause the compiler to pass the
+argument in the same manner as an inout argument except performing an immutable
+exclusive access to memory instead of a mutable exclusive access. The result of
+the usage of borrow is that a performance minded programmer would be able to
+sleep well at night knowing that their code doesn't have copies in this code.
```
tbkka, line 81: This last sentence serves no purpose.

proposals/0000-borrow-operator.md
```diff
@@ -0,0 +1,195 @@
+# Borrow Operator, Pass-by-Immutable-Borrow, and Pass-by-Value
```
tbkka, line 1: "Pass-by-Value" should not be in the title, since that's not something you're proposing.

proposals/0000-borrow-operator.md
```diff
@@ -0,0 +1,195 @@
...
+The `borrow` keyword can be applied in argument positions to local, computed,
+and global `vars`. It will cause the `var` to use a `pass-by-immutable-borrow`
+convention. Lets revisit our `useValueTwice` example from above in more detail:
+
+```
+func useValueTwice(_ x: Type, _ y: inout Type) {}
+
+var a: Type = ...
+useValueTwice(a, &a)
+```
```
tbkka, line 94: I am very confused at this point.  This is exactly the same code as above, but the discussion below makes it sound like you are proposing to _change_ the semantics of this code.

Shouldn't there be an example here that actually has `borrow` keywords in it?

proposals/0000-borrow-operator.md
```diff
@@ -0,0 +1,195 @@
...
+
+// Ok! Immutable exclusivity doesn't conflict.
+takeBorrowAndBorrow(borrow x, borrow x)
+
+// Error! Exclusivity violation!
+takeInOutAndBorrow(&x, borrow x)
+
+// Error! Exclusivity violation!
+takeBorrowAndClosure(borrow x, { x = Type() })
+```
```
tbkka, line 159: You've not shown a single line of code with `borrow` in a function definition.

tbkka (commented):
Some initial comments.  I find this pretty confusing still.

proposals/0000-borrow-operator.md
gottesmm: The reason why I included it in the title is that I wanted to contrast it with Pass-by-Immutable borrow. That being said, I agree that it can be removed.

proposals/0000-borrow-operator.md
gottesmm: I am explaining to the reader in the detailed design section in detail the way the callsite evaluation occurs in order to contrast it with when we apply the borrow convention. I don't think that I can assume that all of our users understand the way that the language defines formal evaluation of call sites.

gottesmm: I think I have an idea of how to make this a little clearer. I should first define the rules for formal evaluation of callsites and then show the example afterwards.

PR #1860, opened 2022-12-05
[SE-0379][Opt-In Reflection] Allow same file conformance declaration

maxovtsin:
Allow same file conformance declaration


AnthonyLatsis:
@jckarter ping

rjmccall:
This seems to have stalled since the return for revision. Is it actually just waiting on LSG input?

PR #1867, opened 2022-12-07
Fix ABI / API typo

keith:
I believe the intent here was API


keith:
I believe the intent here was `API`

PR #1874, opened 2022-12-08
Swift Mach Port Interface

loffgren:


loffgren:
Moving this to the Swift forum:
https://forums.swift.org/t/mach-port-api/61930

tkremenek:
This PR has already been closed, but for context, Swift System is an independent OSS project, not part of the core Swift project, and as such Swift Evolution is not the process to get changes into it.

PR #1876, opened 2022-12-10
Update 0380-if-switch-expressions.md

benrimmington:
* Add missing text.
* Enable syntax highlighting.
* Convert tabs to spaces.


hborla (approved):
Thank you!

PR #1877, opened 2022-12-12
Add package-access-modifier proposal

elsh:


proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,176 @@
...
+If `-package-name` is not passed, but `package` access modifier is used, then an error will be thrown; Building files that do not have `package` symbols should continue to work without needing to pass in `-package-name`. 
+
+
+### Exportability
+
+The package symbols will be stored in .swiftmodule but not in .swiftinterface. The package name will be stored in both .swiftmodule and .swiftinterface, but only the swiftmodule will be loaded when building a same-package module; this will prevent loading a fall-back module (swiftinterface) which does not contain package symbols.  
+
+The exportability rule at the declaration site will be similar to the the existing behavior; for example, a public class is not allowed to inherit a package class, and a public func is not allowed to have a package type in its signature. 
+
+The `@inlinable` will be applicable to `package`; we plan to introduce`@usableFromPackageInline` to allow references from `@inlinable package` and not from `@inlinable public`.
```
rjmccall, line 139:
```suggestion
`package` functions can be made `@inlinable`.  Just like with `@inlinable public`, not all symbols are usable within the function: they must be `open`, `public`, `package`, or `@usableFromInline`.  Note that `@usableFromInline` allows the use of a symbol from `@inlinable` functions whether they're `package` or `public`; this proposal does not contain a way to export a symbol only for use by `@inlinable package` functions, although we plan to add one in a future proposal.
```

Although, I've got to say that that's a pretty narrow follow-up proposal.  Have you considered just including this now?

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,176 @@
...
+
+Package symbols will be exported in the final libraries/executables and emitted in llvm.used; we plan to introduce a build setting that lets users decide whether to hide package symbols if statically linked.
```
rjmccall, line 141: Okay, you definitely do not need to mention `llvm.used` in an evolution proposal.  I suggest moving this whole bit to the paragraph in Future Directions where you talk about future optimization potential.

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Subclassing / Overrides
+
+A `package` class will not be allowed to be subclassed from another module, thus a `package` class member will not be allowed to be overidden either. 
+
+This follows the behavior and optimization model for `public`; a `public` class has no subclasses outside of its defining module, and a `public` class member has no overrides outside of its defining module. The optimizer can take advantage of this information in whole-module builds to replace dynamic dispatch with static dispatch among other things.
+
+It also allows a linear progression from `private` through `open` where each step incrementally offers capabilities to a wider range of clients. If `package` were to allow subclassing/overrides outside of its defining modules, making a `package` class `public` would break subclasses defined in other modules within the same package. 
+
+If a `package` class needs to be subclassed, it will need to be declared `open`, or be converted to a protocol, which is a recommended practice.
```
rjmccall, line 152: I understand where you're coming from here, but I don't think the linear-progression argument is very convincing.  You're right that allowing subclassing within the package is inconsistent with `public` and makes going from `package` to `public` problematic.  However, not having a way to even *allow* subclassing within the package without going to `open` makes going from `internal` to `package` (i.e. splitting a subclassable class out into a separate module within the package) impossible without introducing all of the problems from the motivation.  That seems like a serious defect.  At the very least, you need to discuss solving it as a future direction, but I think you should consider whether this is reasonable to just address in the proposal.

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,176 @@
...
+## Future Directions
+
+Limiting the scope of visibility per package can open up a whole lot of optimization opportunities. A package containing several modules can be treated as a resilience domain.
+
+Even with the library evolution enabled, modules within a package could be treated non-resilient. If same-package clients need access to swiftmodules, they don't need to be independently rebuildable and could have an unstable ABI; they could avoid resilience overhead and language rules such as the `@unknown default` requirement.
+
+
+## Source compatibility
+
+The changes is additive, so it's not source breaking.
```
rjmccall, line 164: Refer back to the discussion about keywords (that you haven't written yet).

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Effect on ABI stability
+
+Does not impact ABI stability.
```
rjmccall, line 168: This section is more of a general place to talk about ABI implications, and under that understanding, you should:
- acknowledge that the boundaries between separately-built modules within a package are still potentially ABI boundaries
- say that the ABI for `package` symbols is not different from the ABI for `public` symbols, pending the future direction to not export things that can be resolved within an image

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Alternatives considered
+
+A current workaround is to use `@_spi`, `@_implemenationOnly`, or `@testable`. Each option has caveats. The `@_spi` requires a group name, which makes it verbose to use and harder to keep track of, and `@_implementationOnly` can be too limiting as we want to be able to restrict access to only portions of APIs. The `@testable` elevates all internal symbols to public, which leads to an increase of the binary size and the shared cache size. If there are multiple symbols with the same name from different modules, they will clash and require module qualifiers everywhere. It is hacky and is strongly discouraged for use.
+
+## Acknowledgments
+
+Alexis Laferriere, Doug Gregor, Becca Royal-Gordon, Allan Shortlidge, Artem Chikin, and Xi Ge provided helpful feedback and analysis. Alexis Laferriere has been part of the discussion from the beginning and reviewed implementation of the prototype.
```
rjmccall, line 176: Alexis is listed as a co-author; if you're keeping that, he should not be acknowledged here.

rjmccall (changes requested):
Okay, here's my initial review as review manager.  I'll take a look at the pitch thread tomorrow.

rjmccall:
Notes from the review thread:
- Alternatives Considered should contrast this proposal with the idea of introducing submodules.  Maybe we should talk that over offline at some point.
- Per Tony Allevato's comment, please discuss the possibility of enforcing that this isn't actually done outside of the package.  (This would be possible if e.g. you had to pass the complete set of modules in the package to the compiler.)
- The discussion about dead-stripping and other optimization should be fleshed out.
- Please discuss the expected interaction of packages with library evolution.  (e.g. [this comment](https://forums.swift.org/t/new-access-modifier-package/61459/29))

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,176 @@
...
+* Review: ([pitch](https://forums.swift.org/t/new-access-modifier-package/61459))
+
+## Introduction
+
+This proposal introduces `package` as a new access modifier. Currently, to access a symbol in another module, that symbol needs to be declared `public`. It allows access across modules within the same library but also outside of the library, which is undesirable. The new access modifier enables a more fine control over the visibility scope of such symbols. 
+
+## Motivation
+
+Packages are often composed of multiple modules; packages exist as a way to organize modules in Swift, and organizing often involves splitting a module into smaller modules. For example, a module containing internal helper APIs can be split into a utility module only with the helper APIs and the other module(s) containing the rest. In order to access the helper APIs, however, the helper APIs need to be made public. The side effect of this is that they can “leak” to a client that should not have access to those symbols. Besides the scope of visibility, making them public also has an implication on the code size and performance.
```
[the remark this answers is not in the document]
elsh: Thanks for the suggestion!

proposals/NNNN-package-access-modifier.md
elsh: I can include that in this proposal.

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,176 @@
...
+```
+
+When building the Engine module, the package name 'gamePkg' and package symbols will be stored in the Engine.swiftmodule. When building Game, the input to its package name 'gamePkg' will be compared with Engine's package name; since they match, access to package symbols is allowed. When building App, the name comparison shows 'appPkg' is different from its dependency module's so access to package symbols will be deined, which is what we want.
+
+If `-package-name` is not passed, but `package` access modifier is used, then an error will be thrown; Building files that do not have `package` symbols should continue to work without needing to pass in `-package-name`. 
+
+
+### Exportability
+
+The package symbols will be stored in .swiftmodule but not in .swiftinterface. The package name will be stored in both .swiftmodule and .swiftinterface, but only the swiftmodule will be loaded when building a same-package module; this will prevent loading a fall-back module (swiftinterface) which does not contain package symbols.
```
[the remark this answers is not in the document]
elsh: Similar to .private.swiftinterface, we plan to introduce .package.swiftinterface which contains package symbols. For MVP though, we plan to just store package symbols in .swiftmodule.

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Alternatives considered
+
+A current workaround for the scenario in the Motivation is to use `@_spi` or `@_implemenationOnly`. Each option has caveats. The `@_spi` requires a group name, which makes it verbose to use and harder to keep track of, and `@_implementationOnly` can be too limiting as we want to be able to restrict access to only portions of APIs. There are also hacky workarounds such as `@testable` and the `-disable-access-control` flag. They elevate all `internal` (and `private` with the flag) symbols to `public`. These options are unstable and also quickly lead to an increase of the binary and the shared cache size, not to mention symbol name clashes.
+
+Instead of a package, submodules could have been considered. Introducing another "umbrella" module or dividing a module into "sub" modules, however, would require much more significant work and would be both source and ABI breaking. A package already exists as a unit corresponding to a group of modules, and treating it as a unit above a module in the grouping hierarchy enables a nautral transition into a submodule-like structure.
```
rjmccall, line 222: I think you both want to talk about what submodules mean and argue against why they don't really solve the problem.  ("We already have packages" is not a good argument.  If submodules were a better solution, then we should add submodules, not just try to pick low-hanging fruit when it leads us to a worse solution.)  Maybe something like this?

Instead of adding a new `package` access level above modules, we could allow modules to contain other modules as components.  This is an idea often called "submodules".  Packages would then define an "umbrella" module that contains the package's modules as components. However, there are several weaknesses in this approach:
- It doesn't actually solve the problem by itself. Submodule APIs would still need to be able to declare whether they're usable outside of the umbrella or not, and that would require an access modifier. It might be written in a more general way, like `internal(MyPackage)`, but that generality would also make it more verbose.
- Submodule structure would be part of the source language, so it would naturally be source- and ABI-affecting.  For example, programmers could use the parent module's name to qualify identifiers, and symbols exported by a submodule would include the parent module's name.  This means that splitting a module into submodules or adding an umbrella parent module would be much more impactful than desired; ideally, those changes would be purely internal and not change a module's public interface.  It also means that these changes would end up permanently encoding package structure.
- The "umbrella" submodule structure doesn't work for all packages.  Some packages include multiple "top-level" modules which share common dependencies.  Forcing these to share a common umbrella in order to use package-private dependencies is not desirable.
- In a few cases, the ABI and source impact above *would* be desirable.  For example, many packages contain internal `Utility` modules; if these were declared as submodules, they would naturally be namespaced to the containing package, eliminating spurious collisions. However, such modules are generally not meant to be usable outside of the package at all. It is a reasonable future direction to allow whole modules to be made package-private, which would also make it reasonable to automatically namespace them.

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,226 @@
...
+
+Packages, as expressed by the Swift Package Manager, are a unit of code distribution. Some packages contain just a single module, but it's frequently useful to split a package's code into multiple modules. For example, when a module contains some internal helper APIs, those APIs can be split out into a utility module and maybe reused by other modules or packages.
+
+However, because Swift does not recognize organizations of code above the module level, it is not possible to create APIs like this that are purely internal to the package. To be usable from other modules within the package, the API must be public, but this means it can also be used outside of the package. This allows clients to form unwanted source dependencies on the API. It also means the built module has to export the API, which has negative implications for code size and performance.
+
+For example, here’s a scenario where a client (App) has access to a utility API from a package (gamePkg) it depends on.
```
rjmccall, line 24:
```suggestion
For example, here’s a scenario where a client (`App`) has access to a utility API from a package (`gamePkg`) it depends on.
```

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,226 @@
...
+
+```
+App (Xcode project or appPkg)
+  |— Game (gamePkg)
+      |— Engine (gamePkg)
+```
+
+Here are source code examples.
```
rjmccall, line 32: Putting this sentence makes it feel like you're introducing new examples.  What you're actually doing is continuing the scenario you just started.  The best way to express this, I think, is to skip the little tree above, because honestly the dependency is so simple that you're distracting from the example.  Just describe the dependency in the prose that starts the example, then dive into the code.

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,226 @@
...
+
+```
+[Engine]
```
rjmccall, line 35: These headers are metatextual, not source code, so you can add more text to reinforce the structure.  Maybe `[Engine (a module in gamePkg)]`?

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,226 @@
...
+In the above scenario, App can import Engine (a utility module in 'gamePkg') and access its helper API directly, even though the API is not intended to be used outside of its package.
+
+These costs are particularly unfortunate given that packages are a unit of code distribution. Boundaries between packages often reflect divisions between development teams; different packages are developed on different schedules by different groups of people. These boundaries are therefore especially important to enforce, because accidental dependencies can require extra work and cooperation to detangle.
+
+
+## Proposed solution
+
+Our goal is to introduce a mechanism to Swift to recognize a package as a unit in the aspect of access control. We proposed to do so by introducing a new access modifier `package`. The `package` access modifier will allow accessing symbols from outside of its defining module as long as they are within the same package. This will help set clear boundaries between packages. 
+
+This also paves the path to include a package name in the namespace in the future, which will help disambiguate modules with the same name.
```
rjmccall, line 73:
```suggestion
This also paves the path to including a package name in the namespace in the future, which will help disambiguate modules with the same name.  
```

But really I'm not sure this belongs here.  It doesn't *really* pave this path, either; you can think of it as the entire utility module being package-private, but that's not going to be explicit in the source anywhere, so it doesn't actually rely on this feature existing.

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,226 @@
...
+
+Our goal is to introduce a mechanism to Swift to recognize a package as a unit in the aspect of access control. We proposed to do so by introducing a new access modifier `package`. The `package` access modifier will allow accessing symbols from outside of its defining module as long as they are within the same package. This will help set clear boundaries between packages. 
+
+This also paves the path to include a package name in the namespace in the future, which will help disambiguate modules with the same name.  
+
+## Detailed design
+
+### `package` keyword
+
+A new keyword `package` is introduced. It is a contextual keyword, i.e. a symbol can be named `package`.
```
rjmccall, line 79:
```suggestion
`package` is introduced as an access modifier.  It cannot be combined with other access modifiers.
```

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,226 @@
...
+For example, the following is allowed.
```
rjmccall, line 80:
```suggestion
`package` is a contextual keyword, so existing declarations named `package` will continue to work.  This follows the precedent of `open`, which was also added as a contextual keyword.  For example, the following is allowed:
```

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,282 @@
...
+
+Here's an example of how a package name is passed to a commandline invocation.
+
+```
+swiftc -module-name Engine -package-name gamePkg ...
+[Game] swiftc -module-name Game -package-name gamePkg ...
+[App] swiftc App -package-name appPkg ...
+```
+
+When building the Engine module, the package name 'gamePkg' and package symbols is stored in the Engine binary.  When building Game, the input to its package name 'gamePkg' is compared with Engine's package name; since they match, access to package symbols is allowed.  When building App, the name comparison shows 'appPkg' is different from its dependency module's so access to package symbols is deined, which is what we want.
```
rjmccall, line 137: The "binary" is not a language concept.  (It's also not a precise concept: the compiled code is a "binary", but the package name isn't part of that; also, we may want to record this in `swiftinterface`s eventually.)  Suggestion:

```suggestion
When building the Engine module, the package name 'gamePkg' is recorded in the built interface to the module.  When building Game, its package name 'gamePkg' is compared with the package name recorded in Engine's built interface; since they match, Game is allowed to access Engine's `package` declarations.  When building App, its package name 'appPkg' is different from `gamePkg`, so it is not allowed to access `package` symbols in either Engine or Game, which is what we want.
```

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,282 @@
...
+
+Swift Package Manager has a package identity per package, an identifier that's verified to be unique via a registry, and it will pass it down automatically.  Other build systems such as Bazel can introduce a new build setting to set the value for a package name.  Since it needs to be unique, a reverse-DNS name could be used to avoid clashing.
+
+### Exportability
+
+Package symbols will be stored in .swiftmodule only, and not in the public interface file (.swiftinterface).  We plan to introduce a .package.swiftinterface that contains package symbols, similar to a .private.swiftinterface which contains SPI symbols.  We will store information on whether the .swiftmodule was built from a .package.swiftinterface into the .swiftmodule.  We will also store the package name in both .swiftmodule and .swiftinterface.  This is to enforce loading a .swiftmodule instead of .swiftinterface when building modules in the same package.
```
rjmccall, line 143: This is very implementation-focused, and in theory Swift is an abstract language, not a specific implementation.  I'm not going to say this content is inappropriate for an evolution proposal, but you should try to be clear about the parts of the proposal that are specific to the primary Swift implementation.  I've made some suggestions below.

Also, I think this would be more appropriate in the previous section.

Also, please don't do this "we plan to" stuff.  If it's part of the proposal, it's part of the proposal.  If it's not, it's a future direction.

```suggestion
When the Swift frontend builds a `.swiftmodule` file directly from source, the file will include the package name and all of the `package` declarations in the module.  When the Swift frontend builds a `.swiftinterface` file from source, the file will include the package name, but it will put `package` declarations in a secondary `.package.swiftinterface` file.  When the Swift frontend builds a `.swiftmodule` file from a `.swiftinterface` file that includes a package name, but it does not have the corresponding `.package.swiftinterface` file, it will record this in the `.swiftmodule`, and it will prevent this file from being used to build other modules in the same package.
```

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,282 @@
...
+
+### Exportability
+
+Package symbols will be stored in .swiftmodule only, and not in the public interface file (.swiftinterface).  We plan to introduce a .package.swiftinterface that contains package symbols, similar to a .private.swiftinterface which contains SPI symbols.  We will store information on whether the .swiftmodule was built from a .package.swiftinterface into the .swiftmodule.  We will also store the package name in both .swiftmodule and .swiftinterface.  This is to enforce loading a .swiftmodule instead of .swiftinterface when building modules in the same package.  
+
+`package` functions can be made `@inlinable`.  Just like with `@inlinable public`, not all symbols are usable within the function: they must be `open`, `public`, `package`, or `@usableFromInline`.  Note that `@usableFromInline` allows the use of a symbol from `@inlinable` functions whether they're `package` or `public`.  `@usableFromPackageInline` will be introduced to export a symbol only for use by `@inlinable package` functions.
+
+We plan to introduce an option to hide package symbols.  By default, all package symbols will be exported in the final library/executable.
```
rjmccall, line 147: "we plan to" again

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,282 @@
...
+
+### Exportability
+
+Package symbols will be stored in .swiftmodule only, and not in the public interface file (.swiftinterface).  We plan to introduce a .package.swiftinterface that contains package symbols, similar to a .private.swiftinterface which contains SPI symbols.  We will store information on whether the .swiftmodule was built from a .package.swiftinterface into the .swiftmodule.  We will also store the package name in both .swiftmodule and .swiftinterface.  This is to enforce loading a .swiftmodule instead of .swiftinterface when building modules in the same package.  
+
+`package` functions can be made `@inlinable`.  Just like with `@inlinable public`, not all symbols are usable within the function: they must be `open`, `public`, `package`, or `@usableFromInline`.  Note that `@usableFromInline` allows the use of a symbol from `@inlinable` functions whether they're `package` or `public`.  `@usableFromPackageInline` will be introduced to export a symbol only for use by `@inlinable package` functions.
```
rjmccall, line 145: "will be introduced"

rjmccall (changes requested):

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,281 @@
...
+
+### Subclassing and Overrides
+
+Access control in Swift usually doesn't distinguish between different kinds of use.  If a program has access to a type, for example, that gives the programmer a broad set of privileges: the type name can be used in most places, values of the type can be borrowed, copied, and destroyed, members of the type can be accessed (up to the limits of their own access control), and so on.  This is because access control is a tool for enforcing encapsulation and allowing the future evolution of code.  Broad privileges are granted because restricting them more precisely usually doesn't serve that goal.
+
+However, there are two exceptions.  The first is that Swift allows `var` and `subscript` to restrict mutating accesses more tightly than read-only accesses; this is done by writing a separate access modifier for the setter, e.g. `private(set)`.  The second is that Swift allows classes and class members to restrict subclassing and overriding more tightly than normal references; this is done by writing `public` instead of `open`.  Allowing these privileges to be separately restricted does serve the goal of encapsulation and evolution.
```
rjmccall, line 156:
```suggestion
However, there are two exceptions.  The first is that Swift allows `var` and `subscript` to restrict mutating accesses more tightly than read-only accesses; this is done by writing a separate access modifier for the setter, e.g. `private(set)`.  The second is that Swift allows classes and class members to restrict subclassing and overriding more tightly than normal references; this is done by writing `public` instead of `open`.  Allowing these privileges to be separately restricted serves the goals of promoting encapsulation and evolution.
```

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,281 @@
...
+
+Because setter access levels are controlled by writing a separate modifier from the primary access, the syntax naturally extends to allow `package(set)`.  However, subclassing and overriding are different because they are controlled by writing a specific keyword as the primary access modifier.  This proposal has to decide what `package` by itself means for classes and class members.  It also has to decide whether to support the options not covered by `package` alone or to leave them as a possible future direction.
+
+Here is a matrix showing where each current access level can be used or overridable:
```
rjmccall, line 160:
```suggestion
Here is a matrix showing where symbols with each current access level can be used or overridden:
```

I'm sorry, but I think my matrix layout (with the possible places you might want to subclass / override on each axis and the keywords in the cells) was a lot clearer, especially about implying that it was exhaustive.  Can you use that instead?

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,281 @@
...
+
+However, there are two exceptions.  The first is that Swift allows `var` and `subscript` to restrict mutating accesses more tightly than read-only accesses; this is done by writing a separate access modifier for the setter, e.g. `private(set)`.  The second is that Swift allows classes and class members to restrict subclassing and overriding more tightly than normal references; this is done by writing `public` instead of `open`.  Allowing these privileges to be separately restricted does serve the goal of encapsulation and evolution.
+
+Because setter access levels are controlled by writing a separate modifier from the primary access, the syntax naturally extends to allow `package(set)`.  However, subclassing and overriding are different because they are controlled by writing a specific keyword as the primary access modifier.  This proposal has to decide what `package` by itself means for classes and class members.  It also has to decide whether to support the options not covered by `package` alone or to leave them as a possible future direction.
```
rjmccall, line 158:
```suggestion
Because setter access levels are controlled by writing a separate modifier from the primary access, the syntax naturally extends to allow `package(set)`.  However, subclassing and overriding are controlled by choosing a specific keyword (`public` or `open`) as the primary access modifier, so the syntax does not extend to `package` the same way.  This proposal has to decide what `package` by itself means for classes and class members.  It also has to decide whether to support the options not covered by `package` alone or to leave them as a possible future direction.
```

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,281 @@
...
+
+## Effect on ABI stability
+
+Boundaries between separately-built modules within a package are still potentially ABI boundaries.  The ABI for package symbols is not different from the ABI for public symbols, although in the future we plan to add an option to not export package symbols that can be resolved within an image.  
+
+## Alternatives considered
+
+### Use Current Workarounds
+A current workaround for the scenario in the Motivation is to use `@_spi` or `@_implemenationOnly`.  Each option has caveats.  The `@_spi` requires a group name, which makes it verbose to use and harder to keep track of, and `@_implementationOnly` can be too limiting as we want to be able to restrict access to only portions of APIs.  There are also hacky workarounds such as `@testable` and the `-disable-access-control` flag.  They elevate all `internal` (and `private` with the flag) symbols to `public`.  These options are unstable and also quickly lead to an increase of the binary and the shared cache size, not to mention symbol name clashes.
```
rjmccall, line 268:
```suggestion
### `@_spi`

One workaround for the scenario in the Motivation would be to use the `@_spi(groupName)` attribute, which allows part of the API of a module to be hidden unless it is imported in a special way that explicitly requests access to it.  This is an unsatisfying alternative to package-level access control because it is designed around a very different situation.  An SPI is a "hole" in the normal public interface, one meant for the use of a specific client.  That client is typically outside of the module's normal code-distribution boundary, but the module authors still have a cooperative working relationship.  This relationship is reflected in the design of `@_spi` in multiple ways:

- First, access to the SPI is granted to a specific client by name.  This is a clear and unavoidable communication of intent about who is meant to be using the SPI.  Other clients can still pose as this client and use the SPI, but that would be a clear breach of trust with predictable consequences.

- Second, clients must explicitly request the SPI by name.  This means that clients must opt in to using the SPI in every file, which works to limit its accidental over-use even by the intended client.  It also means that SPI use is obvious in the code, which code reviewers can see and raise questions about, and which SPI authors can easily find with a code search.

The level of care implied by these properties is appropriate for a carefully-targeted hole in an API that must cross a code-distribution boundary and will therefore require equal amounts of care to ever modify or close.  That rarely applies to two modules within the same package, where a package-level interface can ideally be changed with just a quick edit to a few different parts of a repository.  The `@_spi` attribute is intentionally designed to not be as lightweight as a package-local change should be.

`@_spi` would also not be easy to optimize.  By design, clients of an SPI can be anywhere, making it effectively part of the public ABI of a module.  To avoid exporting an SPI, the build system would have to know about that specific SPI group and promise the compiler that it was only used in the current built image.  Recognizing that all of the modules in a package are being linked into the same image and can be optimized together is comparatively easy for a build system and so is a much more feasible future direction.

### `@_implementationOnly`

Another workaround for the scenario in the Motivation is to use the `@_implementationOnly` attribute on the import of a module.  This attribute causes the module to be imported only for the use of the current module; clients of the current module don't implicitly transitively import the target module, and the symbols of the target module are restricted from appearing in the `public` API of the current module.  This would prevent clients from accidentally using APIs from the target module.  However, this is a very incomplete workaround for the lack of package-level access control.  For one, it doesn't actually prevent access to the module, which can still be explicitly imported and used.  For another, it only works on an entire module at a time, so a module cannot restrict some of its APIs to the package while making others available publicly.  Taming transitive import would be a good future direction for Swift, but it does not solve the problems of package-level APIs.

### Other workarounds

There are a few other workarounds to the absence of package-level access control, such as using `@testable` or the `-disable-access-control` flag.  These are hacky subversions of Swift's language design, and they severely undermine the use of module boundaries for encapsulation.  `-disable-access-control` is also an unstable and unsupported feature that can introduce build failures by causing symbol name collisions.
```

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,326 @@
...
+* Proposal: [SE-NNNN](NNNN-package-access-modifier.md)
+* Authors: [Ellie Shin](https://github.com/elsh), [Alexis Laferriere](https://github.com/xymus)
+* Review Manager: [John McCall](https://github.com/rjmccall)
+* Status: **Awaiting review**
+* Implementation: [apple/swift#61546](https://github.com/apple/swift/pull/62700), [apple/swift#62704](https://github.com/apple/swift/pull/62704), [apple/swift#62652](https://github.com/apple/swift/pull/62652), [apple/swift#62652](https://github.com/apple/swift/pull/62652)
+* Review: ([pitch](https://forums.swift.org/t/new-access-modifier-package/61459))
+
+## Introduction
+
+This proposal introduces `package` as a new access modifier.  Currently, to access a symbol in another module, that symbol needs to be declared `public`.  However, a symbol being `public` allows it to be accessed from any module at all, both within a package and from outside of a package, which is sometimes undesirable.  We need a new access modifier to enable a more fine control over the visibility scope of such symbols.
```
rjmccall, line 12:
```suggestion
This proposal introduces `package` as a new access modifier.  Currently, to access a symbol in another module, that symbol needs to be declared `public`.  However, a symbol being `public` allows it to be accessed from any module at all, both within a package and from outside of a package, which is sometimes undesirable.  We need a new access modifier to enable more control over the visibility scope of such symbols.  
```

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,326 @@
...
+
+Packages, as expressed by the Swift Package Manager, are a unit of code distribution.  Some packages contain just a single module, but it's frequently useful to split a package's code into multiple modules.  For example, when a module contains some internal helper APIs, those APIs can be split out into a utility module and maybe reused by other modules or packages.
+
+However, because Swift does not recognize organizations of code above the module level, it is not possible to create APIs like this that are purely internal to the package.  To be usable from other modules within the package, the API must be public, but this means it can also be used outside of the package.  This allows clients to form unwanted source dependencies on the API.  It also means the built module has to export the API, which has negative implications for code size and performance.
+
+For example, here’s a scenario where a client has access to a utility API from a package it depends on.  The client `App` could be an executable or an Xcode project.  It depends on a package called `gamePkg`, which contains two modules, `Game` and `Engine`.  
+
+Here are source code examples.
```
rjmccall, line 26:
```suggestion
```

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,326 @@
...
+
+```
+swiftc -module-name Engine -package-name gamePkg ...
+swiftc -module-name Game -package-name gamePkg ...
+swiftc App -package-name appPkg ...
+```
+
+When building the `Engine` module, the package name 'gamePkg' is recorded in the built interface to the module.  When building `Game`, its package name 'gamePkg' is compared with the package name recorded in `Engine`'s built interface; since they match, `Game` is allowed to access `Engine`'s `package` declarations.  When building `App`, its package name 'appPkg' is different from `gamePkg`, so it is not allowed to access `package` symbols in either `Engine` or `Game`, which is what we want.
+
+The Swift Package Manager already has a concept of a package idpackage identity string for every package, as specified by [SE-0292](https://github.com/apple/swift-evolution/blob/main/proposals/0292-package-registry-service.md).  This string is verified to be unique via a registry, and it always works as a package name, so SwiftPM will pass it down automatically.  Other build systems such as Bazel may need to introduce a new build setting for a package name.  Since it needs to be unique, a reverse-DNS name may be used to avoid clashing.
```
rjmccall, line 135:
```suggestion
The Swift Package Manager already has a concept of a package identity string for every package, as specified by [SE-0292](https://github.com/apple/swift-evolution/blob/main/proposals/0292-package-registry-service.md).  This string is verified to be unique via a registry, and it always works as a package name, so SwiftPM will pass it down automatically.  Other build systems such as Bazel may need to introduce a new build setting for a package name.  Since it needs to be unique, a reverse-DNS name may be used to avoid clashing.
```

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,326 @@
...
+
+When the Swift frontend builds a `.swiftmodule` file directly from source, the file will include the package name and all of the `package` declarations in the module.  When the Swift frontend builds a `.swiftinterface` file from source, the file will include the package name, but it will put `package` declarations in a secondary `.package.swiftinterface` file.  When the Swift frontend builds a `.swiftmodule` file from a `.swiftinterface` file that includes a package name, but it does not have the corresponding `.package.swiftinterface` file, it will record this in the `.swiftmodule`, and it will prevent this file from being used to build other modules in the same package.
+
+### Package Symbols and `@inlinable` Functions
+
+`package` functions can be made `@inlinable`.  Just like with `@inlinable public`, not all symbols are usable within the function: they must be `open`, `public`, `package`, or `@usableFromInline`.  Note that `@usableFromInline` allows the use of a symbol from `@inlinable` functions whether they're `package` or `public`.  `@usableFromPackageInline` is introduced to export a symbol only for use by `@inlinable package` functions.
```
rjmccall, line 143: Looks like there's an editorial mistake here — there's a lot of redundancy between this paragraph and the next.

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,326 @@
...
+
+`package` functions can be made `@inlinable`.  Just like with `@inlinable public`, not all symbols are usable within the `@inlinable package` function: they must be `open`, `public`, `package`, `@usableFromInline`, or `usableFromPackageInline`.
```
rjmccall, line 145: Missing newline between paragraphs (if you keep this paragraph).

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,326 @@
...
+@inlinable public func publicUse() {
+    internalFuncA() // ERROR
+    internalFuncB() // OK
+    packageFunc() // OK
+}
+```
+ 
+### Resiliency
+
+Library evolution makes modules resilient.  We can incorporate a package name into a resiliency check and bypass it if modules are in the same package.  This will remove the need for resilience overhead such as indirection and language requirements such as `@unknown default` for an unfrozen `enum`.
```
rjmccall, line 182: I'm sorry for overlooking this section in earlier passes.  The "we can" makes it unclear whether you're actually proposing this as part of this proposal.  If you're proposing it, you should say so affirmatively, like "The Swift frontend will assume that modules defined in the same package will always be rebuilt together and do not require a resilient ABI boundary between them.  This removes the need for performance and code size overhead introduced by ABI-resilient code generation, and it also eliminates language requirements such as `@unknown default` for a non-`frozen` `enum`."  However, I suspect you mean this to just be a future direction, right?

proposals/NNNN-package-access-modifier.md
```diff
@@ -0,0 +1,326 @@
...
+
+Because setter access levels are controlled by writing a separate modifier from the primary access, the syntax naturally extends to allow `package(set)`.  However, subclassing and overriding are controlled by choosing a specific keyword (`public` or `open`) as the primary access modifier, so the syntax does not extend to `package` the same way.  This proposal has to decide what `package` by itself means for classes and class members.  It also has to decide whether to support the options not covered by `package` alone or to leave them as a possible future direction.
+
+Here is a matrix showing where symbols with each current access level can be used or overridden:
+
+<table>
+<thead>
+<tr>
+<th>Subclassable in... \ Accessible in...</th>
```
rjmccall, line 197: Hmm.  I get what the slash is aiming to do, but I think it would be easier to read to just make the row labels be "Subclassable anywhere" etc. and the column labels "Accessible anywhere" etc.  Alternatively, you could put the prefixes in cells that use `rowspan` and `colspan`, if the markdown processing doesn't mess those up.

rjmccall (commented):
Okay.  This should be the final pass.  Looking very close to ready!

rjmccall (approved):

PR #1889, opened 2022-12-21
[SE-0381] discardResults becomes [Throwing]DiscardingTaskGroup

ktoso:
According to SE review, we are introducing a new type after all for this functionality.

This updates the proposal to reflect this, along with more discussion about the implicit error handling.


ktoso:
Discussed with @DougGregor earlier that we can amend and keep this review going 👍

PR #1896, opened 2023-01-03
[SE-0368] StaticBigInt: remove prefix `+` operator

benrimmington:


stephentyrone:
We’re going to run a mini-proposal through evolution for this, so let’s leave the proposal as-is for now.

stephentyrone:
(In the meantime, we’ll remove the actual operator. I’ll tag you to review that for main and 5.8.)

hborla:
Hey @benrimmington and @stephentyrone! I'm going to kick off this change for review as an amendment to SE-0368. Steve, shall we close https://github.com/apple/swift-evolution/pull/1924?

hborla:
@benrimmington We always run proposal revisions through a short review. We can certainly help provide an explanation of the source break for the proposal text! I suggest taking this modified prose from @stephentyrone 's proposal and incorporating it into the Alternatives Considered section of this proposal

---

A previous version of this proposal included the following operation on `StaticBigInt`:
```swift
/// Returns the given value unchanged.
public static prefix func + (_ rhs: StaticBigInt) -> StaticBigInt
```
This was included so that prefix `+` could be included as a literal prefix in contexts where the type is `StaticBigInt`, rather than literals of another type, for symmetry with `-`:
```swift
let signs: [StaticBigInt] = [-1, 0, +1]
```

It turns out that this was an accidental source breaking change, because of examples like the following:
```swift
let a: Int = 7
let b = +1     // Inferred as `StaticBigInt` because concrete `+` beats
               // the generic one on `AdditiveArithmetic`
let c = a + b  // Error: Cannot convert `b` from `StaticBigInt` to `Int`
```

Previously, `b` was given type `Int`, and this example compiled correctly. However, with the addition of a concrete `+` overload on `StaticBigInt`, the generic `+` that produced type `Int` is no longer chosen, because concrete overloads are always preferred over generic overloads.

stephentyrone:
What Holly said. I'll also help answer any questions in the actual review thread.

PR #1902, opened 2023-01-04
Amend SE-0378

yim-lee:
- Per review feedback, provide a way for user to override default credential store and use netrc file instead. (`--netrc`)
- Replace "`.netrc`" with "netrc file" since the filename doesn't have to be `.netrc`.
- Update `login` subcommand outputs to match implementation.


yim-lee:
cc @tomerd

tomerd (approved):

PR #1908, opened 2023-01-06
[Pitch] Result builder scoped unqualified lookup

angela-laar:
I will incorporate community feedback in a subsequent commits.


rjmccall:
This has stalled out, and the LSG is not enthusiastic about the idea. It may be largely subsumed by macros at this point.

PR #1912, opened 2023-01-17
[SE-0382] Review updates to simplify and protocol-ize MacroExpansionContext

DougGregor:
Generalize `MacroExpansionContext` by making it a class-bound protocol (since the compiler would prefer to have an implementation different from that of the test harness), and remove `moduleName`/`fileName`.


DougGregor:
Ditching this for the more-complete and appropriately-branch-named https://github.com/apple/swift-evolution/pull/1928

PR #1918, opened 2023-01-21
Remove if/switch expressions from common rejections

Jumhyn:
Obsoleted by [SE-0380](https://github.com/apple/swift-evolution/blob/main/proposals/0380-if-switch-expressions.md) :)


hborla (approved):

stephentyrone (approved):
Make it so!

PR #1924, opened 2023-01-24
Proposal to remove + on StaticBigInt

stephentyrone:


proposals/0000-formally-remove-bigint-plus.md
```diff
@@ -0,0 +1,60 @@
...
+```
+/// Returns the given value unchanged.
+public static prefix func + (_ rhs: Self) -> Self
+```
+This was included so that prefix `+` could be included as a literal prefix in contexts where the type is StaticBigInt, rather than literals of another type, for symmetry with `-`:
+```
+let signs: [StaticBigInt] = [-1, 0, +1]
+```
+
+It turns out that this was a accidental source breaking change, because of examples like the following:
```
stephentyrone, line 20:
```suggestion
It turns out that this was an accidental source breaking change, because of examples like the following:
```

proposals/0000-formally-remove-bigint-plus.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Effect on API resilience
+
+None.
+
+## Alternatives considered
+
+We could keep prefix `+` around but implement it in a different fashion.
+However, at this point there is little value in doing so.
+No existing Swift programs need it, and contexts where it would make an actual difference are few and far between; it will only ever matter when you explicitly want to have values of type `StaticBigInt`, which is rare.
```
stephentyrone, line 59:
```suggestion
No existing Swift programs need it, and contexts where it would make any difference are few and far between; it will only ever matter when you explicitly want to have values of type `StaticBigInt`, which is rare.
```

proposals/0000-formally-remove-bigint-plus.md
```diff
@@ -0,0 +1,60 @@
...
+## Effect on API resilience
+
+None.
+
+## Alternatives considered
+
+We could keep prefix `+` around but implement it in a different fashion.
+However, at this point there is little value in doing so.
+No existing Swift programs need it, and contexts where it would make any difference are few and far between; it will only ever matter when you explicitly want to have values of type `StaticBigInt`, which is rare.
+If it turns out that people have significant use cases for this operation in the future, we can consider a proposal to add it via some other mechanism.
```
stephentyrone, line 60:
```suggestion
If it turns out that people have need for this operation in the future, we can consider a proposal to add it via some other mechanism that is not source-breaking.
```

xwu:
@stephentyrone For concreteness, I can share a full implementation of the alternative considered :)

stephentyrone:
I don't think that the exact mechanism we would choose is necessary for this proposal, but if it comes up in discussion we can add the options as a note.

PR #1932, opened 2023-01-27
Attached macros

DougGregor:


proposals/nnnn-attached-macros.md
```diff
@@ -0,0 +1,601 @@
...
+
+#### Peer macros
+
+Peer macros produce new declarations alongside the declaration to which they are attached.  For example, here is a declaration of a macro that introduces a completion-handler version of a given asynchronous function:
+
+```swift
+@attached(peer, names: overloaded) macro addCompletionHandler()
+```
+
+This macro uses the `@attached` attribute to indicate that it is an attached peer attribute. The `names` argument specifies that how the names of the peer declarations are created, using an extended form of the scheme introduced with [freestanding macros](https://github.com/DougGregor/swift-evolution/blob/freestanding-macros/proposals/nnnn-freestanding-macros.md#up-front-declarations-of-newly-introduced-names). In this case, `overloaded` means that this macro will produce a peer that is overloaded with the declaration to which it is attached, i.e., it has the same base name.
```
allevato, line 49: Here and elsewhere, if we start this review before freestanding macros, I wonder if we should move the linked content for the naming scheme into a section of this proposal. Alternatively, it could go into the vision and both proposals could link to it?

allevato, line 49: Since functions can be distinguished by their full decl name with argument labels, what is the significance of telling the compiler that the peer declaration has the same base name? Can you expand on what the compiler does with this information?

proposals/nnnn-attached-macros.md
```diff
@@ -0,0 +1,601 @@
...
+  static var b = MyOptions(rawValue: 1 << Option.b.rawValue)
+  static var c = MyOptions(rawValue: 1 << Option.c.rawValue)
+}
+```
+
+The macro itself will be declared as a member macro that defines an arbitrary set of members:
+
+```swift
+/// Create the necessary members to turn a struct into an option set.
+@attached(member, names: names(rawValue), arbitrary]) macro optionSetMembers
```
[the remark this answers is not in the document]
DougGregor: Right on both counts, thank you!

proposals/nnnn-attached-macros.md
DougGregor: I've pulled in all of the linked content so this proposal is... freestanding. Also, added more detail

DougGregor: The compiler tends to ignore the labels when it does name lookup. I've added some more text about how this information is used.

proposals/nnnn-attached-macros.md
```diff
@@ -0,0 +1,601 @@
...
+    of node: AttributeSyntax,
+    providingAccessorsOf declaration: some DeclSyntaxProtocol,
+    in context: some MacroExpansionContext
+  ) async throws -> [AccessorDeclSyntax]
+}
+```
+
+The implementation of the `dictionaryStorage` macro would create the accessor declarations shown above, using either the `key` argument (if present) or deriving the key name from the property name. The effect of this macro isn't something that can be done with a property wrapper, because the property wrapper wouldn't have access to `self.storage`.
+
+The presence of an accessor macro on a stored property removes the initializer. It's up to the implementation of the accessor macro to either diagnose the presence of the initializer (if it cannot be used) or incorporate it in the result.
```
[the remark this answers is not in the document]
DougGregor: Right, good catch. This was a hole in the design; I've updated the proposal to address this.

proposals/nnnn-attached-macros.md
```diff
@@ -0,0 +1,601 @@
...
+
+As with freestanding macros, attached declaration macros are declared with `macro`, and have [type-checked macro arguments](https://github.com/apple/swift-evolution/blob/main/proposals/0382-expression-macros.md#type-checked-macro-arguments-and-results) that allow their behavior to be customized. Attached macros are identified with the `@attached` attribute, which also provides the specific role as well as [any names they introduce](https://github.com/DougGregor/swift-evolution/blob/freestanding-macros/proposals/nnnn-freestanding-macros.md#up-front-declarations-of-newly-introduced-names). For example, the aforemented macro emulating property wrappers could be declared as follows:
+
+```swift
+@attached(peer, names: prefixed(_))
```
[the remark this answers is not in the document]
DougGregor: Yes, thank you!

allevato (approved):
Thanks! I'm going to start the review momentarily.

PR #1943, opened 2023-02-07
[SE-0382] Fix link

ole:
The link to the second pitch thread should point to the top of the thread, not to a specific post.


xwu:
Thanks!

PR #1949, opened 2023-02-14
Reflection

Azoy:


hborla:
Thank you for taking on the initial design and implementation work for richer reflection in Swift!

I've read through the pitch discussion, and I took note of the following constructive areas surfaced by reviewers that need to be addressed:
* Alternative designs to a top-level `Type` struct, such as:
  * An alternative name to avoid confusing the user-facing concept of a "type" with `Reflection.Type` (https://forums.swift.org/t/pitch-reflection/61438/48, https://forums.swift.org/t/pitch-reflection/61438/24)
  * A `Type` protocol with a family of conforming types to describe the difference kinds of types that appear in the language, such as `StructType`, `ClassType`, `FunctionType`, `ExistentialType`, etc, which may each have different specific queries. (https://forums.swift.org/t/pitch-reflection/61438/52)
* Alternative approaches to `ParitalType`, such as:
  * Alternative names for `PartialType`. (https://forums.swift.org/t/pitch-reflection/61438/15)
  * Including the functionality of `PartialType` as methods on `Type` rather than in a separate type. (https://forums.swift.org/t/pitch-reflection/61438/29)
* Requests to include a more extensive vision / future directions section about the future evolution of the Reflection API and its interaction with other language features, such as static reflection and macros (https://forums.swift.org/t/pitch-reflection/61438/37, https://forums.swift.org/t/pitch-reflection/61438/38)

The only topic above that I see mentioned in the Alternatives Considered section is `Type` as a protocol. However, I find the arguments presented against this approach unconvincing, as it solves a number of issues brought up during the pitch discussion, such as providing information about what kind of type you're working with, and surfacing a richer representation of the Swift type system.

I think the above areas need more discussion in the proposal. This proposed Reflection library assumes a very critical role in the Swift ecosystem by providing introspection into the Swift type system; the initial API surface warrants careful consideration of the design and alternatives to ensure that we're leaving space to pursue missing functionality in future proposals!

proposals/NNNN-reflection.md
```diff
@@ -0,0 +1,704 @@
...
+  /// Returns a collection wrapper over 'StoredProperty'
+  @inlinable
+  public var storedProperties: StoredProperties { get }
+
+  /// Returns a collection wrapper over the generic arguments needed to
+  /// construct this concrete 'Type'. The collection will always be empty for
+  /// types who are not structs, classes, or enums as well as types who are not
+  /// generic.
+  @inlinable
+  public var genericArguments: GenericArguments { get }
```
hborla, line 239: I don't see a description of the `GenericArguments` type in the proposal. Could you please include more details about this type, such as what its `Element` type is? I assume the elements are also `Type`s but that may not be obvious to a reviewer.

proposals/NNNN-reflection.md
```diff
@@ -0,0 +1,704 @@
...
+
+  /// Converts the type back to a Swift metatype.
+  @inlinable
+  public var swiftType: Any.Type { get }
+}
+
+extension Type {
+  /// Returns a collection wrapper over the function parameter types.
+  @inlinable
+  public var functionParameters: FunctionParameters
```
hborla, line 262: I notice a design choice here that isn't explicitly mentioned; all collections use a separate collection wrapper type, e.g. `Cases`, `GenericArguments`, `FunctionParameters`, etc, rather than using a common type parameterized over the `Element`. Could you please add some justification for why these types are distinct?

proposals/NNNN-reflection.md
```diff
@@ -0,0 +1,704 @@
...
+
+`Reflection` will be a new module included in the Swift toolchain. In order to
+use this module, one must explicitly declare an `import Reflection`.
+
+### `Type`
+
+This is our main entry point into the reflection module. `Type` represents all
+possible concrete types that can occur in Swift. Ranging from structs, enums,
+tuples, functions, existentials, etc. Anything that can be on the right hand
+side of an `as` is represented here in `Type`.
```
hborla, line 208: A subtle but important design choice here is that the `Type` type must grow to contain all information about any kind of type that a programmer might want to query, even if that information is not applicable to the underlying type at runtime; for example, you can access a function parameter list for all types, even if the underlying type is not a function type. I think it's important to explicitly underscore this design choice here in the opening design discussion of `Type` and justify why that's the right choice.

proposals/NNNN-reflection.md
```diff
@@ -0,0 +1,704 @@
...
+for child in Mirror(reflecting: sparky).children {
+  let fieldName = child.label
+  let fieldValue = child.value
+  let fieldType = type(of: fieldValue)
+}
+```
+
+This works great, but there is a serious issue here in that it requires an
+instance of the type you want to introspect over. Say you want to register types
+and for each field who conforms to my special protocol `CustomRegister`, cache
```
PsychoH13, line 48:
```suggestion
and for each field which conforms to my special protocol `CustomRegister`, cache
```

proposals/NNNN-reflection.md
```diff
@@ -0,0 +1,704 @@
...
+  @inlinable
+  public var cases: Cases { get }
+
+  /// Returns a collection wrapper over 'StoredProperty'
+  @inlinable
+  public var storedProperties: StoredProperties { get }
+
+  /// Returns a collection wrapper over the generic arguments needed to
+  /// construct this concrete 'Type'. The collection will always be empty for
+  /// types who are not structs, classes, or enums as well as types who are not
```
PsychoH13, line 236:
```suggestion
  /// types which are not structs, classes, or enums as well as types which are not
```

proposals/NNNN-reflection.md
```diff
@@ -0,0 +1,704 @@
...
+  /// Returns the 'PartialType' associated with this type, if it has one. A type
+  /// always has a partial type if it's representing a struct, enum, or class
+  /// type. Any other type never has a partial type.
+  @inlinable
+  public var partial: PartialType? { get }
+
+  /// If this type is representing a class type, return the superclass type if
+  /// it has one.
+  @inlinable
+  public var superclass: Type? { get }
```
PsychoH13, line 252: Since we can know the superclass of a type, should we also have access to the protocols the type conforms to?

Something like:
```Swift
@inlinable
public var protocols: [Type] { get }
```

rjmccall:
This seems to have stalled out, so I'm going to close it pending further development.

PR #1951, opened 2023-02-15
Update stream factory proposal

FranzBusch:
# Motivation

There were a few small mistakes in the async stream proposal that @benrimmington pointed out


beccadax:
Thanks!

PR #1967, opened 2023-03-03
Revise the async stream factory proposal to use a tuple instead and accept it

FranzBusch:


FranzBusch:
@benrimmington Thanks for the comments. Fixed everything up and updated the PR!

beccadax (approved):

beccadax:
I've added the changes needed to mark the proposal as accepted. Merging it now.

PR #1977, opened 2023-03-15
Amend SE-0391

yim-lee:
Amendments:
- Remove in-archive option for submitting package release metadata. The existing method, i.e. including metadata as part of the request body, will continue to be supported.
- By default SwiftPM will look for a file named `package-metadata.json` inside the package directory being published for package release metadata. The file will be included in the archive, AND its contents will be included as a separate section in the publish request body.
- If a source archive is being signed, then any submitted metadata will be signed also. Introduce a new part in the publish request body named `metadata-signature` to hold metadata signature.
- Remove `sign` subcommand because with `--dry-run` the `publish` subcommand can do essentially the same.
- Change `publish` subcommand to take cert chain (`cert-chain-paths`) rather than just the signing cert. In some cases the user needs to provide the cert chain such that the intermediate CA(s) get included in the signature and can be used for signature validation later.
- Add section for changes to "download source archive" API--the response needs to include signature related reponse headers if source archive is signed.

### DIFF
--- a/proposals/0391-package-registry-publish.md
+++ b/proposals/0391-package-registry-publish.md
@@ -8,8 +8,14 @@
   * https://github.com/apple/swift-package-manager/pull/6101
   * https://github.com/apple/swift-package-manager/pull/6146
   * https://github.com/apple/swift-package-manager/pull/6159
-  * https://github.com/apple/swift-package-manager/pull/6184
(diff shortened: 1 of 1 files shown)


tomerd (approved):
nice

PR #1984, opened 2023-03-17
[SE-390] fixe an example, add omitted details about function casts

kavon:
I don't think these changes are controversial. Mostly just some minor omissions / clarifications.

### DIFF
--- a/proposals/0390-noncopyable-structs-and-enums.md
+++ b/proposals/0390-noncopyable-structs-and-enums.md
@@ -662,6 +662,11 @@ extension FileDescriptor {
 }
 ```
 
+Casts of function types that change the ownership modifier of a noncopyable 
+parameter are currently invalid. One reason is that it is impossible to cast a 
+function with a noncopyable `consuming` parameter, into one where that 
+parameter is `borrowed`. See Future Directions for details.
+
 ### Declaring properties of noncopyable type
 
 A class or noncopyable struct may declare stored `let` or `var` properties of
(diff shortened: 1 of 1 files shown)


tbkka (approved):

proposals/0390-noncopyable-structs-and-enums.md
```diff
@@ -1575,6 +1590,35 @@ We can expose the ability for code to implement these coroutines directly,
...
 more expressivity with noncopyable properties.
 
+### Casting function types based on ownership modifiers
+The current rules for casting functions with a noncopyable parameter is that
+the ownership modifier must remain the same. 
+
+One reason behind this limitation is a matter of scope. There is a broader need
+to support such casts even for copyable types. For example, a function
+containing an `inout` parameter cannot currently be cast to or from any other
+function type, unless if the parameter remains `inout`. But it should be safe to
```
glessard, line 1600:
```suggestion
function type, unless the parameter remains `inout`. It should be safe to
```

glessard (approved):

proposals/0390-noncopyable-structs-and-enums.md
glessard: The "cannot.... unless..." construction would be more readable by describing only what is currently supported, than describing what _could_ be supported. My suggestion doesn't achieve that.

kavon: Thanks. I'm just gonna drop the sentence / point since it's not adding much anyway.

proposals/0390-noncopyable-structs-and-enums.md
```diff
@@ -245,6 +245,9 @@ protocol Foo where Self: ~Copyable {
   // ERROR: associated type requirements must be `Copyable`
   associatedtype Bar: ~Copyable
 }
+
+// ERROR: cannot suppress `Copyable` in extension of `FileWithPath`
+extension FileWithPath: ~Copyable {}
```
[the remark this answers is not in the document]
kavon: hm. good point. While that syntax wouldn't currently be valid for `FileWithPath`, I'll leave this point about not allowing it in extensions "currently" as-is, since we're not sure how we want people to express that idea. Personally I think it should be fine for people to just say `extension Optional { ... }` and all methods in there apply to all Optionals, whether they're copyable or not. Thus, methods in there would get checks and errors about copying a noncopyable `self`, etc, since that's the most general kind of Optional.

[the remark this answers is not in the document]
kavon: Thus, I was thinking you have to explicitly exclude the most general kind of noncopyable optional via `extension Optional: Copyable { ... }` or something like that.

PR #1997, opened 2023-03-29
Update 'Feature Identifier' to 'Upcoming Feature Flag' in template

dempseyatgithub:
This PR updates the proposal template header field 'Feature Identifier' to 'Upcoming Feature Flag' to be consistent with existing proposals with upcoming feature flags as added in 70f9605.

The PR also updates the header field name to use title case instead of sentence case to be consistent with all other proposal header field names.


rjmccall:
Thanks!

PR #1998, opened 2023-03-29
Observation

phausler:
A new module along side the standard library and concurrency for making types observable powered by macros for easy to use annotations.


proposals/NNNN-observability.md
```diff
@@ -0,0 +1,655 @@
...
+    init() { }
+  
+    func purchase(alternateIcons: Bool, allRecipes: Bool) {
+        alternateIconsUnlocked = alternateIcons
+        allRecipesUnlocked = allRecipes
+    }
+}
+```
+
+When a property does not have a default value that corisponding argument in the initializer does not have a default value. This means that the following example has a macro synthesized initializer of `init(a: Int, b: Int = 3)`.
```
PsychoH13, line 322:
```suggestion
When a property does not have a default value that corresponding argument in the initializer does not have a default value. This means that the following example has a macro synthesized initializer of `init(a: Int, b: Int = 3)`.
```

proposals/NNNN-observability.md
```diff
@@ -0,0 +1,655 @@
...
+}
+```
+
+### `TrackedProperties`
+
+When observing changes to a type, there may be associated side effects to members that are not publicly visible. The `TrackedProperties` type allows for internal key paths to be included in a transaction without being exposed beyond their visibility. 
+
+```swift
+public struct TrackedProperties<Root>: ExpressibleByArrayLiteral, @unchecked Sendable {
+    public typealias Element = PartialKeyPath<Root>
```
PsychoH13, line 375: This declaration seems unnecessary? It seems to have been added to support SetAlgebra but the protocol conformance isn't there, and the alias isn't used otherwise.

proposals/NNNN-observability.md
```diff
@@ -0,0 +1,760 @@
...
+```swift
+@Observable final class InitializationSample {
+    var a: Int
+    var b: Int = 3
+}
+```
+
+Because the memberwise initializer and backing storage are generated together, the initializer is able to initialize that storage. User-defined initializers should call the generated memberwise initializer instead of attempting to initialize the properties directly.
+
+Due to limitations with macros, all fields must have type information; this restriction may be able to be lifted later when the type system grows a mechanism to detect inferred types.
```
airspeedswift, line 358:
```suggestion
Since macros are only syntactic transformations, all fields must have explicit type information. This restriction may be able to be lifted later when the type system grows a mechanism to detect inferred types.
```

proposals/0396-observability.md
```diff
@@ -0,0 +1,760 @@
...
+object.someOtherProperty += 1
+// nothing is printed
+```
+
+Observable objects can also provide changes grouped into transactions, which coalesce any changes that are made between suspension points. Transactions are accessed using the `changes(for:)` method, and are delivered isolated to an actor that you provide, or the main actor by default. Because transactions can cover changes to more than one property, the `changes(for:)` method provides a sequence of `TrackedProperties` instances, which can be queried to find the changed properties.
+
+```swift
+func processTransactions(_ object: MyObject) async {
+    for await change in objects.changes(for: [\.someProperty, \.someOtherProperty]) {
+        print(myObject.someProperty, myObject.someOtherProperty)
```
Dev1an, line 102: Where does `myObject` come from here? Is it possibly a typo and was it meant to be `object` instead?

phausler: it was.

PR #2002, opened 2023-04-04
Freestanding macros update

rxwei:
- Scope code item macros out as a future direction.
- Update `#warning` macro definition to make sure it compiles.
- Add more examples.


rxwei:
Wrong branch....

PR #2003, opened 2023-04-04
Proposal: Conform `Never` to `Codable`

natecook1000:
Pitch thread: https://forums.swift.org/t/pitch-conform-never-to-codable/64056/21

### DIFF
--- a/proposals/NNNN-never-codable.md
+++ b/proposals/NNNN-never-codable.md
@@ -35,7 +35,7 @@ The standard library should add `Encodable` and `Decodable` conformance to the `
 
 The `Encodable` conformance is simple — since it's impossible to have a `Never` instance, the `encode(to:)` method can simply be empty.
 
-The `Decodable` protocol requires the `init(from:)` initializer, which clearly can't create a `Never` instance. The implementation throws a `DecodingError` if decoding is attempted.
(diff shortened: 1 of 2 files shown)


proposals/NNNN-never-codable.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Proposed solution
+
+The standard library should add `Encodable` and `Decodable` conformance to the `Never` type.
+
+## Detailed design
+
+The `Encodable` conformance is simple — since it's impossible to have a `Never` instance, the `encode(to:)` method can simply be empty.
+
+The `Decodable` protocol requires the `init(from:)` initializer, which clearly can't create a `Never` instance. The implementation throws a `DecodingError` if decoding is attempted.
```
allevato, line 38: I think the point Joe made in [this post on the pitch thread](https://forums.swift.org/t/pitch-conform-never-to-codable/64056/10) would be a good one to include in the proposal text here, to explain why you're specifically throwing an error and not trapping.

Do you want to call out the specific instance of `DecodingError` that the implementation throws?

allevato (approved):
Looks great! Just one minor comment, but we can start the review for this one shortly.

proposals/NNNN-never-codable.md
natecook1000: Included both points, thank you!

PR #2007, opened 2023-04-10
Add proposal for `is case` expressions

matthewcheok:
This PR adds a proposal for pattern matching with `is case` expressions.

**Implementation**: https://github.com/apple/swift/pull/64901
**Swift-evolution thread**: [is case expressions](https://forums.swift.org/t/pitch-is-case-expressions/64185)

For example:
```swift
enum Destination {
  case inbox
  case messageThread(id: Int)
}

let destination = Destination.messageThread(id: 42)
print(destination is case .inbox) // false
print(destination is case .messageThread) // true
print(destination is case .messageThread(id: 0)) // false
print(destination is case .messageThread(id: 42)) // true

// SwiftUI view
VStack {
  HeaderView(inThread: destination is case .messageThread)
  ...
}
```


matthewcheok:
Closing as per [update](https://forums.swift.org/t/pitch-is-case-expressions/64185/69) in the pitch thread

PR #2010, opened 2023-04-12
fix typo in SE-0395

JCSooHwanCho:
I'm not sure this fix is correct, but current code snippent is wrong.


rjmccall:
I think Philippe fixed this as part of revising the proposal; closing, but please verify.

PR #2013, opened 2023-04-13
Pitch variadic generic types

slavapestov:


proposals/nnnn-variadic-types.md
```diff
@@ -0,0 +1,239 @@
...
+```
+
+If the generic parameter list of a variadic type consists of a single generic parameter pack and nothing else, it is possible to reference it with an empty generic argument list:
+
+```swift
+struct V<each T> {}
+
+V<>.self
+```
+Note that `V<>` is not the same as `V`. The former substitutes the generic parameter pack `T` with the empty pack. The latter does not constrain the pack at all and is only permitted in contexts where the generic argument can be inferred.
```
Jumhyn, line 105: This is sort of covered under 'inferred' I guess, but maybe worth being explicit about the 'bare name means `Self` within the generic type' rule here (I don't think this language is intended to _ban_ that behavior inside variadic generic types, right?)

proposals/nnnn-variadic-types.md
```diff
@@ -0,0 +1,239 @@
...
+## Detailed design
+
+Swift has the following kinds of generic type declarations:
+
+- Structs
+- Enums
+- Classes (and actors)
+- Type aliases
+
+A generic type is _variadic_ if it directly declares a type parameter pack with `each`, or if it is nested inside of another variadic type. In this proposal, structs, classes, actors and type aliases can be variadic. Enums will be addressed in a follow-up proposal.
```
Jumhyn, line 52: Might be nice to have a brief mention of why they're being treated separately, like we have for classes below?

proposals/nnnn-variadic-types.md
```diff
@@ -0,0 +1,239 @@
+# Allow Generic Types to Abstract Over Packs
+
+* Proposal: [SE-NNNN](nnnn-variadic-types.md)
+* Authors: [Slava Pestov](https://github.com/slavapestov), [Holly Borla](https://github.com/hborla)
+* Review Manager: TBD
+* Upcoming Feature Flag: `VariadicGenerics`
+* Previous Proposal: [SE-0393](0393-parameter-packs.md)
```
Jumhyn, line 7: Just checking, does the same line from SE-0393 apply here, i.e.:
> * Implementation: On main gated behind the frontend flag `-enable-experimental-feature VariadicGenerics`

?

proposals/nnnn-variadic-types.md
```diff
@@ -0,0 +1,239 @@
...
+typealias Element = (repeat each S.Element)
+typealias Callback = (repeat each S) -> ()
+typealias Factory = Other<repeat each S>
+```
+
+Unlike structs and classes, but like other type aliases, variadic type aliases can also be nested inside of generic functions.
+
+### Classes
+
+While there are no restrictions on non-final classes adopting type parameter packs, for the time being the proposal restricts such classes from being the superclass of another class.
```
Jumhyn, line 190: Ought it at least be an error at this stage to declare variadic classes as `open`?

proposals/nnnn-variadic-types.md
```diff
@@ -0,0 +1,239 @@
...
+  func demostrate1<each U>(_: repeat ImposeRequirement<each U>)
+  
+  // Infers 'Int: P, V: P, repeat each U: P'
+  func demostrate2<each U, V>(_: ImposeRepeatedRequirement<Int, V, repeat each U>)
+  
+  // Error. Would attempt to infer 'repeat <U' = each U> repeat <V' = each V> U'.A == V' which is not a supported requirement in the language
+  func demonstrate3a<each U, each V>(_: repeat ImposeRepeatedSameType<each U, repeat each V>))
+  
+  // Infers 'Int: P, repeat each V: P'
+  func demostrate3b<each U, each V>(_: repeat (each U, ImposeRequirementExpansion<Int, repeat each V>))
```
Jumhyn, line 167: Is this missing a type definition above for `ImposeRequirementExpansion` or should it reuse `ImposeRepeatedRequirement`?

proposals/nnnn-variadic-types.md
hborla: Yeah, I think we should replace 

> Upcoming Feature Flag: `VariadicGenerics`

with

> Implementation: On `main` and `release/5.9` gated behind the frontend flag `-enable-experimental-feature VariadicGenerics`

Because `VariadicGenerics` is an experimental feature, not an upcoming one.

proposals/nnnn-variadic-types.md
hborla: We should discuss this during the review. Slava and I figured it might still be useful to be able to declare `open` classes with parameter packs and use them, with the subclassing limitation lifted in a later Swift version (I think the work here is all on the code generation side, right @slavapestov?), because the implementation of this proposal has new runtime support and the `open` vs `public` decision is ABI.

jumhyn-duolingo: Wait, I thought `public` to `open` was a permitted change? I agree with the (similar) rationale below that `final` should not be required. Anyway, yeah, this is appropriate for the review thread.

proposals/nnnn-variadic-types.md
hborla: Oops, I renamed `ImposeRequirementExpansion` to `ImposeRepeatedRequirement` and forgot to fix this line. It should be `ImposeRepeatedRequirement`

PR #2016, opened 2023-04-14
Tuple of value pack expansion proposal

sophiapoirier:
pitch: https://forums.swift.org/t/tuple-of-value-pack-expansion/64269

### DIFF
--- a/proposals/nnnn-tuple-of-value-pack-expansion.md
+++ b/proposals/nnnn-tuple-of-value-pack-expansion.md
@@ -1,0 +1,88 @@
+# Tuple of value pack expansion
+
+* Proposal: [SE-NNNN](nnnn-tuple-of-value-pack-expansion.md)
+* Authors: [Sophia Poirier](https://github.com/sophiapoirier), [Holly Borla](https://github.com/hborla)
+* Review Manager: TBD
+* Status: **Awaiting implementation**
+* Upcoming Feature Flag: `VariadicGenerics`
+* Previous Proposal: [SE-0393](0393-parameter-packs.md)
+
+## Introduction
+
+Building upon the **Value and Type Parameter Packs** proposal [SE-0393](https://forums.swift.org/t/se-0393-value-and-type-parameter-packs/63859), this proposal enables referencing a tuple value that contains a value pack inside a pack repetition pattern.
+
+## Motivation
+
+When a tuple value contains a value pack, there is no way to reference those pack elements or pass them as a value pack function argument.
+
+Additionally, type parameter packs are only permitted within a function parameter list, tuple element, or generic argument list. This precludes declaring a type parameter pack as a function return type, type alias, or as a local variable type to permit storage. The available solution to these restrictions is to contain the value pack in a tuple, which makes it important to provide full functional parity between value packs and tuple values containing them. This proposal fills that functionality gap by providing a method to reference individual value pack elements contained within a tuple.
+
+## Proposed solution
+
+This proposal extends the functionality of pack repetition patterns to values of _abstract tuple type_, which enables an implicit conversion of an _abstract tuple value_ to its contained value pack. An _abstract tuple type_ is a tuple that has an unknown length and elements of unknown types. Its elements are that of a single type parameter pack and no additional elements, and no label. In other words, the elements of the type parameter pack are the elements of the tuple. An _abstract tuple value_ is a value of _abstract tuple type_. This proposal provides methods to individually access the dynamic pack elements of an abstract tuple value inside of a repetition pattern.
+
+### Distinction between tuple values and value packs
+
+The following example demonstrates a pack repetition pattern on a value pack and an abstract tuple value separately first, then together but with the repetition pattern operating only on the value pack, and finally with the repetition pattern operating on both the value pack and the tuple value's contained value pack together interleaved.
+
+```swift
+func example<each U>(packElements value: repeat each U, tuple: (repeat each U)) {
+  print((repeat each value))
+  print((repeat each tuple))
+  
+  print((repeat (each value, tuple)))
+  
+  print((repeat (each value, each tuple)))
+}
+
+example(packElements: 1, 2, 3, tuple: (4, 5, 6))
+
+// Prints the following output:
+// (1, 2, 3)
+// (4, 5, 6)
+// ((1, (4, 5, 6)), (2, (4, 5, 6)), (3, (4, 5, 6)))
+// ((1, 4), (2, 5), (3, 6))
+```
+
+## Detailed design
+
+Pack reference expressions inside a repetition pattern can have abstract tuple type. The outer structure of the tuple is removed, leaving the elements of a value pack:
+
+```swift
+func expand<each T>(value: (repeat each T)) -> (repeat (each T)?) {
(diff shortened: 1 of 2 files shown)


xwu:
Hi @sophiapoirier, any update on the state of implementation for this proposal?

proposals/nnnn-tuple-of-value-pack-expansion.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Proposed solution
+
+This proposal extends the functionality of pack repetition patterns to values of _abstract tuple type_, which enables an implicit conversion of an _abstract tuple value_ to its contained value pack. An _abstract tuple type_ is a tuple that has an unknown length and elements of unknown types. Its elements are that of a single type parameter pack and no additional elements, and no label. In other words, the elements of the type parameter pack are the elements of the tuple. An _abstract tuple value_ is a value of _abstract tuple type_. This proposal provides methods to individually access the dynamic pack elements of an abstract tuple value inside of a repetition pattern.
```
xwu, line 22: Can you expand on this slightly--perhaps an example or two to illustrate use of the proposed feature? Ideally, these examples would also illustrate what isn't permitted (in terms of labels, etc.).

I think reviewers are going to be interested to know what this proposal will allow them or us to do. For example, will it make it possible to implement a variadic `ZipSequence` or variadic tuple `==`?

proposals/nnnn-tuple-of-value-pack-expansion.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Distinction between tuple values and value packs
+
+The following example demonstrates a pack repetition pattern on a value pack and an abstract tuple value separately first, then together but with the repetition pattern operating only on the value pack, and finally with the repetition pattern operating on both the value pack and the tuple value's contained value pack together interleaved.
+
+```swift
+func example<each U>(packElements value: repeat each U, tuple: (repeat each U)) {
+  print((repeat each value))
+  print((repeat each tuple))
```
xwu, line 31: Question: how many of these nested parens are mandatory, and which ones are written here only for clarity? It's particularly eye-catching with the line after that ends with `)))`. Is there an addition to the formal grammar of the language that can be specified in this proposal, whether here or in the Detailed Design?

proposals/nnnn-tuple-of-value-pack-expansion.md
```diff
@@ -0,0 +1,88 @@
...
+Applying the pack repetition pattern effectively removes the outer structure of the tuple leaving just the value pack. In the repetition expression, the base tuple is evaluated once before iterating over its elements.
+
+```swift
+repeat each <tuple expression>
+
+// the above is evaluated like this
+let tempTuple = <tuple expression>
+repeat each tempTuple
+```
+TODO: is this example actually informative or illustrative?
```
xwu, line 66: Please address this TODO when able :)
Illustrative and informative examples are, of course, super helpful here.

proposals/nnnn-tuple-of-value-pack-expansion.md
```diff
@@ -0,0 +1,88 @@
...
+
+This proposal does not add or affect ABI as its impact is only on expressions. It does not change external declarations or types. It rests atop the ABI introduced in the **Value and Type Parameter Packs** proposal.
+
+## Implications on adoption
+
+Given that this change rests atop the ABI introduced in the **Value and Type Parameter Packs** proposal, this shares with it the same runtime back-deployment story.
+
+## Alternatives considered
+
+The **Value and Type Parameter Packs** proposal provides no means for declaring local variable value packs or stored property packs. This is mentioned as a potential future direction, but without that, containing the packs within tuples is the way to access such functionality. Therefore an alternative to this proposal would be to instead pursue those future directions instead. However that would only partially address the motivations for this proposal, still leaving a broader pack repetition pattern design needed for abstract tuples.
```
xwu, line 82: This sounds like it's not really an alternative to the proposal but rather a potentially orthogonal future direction?

An alternative here might be just the originally implemented syntax (which some users in the pitch stage said they favored)--that way, you've got an opportunity to explain/defend why your chosen syntax proposed here is better.

proposals/nnnn-tuple-of-value-pack-expansion.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Future directions
+
+* Repetition patterns for concrete tuples
+* Pack repetition patterns for arrays containing a value pack where all pack elements are of the same type
+* Lifting the restriction that a tuple containing a value pack must contain only the single value pack and use no label. This would be required to enable pack repetition patterns for a contained value pack amongst arbitrary other tuple elements, addressable via a label.
```
xwu, line 88: It would be great if there could be a _brief_ explanation as to what these future directions mean; doesn't have to be a novel but a vignette or a description of the features. This would help reviewers better understand not just what they are but also give feedback if they have reason to believe that any aspect of the current proposed design would close off a desired future direction.

sophiapoirier:
> Hi @sophiapoirier, any update on the state of implementation for this proposal?

The current implementation requires adding `.element` to the abstract tuple when you want to expand it, however we decided to not require that (and pitched it as not being required), and the implementation for dropping `.element` is in progress (a bit delayed in favor of WWDC work alas).

proposals/nnnn-tuple-of-value-pack-expansion.md
sophiapoirier: Things like `ZipSequence` and tuple equality are already possible with the main feature proposal, however you become stuck with being able to do anything with a pack's elements after a value pack is forced into a tuple in order to be returned from a function or stored to a variable. So this is less about enabling the implementation of some function, but more what you can do next when you call any function that returns value packs. I will clarify this and add examples (including invalid ones).

proposals/nnnn-tuple-of-value-pack-expansion.md
xwu: Ahhhh...good point about `print`'s current declaration. Would it be helpful, simply for clarity/explanation purposes, if you declared a variadic `func print` with a code comment that this doesn't exist (yet) in the standard library, then illustrate calling that function with values? That way, the use site illustrates this specific proposal's feature without simultaneously trying to working around a library deficiency.

xwu:
Fantastic to hear that the last bit of implementation is in progress; if confident that it's workable from the implementation standpoint but not far along to be merged soon, I suppose we could run the proposal without the final syntax in place, but the ideal scenario would be if that PR could be in the nightlies :)

proposals/nnnn-tuple-of-value-pack-expansion.md
sophiapoirier: I liked this suggestion and I started making this change, and then I wasn't really happy about the fact that the current example in comments shows you the output that you get from the example code, except by making this change, there is no literal output anymore because it uses a library function that does not exist, and I think that imagining what the output would be is less informative. My inclination now is to maybe just add a comment about why the extra parenthesis are there and stick with an example that actually runs with verifiably known output. Thoughts/opinions?

xwu: Good point: whatever you feel is most clear!

PR #2021, opened 2023-04-21
[SE-0396] Correct description for Either type with Codable generics

Jager-yoo:
Hi, I corrected the description for Either with Codable generics.
Please review the changes. Thanks!


allevato:
For what it's worth, the original text is also grammatically correct; and having the condition (the "where" clause) precede the thing that's applied based on that condition ("are `Codable`") discloses the information in clearer order, in my opinion.

Jager-yoo:
I see, makes sense! Thank you for clarifying.

PR #2027, opened 2023-04-26
Fix review manager link in SE-0384

dempseyatgithub:
Ensure review manager name is a markdown link to their GitHub profile URL.

This also ensures correct link behavior for the review manager field on the Swift Evolution dashboard.


xwu:
Thanks!

PR #2028, opened 2023-04-27
Fix review manager link in SE-0389

dempseyatgithub:
Ensure review manager name is a markdown link to their GitHub profile URL.

This also ensures correct link behavior on the
Swift Evolution dashboard.

Sorry I didn't catch this and include it with the previous PR.
I just ran a script over all proposals and this should be the only remaining review manager or author that needs this fix.


rjmccall (approved):

rjmccall:
Thanks!

PR #2069, opened 2023-05-31
Add syntax highlighting.

ZevEisenberg:
I think I got the right blend of Swift and C++, but it's worth double checking. I had to split one example because it had code in both languages.


rjmccall (approved):

rjmccall:
We already merged a PR doing some of this; could you re-prepare the rest of the patch?

ZevEisenberg:
There's not much left after a rebase. Feel free to merge or close.

rjmccall:
Ah, yeah, I guess we don't need this anymore.  Sorry, that was hard to see in the web view, thanks for following up.

PR #2070, opened 2023-05-31
Nested Protocols

karwa:


hborla:
Hey @karwa, I'll be the review manager for this. Will you be available July 24-August 7th to answer review questions?

proposals/xxxx-nested-protocols.md
```diff
@@ -0,0 +1,150 @@
+# Allow Protocols to be Nested in Non-Generic Contexts
+
+* Proposal: [SE-NNNN](NNNN-filename.md)
+* Authors: [Karl Wagner](https://github.com/karwa)
+* Review Manager: TBD
+* Status: **Awaiting review**
+* Implementation: [apple/swift#66247](https://github.com/apple/swift/pull/66247)
+* Upcoming Feature Flag: `NestedProtocols`
+* Review: ([pitch](https://forums.swift.org/...))
```
hborla, line 9:
```suggestion
* Review: ([pitch](https://forums.swift.org/t/pitch-allow-protocols-to-be-nested-in-non-generic-contexts/65285))
```

proposals/xxxx-nested-protocols.md
```diff
@@ -0,0 +1,150 @@
+# Allow Protocols to be Nested in Non-Generic Contexts
+
+* Proposal: [SE-NNNN](NNNN-filename.md)
+* Authors: [Karl Wagner](https://github.com/karwa)
+* Review Manager: TBD
+* Status: **Awaiting review**
+* Implementation: [apple/swift#66247](https://github.com/apple/swift/pull/66247)
+* Upcoming Feature Flag: `NestedProtocols`
```
hborla, line 8: An "upcoming feature flag" is specifically for source breaking changes; I think you should mention the feature is gated behind `-enable-experimental-feature NestedProtocols` in the **Implementation** field above.

proposals/xxxx-nested-protocols.md
```diff
@@ -0,0 +1,150 @@
...
+
+This proposal is purely an extension of the language's ABI and does not change any existing features.
+
+As with other nested types, the parent context forms part of the mangled name of a nested protocol. Therefore, moving the protocol around is an ABI-incompatible change.
+
+## Implications on adoption
+
+This feature can be freely adopted and un-adopted in source
+code with no deployment constraints and without affecting source or ABI
+compatibility.
```
hborla, line 132: You speak to adoption constraints in the above section:

>As with other nested types, the parent context forms part of the mangled name of a nested protocol. Therefore, moving the protocol around is an ABI-incompatible change.

I think this sentence belongs in **Implications on adoption**. You might also consider mentioning that moving an existing protocol inside a type is, of course, a source-breaking change, but for non-resilient libraries, you can make it a source-compatible change using a type-alias to the newly-nested protocol.

proposals/xxxx-nested-protocols.md
```diff
@@ -0,0 +1,150 @@
...
+   func impl<T: Abstraction>(_ input: T) -> T.ResultType {
+     // ...
+   }
+   
+   let _: Int = impl(SomeConformance())
+   let _: String = impl(AnotherConformance())
+}
+```
+
+## Detailed design
```
hborla, line 78: A question came up in the pitch thread about whether or not a nested protocol satisfies an associated type requirement (as an existential type) of the same name in the same way that nested types or typealiases do: https://forums.swift.org/t/pitch-allow-protocols-to-be-nested-in-non-generic-contexts/65285/22. It's worth specifying the witness matching behavior in the detailed design section here.

hborla (commented):
The proposal looks good overall; I left a few minor comments inline.

karwa:
@hborla I've made the adjustments you suggested, and I'm available to answer any questions over July 24-August 7th.

There are still a couple of bits I need to improve with the implementation (we didn't used to model nested protocols in the AST, so now that we do, the requirement machine can get tripped up by some invalid structures). IMO that doesn't need to hold up the review process, but it's up to you. I messaged Slava about it and I have a good idea of what to do; it's not major and hopefully I'll have it this week sometime.

hborla:
@karwa thank you! I'll plan to kick off the review tomorrow.

I took a look at your implementation PR and I think it's in good enough shape for evolution review. It's okay if it takes a bit longer to address PR feedback (unless there are major concerns about feasibility, which isn't the case here). My only ask is that you rebase the current implementation so I can kick off a toolchain build. If you don't have time to do that before the review, I can do that later today.

karwa:
@hborla Thanks. It's a bit late here in Europe so I may not have time to rebase it tonight. If it's not trivial I can take a look tomorrow.

hborla:
@karwa no worries, I'll take care of it 👍

PR #2075, opened 2023-06-07
Fix outdated link in 0395-observability.md

Ryu0118:
The link about attached macros was outdated and has been corrected.


Jumhyn:
Thanks!

PR #2088, opened 2023-06-21
[Proposal] - Add `with(_:transform:)` to the standard library

aggie33:
The pitch thread is [here](https://forums.swift.org/t/pitch-with-functions-in-the-standard-library/65716)


aggie33:
Making a new one now

PR #2099, opened 2023-07-06
Apply `@storageRestrictions` only to init accessors

natecook1000:
Updates the proposal to apply the new attribute only to `init` accessors, rather than to computed property declarations.


xedin (approved):

DougGregor (approved):
Whoops, thank you !

PR #2100, opened 2023-07-07
[SE-0346] Fix error in code snippet in Alternatives considered

ole:
The return type of a convert-to-`Double` function should be `Double`, not `Int`. This looks like a copy-paste oversight.

This is a (trivial and immaterial, I hope) fix for an already accepted proposal.


rjmccall:
Thanks!

PR #2102, opened 2023-07-11
Update Observation proposal with new revisions

natecook1000:
This revision converts `Observable` to a regular (non-marker) protocol, limits the `@Observable` macro to classes, and includes other clarifications to the proposal.


DmT021:
@natecook1000 Did you see my question [here](https://github.com/apple/swift/pull/66993#issuecomment-1630535410)? Could you reply please?

PR #2123, opened 2023-08-07
[Proposal] - Add with methods to Swift

aggie33:
This proposal has an implementation by @calda and uses magic `with` methods instead of the free with function.


calda:
Link to the implementation: https://github.com/apple/swift/pull/67768

proposals/NNNN-add-with-methods-to-swift.md
```diff
@@ -0,0 +1,437 @@
...
+
+// Today this prints "DYNAMIC MEMBER WITH"
+//
+// With the new `with` method, by default this would produce an error: 
+// ambiguous use of 'with'
+let string = Foo().with
+print(string.uppercased())
+```
+
+To prevent this source break, we can make it so that the new `with` overload doesn't suppress the dynamic member subscrpt overload when accessing a dyanmic member named "with". Combining this with `@_disfavoredOverload` allows the above example to continue compiling / functioning exactly as it did before.
```
Sherlouk, line 262:
```suggestion
To prevent this source break, we can make it so that the new `with` overload doesn't suppress the dynamic member subscrpt overload when accessing a dynamic member named "with". Combining this with `@_disfavoredOverload` allows the above example to continue compiling / functioning exactly as it did before.
```

PR #2131, opened 2023-08-19
Member macro conformances

DougGregor:


proposals/nnnn-member-macro-conformances.md
```diff
@@ -0,0 +1,116 @@
...
+- An overridable member of a non-final class.
+- A stored property or case can only be in the primary type definition.
+
+For these cases, a member macro can produce the declarations. However, member macros aren't provided with any information about which protocol conformances they should provide members for, so a macro might erroneously try to add conforming members to a type that already conforms to the protocol (e.g., through a superclass). This can make certain macros---such as macros that implement the `Encodable` or `Decodable` protocols---unimplemented.
+
+## Proposed solution
+
+To make it possible for a member macro to provide the right members for a conformance, we propose to extend member macros with the same ability that extension macros have to reason about conformances. Specifically:
+
+* The `attached` attribute specifying a `member` role gains the ability to specify the set of protocol conformances it is interest in, the same way an `extension` macro specifies the conformances it can provide.
```
rjmccall, line 27:
```suggestion
* The `attached` attribute specifying a `member` role gains the ability to specify the set of protocol conformances it is interested in, the same way an `extension` macro specifies the conformances it can provide.
```

proposals/nnnn-member-macro-conformances.md
```diff
@@ -0,0 +1,116 @@
...
+
+```swift
+@attached(member, conformances: Decodable, Encodable, names: named(init(from:), encode(to:)))
+@attached(extension, conformances: Decodable, Encodable, names: named(init(from:), encode(to:)))
+macro Codable() = #externalMacro(module: "MyMacros", type: "CodableMacro")
+```
+
+This macro has several important decisions to make about where and how to generate `init(from:)` and `encode(to:)`:
+
+* For a struct, enum, actor, or final class, `init(from:)` and `encode(to:)`should be emitted into an extension (via the member role) along with the conformance. This is both good style and, for structs, ensures that the initializer doesn't inhibit the memberwise initializer.
```
rjmccall, line 42:
```suggestion
* For a struct, enum, actor, or final class, `init(from:)` and `encode(to:)` should be emitted into an extension (via the member role) along with the conformance. This is both good style and, for structs, ensures that the initializer doesn't inhibit the memberwise initializer.
```

proposals/nnnn-member-macro-conformances.md
```diff
@@ -0,0 +1,116 @@
...
+* For a non-final class, `init(from:)` and `encode(to:)`should be emitted into the main class definition (via the member role) so that can be overridden by subclasses.
```
rjmccall, line 43:
```suggestion
* For a non-final class, `init(from:)` and `encode(to:)` should be emitted into the main class definition (via the member role) so that they can be overridden by subclasses.
```

proposals/nnnn-member-macro-conformances.md
```diff
@@ -0,0 +1,116 @@
...
...primary drawback to this notion of implementation extensions is that it will no longer be possible to look at the primary definition of a type to find it's full "shape": it's stored properties, designated/required initializers, overridable methods, and so on.  Instead, that information can be scattered amongst the original type definition and any implementation extensions, requiring one to stitch together a view of the whole type. This would have a particularly negative effect on macros that want to reason about the shape of the type, because macros only see a single entity (such as a class or struct definition), and not across extensions to that entity. The `Codable` macro discussed in this proposal, for example, would not be able to encode or decode stored properties that are written in an implementation extension, and the [`Observable`](https://github.com/apple/swift-evolution/blob/main/proposals/0395-observability.md) macro would silently fail to observe any properties written in an implementation extension. By trying to use implementation extensions to address the shortcoming of macros described in this proposal, we would end up creating a larger problem for those same macros.
```
rjmccall, line 116:
```suggestion
The primary drawback to this notion of implementation extensions is that it would no longer be possible to look at the primary definition of a type to find its full "shape": its stored properties, designated/required initializers, overridable methods, and so on.  Instead, that information could be scattered amongst the original type definition and any implementation extensions, requiring readers to stitch together a view of the whole type. This would have a particularly negative effect on macros that want to reason about the shape of the type, because macros only see a single entity (such as a class or struct definition) and not extensions to that entity. The `Codable` macro discussed in this proposal, for example, would not be able to encode or decode stored properties that are written in an implementation extension, and the [`Observable`](https://github.com/apple/swift-evolution/blob/main/proposals/0395-observability.md) macro would silently fail to observe any properties written in an implementation extension. By trying to use implementation extensions to address the shortcoming of macros described in this proposal, we would end up creating a larger problem for those same macros.
```

rjmccall (commented):
Just a few editorial fixes.  It looks like you've summarized the pitch conversation well.

PR #2160, opened 2023-09-12
Update 0397-freestanding-declaration-macros.md

natecook1000:
The required method for `DeclarationMacro` conformance isn't actually `async`.


rjmccall:
@natecook1000 I asked @DougGregor about this, and he thinks we should probably change all of the macro proposals at once if we're going to fix this.

PR #2165, opened 2023-09-17
[Proposal] Access Level must follow as Guiding Principle.md

shlim0:


rjmccall:
You don't need an evolution proposal to change the documentation.  You'd need an evolution proposal to change the actual language.

PR #2169, opened 2023-09-19
Noncopyable generics

kavon:


proposals/NNNN-noncopyable-generics.md
```diff
@@ -0,0 +1,975 @@
...
+```
+
+Once `Elm` does not have a `Copyable` constraint, a noncopyable type like
+`FileDescriptor` can be substituted in `List`, in addition to copyable ones.
+
+#### Automatic synthesis of conditional `Copyable` conformance
+A common reason for having a type parameter is that values of that generic type
+will be stored somewhere within the parameterized struct or enum type. Because
+of `Copyable`'s containment rule, that implies the parameterized type itself
+is commonly noncopyable if any of its type parameters is noncopyable. This
```
airspeedswift, line 235: maybe even "usually" rather than "commonly"?

proposals/NNNN-noncopyable-generics.md
```diff
@@ -0,0 +1,975 @@
...
+struct Factory<Product: ~Copyable, Token: ~Copyable, Policy> {
+  var seed = 0
+
+  // signature <Product, Token, Policy where Policy : Copyable>
+  mutating func produce(_ payment: consuming Token) -> Product { /* ... */ }
+}
+/* extension Factory: Copyable where Product: Copyable, Token: Copyable {} */
+```
+
+Notice that despite this `Factory` not storing any noncopyable data, it does not
```
airspeedswift, line 263: This might need justification, about how you wouldn't want the synthesizing behavior to have to be determined by the implementation but by the signature.

proposals/NNNN-noncopyable-generics.md
```diff
@@ -0,0 +1,975 @@
...
+parameters has removed its default `Copyable` constraint. Instead, a
+conditional conformance was synthesized to make `Factory` copyable once those
+type parameters are `Copyable`.
+
+The synthesis procedure generates the `where` clause of the conditional
+`Copyable` conformance by enumerating _all_ of the type parameters and requiring
+each one to be `Copyable`. It does not matter whether the parameter is used in
+the storage of the `Factory` struct.
+
+One case where this synthesis procedure produces an invalid conformance is
```
airspeedswift, line 274: this feels like it is just one more case where you need control over the synthesis, like the subsequent paragraphs, but is instead special casted just because it's not _optional_ to use this control, it's mandatory.

proposals/NNNN-noncopyable-generics.md
```diff
@@ -0,0 +1,975 @@
...
+`~Copyable` on the type parameter:
+
+```swift
+class FileHandle<File: ~Copyable> /* : Copyable */ {
+  var file: File
+  // ...
+}
+```
+
+Classes can contain noncopyable storage without itself becoming noncopyable,
```
airspeedswift, line 343:
```suggestion
Classes can contain noncopyable storage without themselves becoming noncopyable,
```

proposals/NNNN-noncopyable-generics.md
```diff
@@ -0,0 +1,975 @@
...
+struct DataSet /* : Copyable */ {
+  var samples: [Double]
+}
+```
+
+As with any other generic parameter, there is a `Copyable` constraint on the
+`Elm` in this `List` too:
+
+```swift
+enum List<Elm> /* : Copyable where Elm: Copyable */ {
```
airspeedswift, line 215: Is there a reason to use `Elm` everywhere instead of `Element`? Do you want to avoid confusion with `Collection.Element`?

airspeedswift: (a reason other than "it's shorter :)

proposals/NNNN-noncopyable-generics.md
```diff
@@ -0,0 +1,975 @@
...
+}
+```
+
+Of course, when the conformer _does_ happen to be copyable, additional 
+functionality can be made available to it using a conditional extension:
+
+```swift
+extension EventLog where Self: Copyable {
+  func duplicate() -> Self { 
+    return copy self // OK
```
airspeedswift, line 432:
```suggestion
   copy self // OK
```

proposals/NNNN-noncopyable-generics.md
```diff
@@ -0,0 +1,975 @@
...
+```swift
+protocol Pizza: ~Copyable {
+  associatedtype Topping: ~Copyable
+  func peelOneTopping() -> Topping
+}
+
+let t: any Token = ... // signature <Self where Self: Token>
+let _: any ~Copyable = t.peelOneTopping() // signature <Self>
+```
+For associated types within a protocol erased to an existential preserve
```
airspeedswift, line 452: this sentence is ungrammatical, is the fix to remove "For" at the start?

proposals/NNNN-noncopyable-generics.md
```diff
@@ -0,0 +1,975 @@
...
+conformances to a layout protocol like `Copyable`. Thus, when calling
```
airspeedswift, line 453:
```suggestion
conformances to a layout protocol like `Copyable`. So when calling
```

airspeedswift (changes requested):
In general, I think this needs some more discussion of why the behaviors  chosen here are the right ones - that it's for more than just for source compatibility, but for progressive discover where e.g. it encourages a world where types are created so that someone who is extending them does not need to know it can handle non-copyable types until they need to.

proposals/NNNN-noncopyable-generics.md
kavon: No good reason

proposals/NNNN-noncopyable-generics.md
kavon: One issue with the justification is that Sendable does it differently; it won't synthesize the Sendable implicitly if some storage is not Sendable. Will need to dig into the details and see if we can get them to work the same.

proposals/NNNN-noncopyable-generics.md
```diff
@@ -0,0 +1,713 @@
...
+In particular, the tightly defined lifecycle allows noncopyable structs
+to carry deinitializers:
+```swift
+struct HeapBuffer: ~Copyable {
+  init() { ... allocate working storage on heap ... }
+  deinit() { ... release storage ... }
+}
+```
+
+But SE-0390 also made a number of concessions to simplify the initial implementation,
```
hborla, line 83: I would re-frame this in terms of making incremental progress on supporting non-copyable types in the language. SE-0390 is a useful step on its own, and the generics interactions are complicated with a lot of nuanced design decisions (as demonstrated in this proposal!) so it was best to subset it out into its own proposal to carefully consider the implications of those interactions in the generics system.

proposals/NNNN-noncopyable-generics.md
```diff
@@ -0,0 +1,713 @@
...
+
+#### Protocol Extensions
+
+An extension of a protocol is implicitly constrained to `Self: Copyable` unless you explicitly state otherwise.
+```swift
+protocol EventLog: ~Copyable {
+  ...
+}
+
+extension EventLog /* where Self: Copyable */ {
```
hborla, line 312: Is `Copyable` also the default in protocol extensions for associated types that are `~Copyable`? Can you please specify the rule and add an example?

proposals/NNNN-noncopyable-generics.md
```diff
@@ -0,0 +1,713 @@
...
+to conditionally declare copyability.
+
+Since `Copyable` is an additive capability, this must be done by declaring
+the base type to not require copyability and then conditionally extending it:
+```swift
+struct MaybeCopyable<T: ~Copyable>: ~Copyable {
+  var t: T
+}
+
+extension MaybeCopyable /* : Copyable where T: Copyable */ { }
```
hborla, line 388: I was very confused by this code example and originally thought it implied that adding an empty `extension` on a type that is `~Copyable` with a `~Copyable` type parameter implied a conditional conformance like the commented where clause is suggesting. I'm not sure what the correct interpretation of this example is. Can you please clarify what this means?

proposals/NNNN-noncopyable-generics.md
```diff
@@ -0,0 +1,713 @@
...
+```
+
+Conditional conformances to `Copyable` can only depend on
+generic parameters being `Copyable`.
```
hborla, line 392: Can you please add justification for this design decision? Is this an implementation difficulty that could be lifted in the future?

proposals/NNNN-noncopyable-generics.md
```diff
@@ -0,0 +1,713 @@
...
+extension Foo: Copyable where T: Copyable { } // OK
+extension Foo: Copyable where T: P { } // 🛑
+extension Foo: Copyable where T == Int {} // 🛑
+extension Foo: Copyable where T: Sendable {} // 🛑
+extension Foo: Copyable where T.A: Copyable {} // 🛑
+```
+
+### The top type
+
+The type `Any` is no longer the "top" type in the language, which is the type that is a supertype of all types.
```
hborla, line 409: I'm not sure what the purpose of this section is. Can you remove it and instead show some of the valid and invalid subtype conversions for existentials in the following section? I think including this in the Existentials section makes more sense, because right now you're talking about existentials suppressing Copyable before introducing that as a concept.

proposals/NNNN-noncopyable-generics.md
```diff
@@ -0,0 +1,713 @@
...
+**Source**: Adding `~Copyable` to a generic parameter on a function
+is generally not source-breaking for existing clients that provide
+implicitly-`Copyable` types for such parameters.
+
+Adding `~Copyable` to an existing type or protocol (generic or concrete) is typically
+source-breaking, as existing code may rely on the ability to copy values of this type.
+
+Adding `~Copyable` to an associated type is also generally source breaking.
+
+## Alternatives Considered
```
hborla, line 514: I know there's been a lot of careful consideration around the defaults for protocols and their associated types. If there were any alternative directions that were considered in that area, e.g. not inferring protocols to refine `Copyable` by default, please specify them here and why those directions were not chosen.

proposals/NNNN-noncopyable-generics.md
kavon: yeah this was a typo i think

kavon:
this PR was subsumed by  https://github.com/apple/swift-evolution/pull/2356

PR #2177, opened 2023-10-01
Update url to swift package registry spec

Palleas:
The registry specification file has moved from `https://github.com/apple/swift-package-manager/blob/main/Documentation/Registry.md` to `https://github.com/apple/swift-package-manager/blob/main/Documentation/PackageRegistry/Registry.md`


yim-lee (approved):
Thanks @Palleas

PR #2178, opened 2023-10-02
Remove mention of obsolete feature flags

slavapestov:
It's caused confusion several times already in the past and I don't think there's any historical value in preserving the original flag names here.


rjmccall:
Maybe we should just remove the `Implementation` line from released proposal documents if people are finding it confusing?  I'll talk it over with the LSG.

rjmccall:
Okay, LSG has decided that we should just remove Implementation lines from the header after the proposal status moves to `Implemented (Swift X.Y)`.  Could you make that change?

slavapestov:
@rjmccall Sounds good, I've made the change.

PR #2183, opened 2023-10-06
Add a proposal for isolated default value expressions.

hborla:
Pitched at https://forums.swift.org/t/pitch-isolated-default-value-expressions/67714

### DIFF
--- a/proposals/NNNN-isolated-default-values.md
+++ b/proposals/NNNN-isolated-default-values.md
@@ -245,30 +245,40 @@ Note that this rule is not specific to default values, but it's necessary to spe
 
 For structs, default initializer expressions for stored properties are used as default argument values to the compiler-generated memberwise initializer. For structs and classes that have a compiler-generated no-parameter initializer, the default initializer expressions are also used in the syntheszied `init()` body.
 
-If any of the type's stored properties are actor isolated, then the compiler-synthesized initializer(s) must also be actor isolated. For example:
+If any of the type's stored properties with non-`Sendable` type are actor isolated, or if any of the isolated default initializer expressions are actor isolated, then the compiler-synthesized initializer(s) must also be actor isolated. For example:
 
 ```swift
+class NonSendable {}
+
 @MainActor struct MyModel {
   // @MainActor inferred from annotation on enclosing struct
-  var value = requiresMainActor()
+  var value = .init()
 
   /* compiler-synthesized memberwise init is @MainActor
   @MainActor
-  init(value: Int = requiresMainActor()) {
+  init(value: NonSendable = .init()) {
     self.value = value
   }
   */
 }
 ```
 
-If none of the type's stored properties are actor isolated, then the compiler-synthesized initializer is `nonisolated`. For example:
+If none of the type's stored properties are non-`Sendable` and actor isolated, and none of the default initializer expressions require actor isolation, then the compiler-synthesized initializer is `nonisolated`. For example:
 
 ```swift
 @MainActor struct MyView {
-  // no stored properties
+  // @MainActor inferred from annotation on enclosing struct
+  var value: Int = 0
 
-  /* compiler-synthesized init is 'nonisolated'
-  nonisolated init() {}
+  /* compiler-synthesized 'init's are 'nonisolated'
+
+  nonisolated init() {
+    self.value = 0
+  }
+
+  nonisolated init(value: Int = 0) {
+    self.value = value
+  }
   */
 
   // @MainActor inferred from the annotation on the enclosing struct
@@ -276,7 +286,7 @@ If none of the type's stored properties are actor isolated, then the compiler-sy
 }
 ```
 
-These rules ensure that the default value expressions in compiler-synthesized initializers are always valid. A default value expression can only be isolated if the stored property is isolated. If the stored property is isolated, then the compiler-synthesized initializer must also share the same isolation.
+These rules ensure that the default value expressions in compiler-synthesized initializers are always valid. If a default value expression requires actor isolation, then the enclosing initializer always shares the same actor isolation. It is an error for two different default values to require different actor isolation, because it's not possible to ever use those default values. Initializing an instance of a type using two different initial value expressions with different actor isolation must be done in an `async` initializer, with suspension points explicitly marked with `await`.
 
 ## Source compatibility
 
--- a/proposals/NNNN-isolated-default-values.md
+++ b/proposals/NNNN-isolated-default-values.md
@@ -8,7 +8,7 @@
 * Implementation: [apple/swift#68794](https://github.com/apple/swift/pull/68794)
 * Upcoming Feature Flag: `-enable-upcoming-feature IsolatedDefaultValues`
 * Previous Proposal: *if applicable* [SE-XXXX](XXXX-filename.md)
-* Review: ([pitch](https://forums.swift.org/...))
+* Review: ([pitch](https://forums.swift.org/t/pitch-isolated-default-value-expressions/67714))
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


proposals/NNNN-isolated-default-values.md
```diff
@@ -0,0 +1,173 @@
...
+
+@MainActor func mainActorCaller() {
+  useDefault() // okay
+}
+
+func nonisolatedCaller() async {
+  useDefault() // error
+
+  useDefault(value: await requiresMainActor()) // okay
+}
```
Jumhyn, line 118:
```suggestion
func nonisolatedCaller() async {
  useDefault() // error

  useDefault(value: await requiresMainActor()) // okay
}
```
We'd probably expect the first line here to be written as:
```
await useDefault()
```
since we're moving to the main actor, no? And this second line would still need to `await` the `useDefault` call:
```
await useDefault(value: requiresMainActor())
```
We'd still be able to apply the rule that `await` can swallow multiple suspension points inside the expression, right?

hborla: Ah yes, you're right for both cases, thank you!

proposals/NNNN-isolated-default-values.md
```diff
@@ -0,0 +1,173 @@
...
+@MainActor func useDefaultClosure(
+  closure: () -> Void = {
+    requiresMainActor()
+  }
+) {}
+```
+
+The above `useDefaultClosure` function has a default argument value that is a closure literal. The closure body calls a `@MainActor`-isolated function synchronously, therefore the closure itself must be `@MainActor` isolated.
+
+Note that the inferred actor isolation of a closure literal can never be an actor instance based on the following two properties:
```
DougGregor, line 94: I tripped over this sentence a bit, because I interpreted it as "you can't have a closure literal with actor instance isolation in a default argument". You can, if it's written as something like `{ (isolated a: MyActor) in ... }`. What this means to say is that the inference algorithm will never infer such a thing.

hborla: Oops, yeah the "inferred" in this sentence is load bearing and easy to miss. I'll make this point more clear.

proposals/NNNN-isolated-default-values.md
```diff
@@ -0,0 +1,173 @@
...
+
+```swift
+@MainActor func requiresMainActor() -> Int { ... }
+@AnotherActor func requiresAnotherActor() -> Int { ... }
+
+class C {
+  @MainActor var x1: Int = requiresMainActor()
+  @AnotherActor var x2: Int = requiresAnotherActor()
+
+  nonisolated init() {} // error
```
DougGregor, line 133: Please explain why the error is here: neither `x1` nor `x2` are initialized because they have different isolations from the initializer.

proposals/NNNN-isolated-default-values.md
```diff
@@ -0,0 +1,173 @@
...
+    self.x2 = x2
+  }
+}
+```
+
+In the above example, the no-parameter `nonisolated init()` is invalid, because it does not initialize `self.x1` and `self.x2`. Because the default initializer expressions require different actor isolation, those values are not used in the `nonisolated` initializer. The other two initializers are valid.
+
+### Default value isolation in memberwise initializers
+
+For structs, default initializer expressions for stored properties are used as default argument values to the compiler-generated memberwise initializer. In this case, the default argument value shares the same required isolation as the default initializer expression. Because the default values are always evaluated in the caller's context, all the memberwise initializer does is initialize each field, which can always be performed in a `nonisolated` context. In the interest of only applying global actor isolation when it's necessary for the code to run on the global actor, this proposal also changes the isolation of memberwise initializers to be `nonisolated`.
```
DougGregor, line 151: This would benefit from an example that shows what the synthesized memberwise initializer looks like.

proposals/NNNN-isolated-default-values.md
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  @MainActor init(x2: Int) { // okay
+    // 'self.x1' gets assigned to the default value 'requiresMainActor()'
+    self.x2 = x2
+  }
+}
+```
+
+In the above example, the no-parameter `nonisolated init()` is invalid, because it does not initialize `self.x1` and `self.x2`. Because the default initializer expressions require different actor isolation, those values are not used in the `nonisolated` initializer. The other two initializers are valid.
```
DougGregor, line 147: Most folks don't put global actor isolation directly on stored properties, so I think it might be hard for them to extrapolate from what's here to how it affects them. This might benefit from an example with a `@MainActor class` and some stored properties that have default values.

proposals/NNNN-isolated-default-values.md
DougGregor: Alos: classes have a no-argument initializer when their superclass does and all of their stored properties have default values. I suppose we need to apply the same rules there?

hborla: My thinking is the no-argument initializer would still get the actor isolation of the class, because it's so common for default value expressions to have the same isolation. I suppose we could make the no-argument initializer `nonisolated` if all default value expressions are `nonisolated`. Either way, I should specify the behavior of the no-argument initializer in the proposal so it can be discussed on the forums.

proposals/NNNN-isolated-default-values.md
```diff
@@ -0,0 +1,173 @@
...
+@MainActor func useFromMainActor() {
+  f()
+}
+```
+
+## Proposed solution
+
+I propose allowing default value expressions to require the caller to meet an isolation requirement in order to use the default value. The isolation requirement is inferred from the default value expression. If the caller does not meet the isolation requirement, then a value must be written explicitly for the argument or the stored property. This rule makes the stored property example above invalid at the point of the `nonisolated` initializer, because the isolation requirement of the default values for the stored properties is not satisfied. This rule also makes the default argument example above valid, because the `@MainActor` isolation requirement for the default argument of `f` is satisfied by the caller.
+
+## Detailed design
```
DougGregor, line 51: Given the discussion on the pitch, it's probably worth adding a discussion of the order of evaluation for call arguments, because it's strongly influenced the thinking of this proposal.

proposals/NNNN-isolated-default-values.md
hborla: I explain the error in the prose just below the code; is that sufficient?

>In the above example, the no-parameter `nonisolated init()` is invalid, because it does not initialize `self.x1` and `self.x2`. Because the default initializer expressions require different actor isolation, those values are not used in the `nonisolated` initializer.

DougGregor: ... sure, that's sufficient. I suppose it's my personal style to prefer that each "error" in a code sample says why

proposals/NNNN-isolated-default-values.md
hborla: I've expanded the section on compiler-synthesized initializers and included a few examples. The new rules are a little nuanced though, so please let me know if this needs more clarification.

proposals/NNNN-isolated-default-values.md
hborla: I've added a section on argument evaluation.

proposals/NNNN-isolated-default-values.md
```diff
@@ -0,0 +1,318 @@
...
+For structs, default initializer expressions for stored properties are used as default argument values to the compiler-generated memberwise initializer. For structs and classes that have a compiler-generated no-parameter initializer, the default initializer expressions are also used in the syntheszied `init()` body.
+
+If any of the type's stored properties with non-`Sendable` type are actor isolated, or if any of the isolated default initializer expressions are actor isolated, then the compiler-synthesized initializer(s) must also be actor isolated. For example:
+
+```swift
+class NonSendable {}
+
+@MainActor struct MyModel {
+  // @MainActor inferred from annotation on enclosing struct
+  var value = .init()
```
DougGregor, line 255:
```suggestion
  var value: NonSendable = .init()
```

DougGregor (approved):
Looks great!

PR #2186, opened 2023-10-12
Use a star for bullet points in the metadata list

amartini51:
The existing markup is valid, but this change makes these proposals match everything else.


rjmccall (approved):
Sure.

PR #2190, opened 2023-10-20
Process: The Language Workgroup is now the Language Steering Group

AnthonyLatsis:
* Update the group name, generalizing to evolution workgroups where necessary for coherence with the "Community structure" section.
* Introduce the Language Steering Group as the group responsible for maintaining the process in the intro.


DougGregor:
Looks good to me, thank you!

PR #2191, opened 2023-10-21
Update parameter pack syntax in proposal to match implemented syntax

twof:
The written syntax for zero type parameter packs doesn't compile in Swift 5.9. A space needs to be inserted between the angle braces.

Bug report: https://github.com/apple/swift/issues/68290
Forum post: https://forums.swift.org/t/whats-the-syntax-to-create-a-parameter-packed-generic-type-with-zero-type-parameters/67978


rjmccall:
Thanks!

PR #2192, opened 2023-10-24
Fix invalid syntax in variadic-types proposal

MahdiBM:
The proposal contains invalid syntax that doesn't compile on 5.9 RELEASE. This PR fixes that.


DougGregor (approved):

DougGregor:
Thank you!

PR #2204, opened 2023-11-03
[SE-0408] Remove the experimental feature flag

simanerush:
Since the proposal was accepted, and I removed the flag from the implementation, I think I can also remove its mention from the document.


DougGregor:
When the implementation is complete in `main` and enabled by default, we should remove the "Implementation" field entirely from the proposal. But this is a fine partial step

PR #2205, opened 2023-11-07
Normalize Upcoming Feature Flag for SE-0411

dempseyatgithub:
The Upcoming Feature Flag header field typically only includes the feature flag itself, not the compiler option.

Including the compiler option makes it more difficult for automated tools to parse for the Swift Evolution Dashboard.

This PR removes the compiler option to normalize the header field.


DougGregor:
Thank you!

PR #2209, opened 2023-11-09
Nonfrozen enums

karwa:


rjmccall:
The LSG recently talked about this, and we think there isn't an immediate path for acting on this; we need a more comprehensive vision about how we're going to handle this kind of source-compatible library evolution in the language. We can resurrect ideas like this when that's in place.

PR #2213, opened 2023-11-15
Fix JSON comments highlighting in SE-0391

MaxDesiatov:
JSON comments are only supported in JSON 5. We should make code blocks using comments with `json5` explicitly to fix syntax highlighting for those.

Before:

Screenshot 2023-11-08 at 17 47 01

After:

Screenshot 2023-11-08 at 17 47 22


yim-lee (approved):
Thanks @MaxDesiatov!

PR #2219, opened 2023-11-27
Function body macros

DougGregor:


proposals/nnnn-function-body-macros.md
```diff
@@ -0,0 +1,283 @@
...
+
+On the other hand, having a third kind of macro role for function body macros adds yet more language complexity, and introducing this role in lieu of allowing function body macros to replace an existing function body might be overfitting to today's use cases.
+
+### Type-checking bodies as they were written
+
+As noded previously, not checking the body of functions that was written by the user and then replaced by a `body` macro has some down sides. For one, it allows some abuse, where code that wouldn't make sense in Swift is permitted to be written by the user and then significantly altered by the `body` macro. Moreover, wherever the macro is performing some modification that makes ill-formed code into well-formed code (even by something as simple as introducing a `span` variable like `@Tracing` does), tools that cannot reason about the macro expansion might be less useful: code completion won't know to provide `span` as a possible completion, nor will it know what type `span` would have. Therefore, the experience of writing code that makes use of `body` macros could be significantly worse than that for normal Swift code.
```
ktoso, line 270: minor typo fix for consistency: 

```suggestion
As noded previously, not checking the body of functions that was written by the user and then replaced by a `body` macro has some down sides. For one, it allows some abuse, where code that wouldn't make sense in Swift is permitted to be written by the user and then significantly altered by the `body` macro. Moreover, wherever the macro is performing some modification that makes ill-formed code into well-formed code (even by something as simple as introducing a `span` variable like `@Traced` does), tools that cannot reason about the macro expansion might be less useful: code completion won't know to provide `span` as a possible completion, nor will it know what type `span` would have. Therefore, the experience of writing code that makes use of `body` macros could be significantly worse than that for normal Swift code.
```

proposals/nnnn-function-body-macros.md
```diff
@@ -0,0 +1,338 @@
...
+  /// expansion will be inserted at the beginning of the function body.
+  static func expansion(
+    of node: AttributeSyntax,
+    providingPreambleFor declaration: some DeclSyntaxProtocol & WithOptionalCodeBlockSyntax,
+    in context: some MacroExpansionContext
+  ) throws -> [CodeBlockItemSyntax]
+}
+```
+
+Preamble macros don't provide a complete function body. Rather, they provide code items that will be introduced at the beginning of the existing function body. Preamble macros are useful for injecting code without changing anything else about the function body, for example to introduce logging or simple tracing facilities. The `defer` statement can be used to trigger code that will run once the function returns.
```
allevato, line 162: There was an interesting question raised [here](https://forums.swift.org/t/function-body-macros/66471/34) which I don't think has been addressed. If a function body pre-expansion is a single-expression implicit return, how do function body macros interact with it?

```swift
@SomeMacro
func f(a: Int, b: Int) {
  a + b
}
```

* If this is a **preamble** macro, the macro doesn't have an opportunity to update the `a + b` expression. Is there a special carve-out for this to still work as an implicit `return` or would that now fail?
* If this is a **body** macro, is it the responsibility of that macro to transform the `expr` `CodeBlockItem` into an `stmt` item with the equivalent `ReturnStmt`?

allevato (commented):
I'll be review managing this proposal. Looks great! Just a small number of possible clarifications after reading this along with revisiting the pitch thread.

proposals/nnnn-function-body-macros.md
```diff
@@ -0,0 +1,338 @@
...
+
+```swift
+{
+  // code items produced by preamble macro A
+  // code items produced by preamble macro B
+  // code items produced by function body macro C
+}
+```
+
+Preamble macros may introduce new local declarations, which will be visible to later preamble macros (e.g., names introduced by `@A` will be visible in `@B` and `@C` as well) and the remainder of the function body. As with all macros that introduce new names into existing code, those names must be documented with a [`names` clause](https://github.com/apple/swift-evolution/blob/main/proposals/0389-attached-macros.md#specifying-newly-introduced-names). For example, a tracing macro for [swift-distributed-tracing](https://swiftpackageindex.com/apple/swift-distributed-tracing) might want to introduce a local variable `span` that can be used by the rest of the function body to add more information to the trace. Such a macro could be declared as a preamble macro as follows:
```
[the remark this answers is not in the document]
DougGregor: Good catch, it will shadow---I'll add an example like this!

proposals/nnnn-function-body-macros.md
```diff
@@ -0,0 +1,338 @@
...
+Macros are a source-to-source transformation tool that have no effect on API resilience.
+
+## Alternatives considered
+
+### Eliminating preamble macros
+
+Preamble macros aren't technically necessary, because one could always write a function body macro that injects the preamble code into an existing body. However, preamble macros provide several end-user benefits over function body macros for the cases where they apply:
+
+* Preamble macros can be composed, whereas function body macros cannot.
+* Preamble macros are separately type-checked, so the function body as written can be type-checked independently of the macro expansion, providing a better user experience (e.g., for diagnostics, code completion, and so on).
```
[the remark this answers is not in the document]
DougGregor: You are correct. I introduced this inconsistency into the proposal when lifting the prohibition on preamble macros defining names. I'll fix this paragraph, thank you again!

proposals/nnnn-function-body-macros.md
DougGregor: Hmm. I think the expected behavior would be for preamble macros to still perform the implicit `return`. Body macros, on the other hand, would be responsible for adding the `return`.

DougGregor:
@allevato thank you for the feedback, I believe I've addressed it all in this last commit

allevato:
@DougGregor Looks great, thanks! If you can go ahead and add a brief statement about future directions or challenges involving closures that we talked about offline, then I can go ahead and kick it off.

DougGregor:
> @DougGregor Looks great, thanks! If you can go ahead and add a brief statement about future directions or challenges involving closures that we talked about offline, then I can go ahead and kick it off.

I've added the section, in the last commit, and removed the extraneous commit from the beginning. The current `main` snapshots have this feature behind the `BodyMacros` experimental flag.

allevato (approved):

PR #2222, opened 2023-12-01
Unsafe task-local binding

ktoso:
This provides building blocks necessary to build tracing macros (or similar) which use body "preamble" macros.

### DIFF
--- a/proposals/NNNN-unsafe-task-locals.md
+++ b/proposals/NNNN-unsafe-task-locals.md
@@ -43,7 +43,7 @@ func makeSushi() -> Sushi {
 ```
 
 This capability is used primarily to propagate additional metadata throughout a chain of calls, 
-without which the system still would work fine, but thanks to this metadata we can provide
+without which the system would still work fine, but thanks to this metadata we can provide
 enriched logging, security, or other semantics that improve the usability and understandability of the system.
 
 Notably, the `withValue` closure makes it possible for the runtime to reliably "push" and "pop" values onto the 
--- a/proposals/NNNN-unsafe-task-locals.md
+++ b/proposals/NNNN-unsafe-task-locals.md
@@ -1,0 +1,202 @@
+# Unsafe Task Local Binding
+
+* Proposal: SE-NNNN
+* Author: [Konrad 'ktoso' Malawski](https://github.com/ktoso)
+* Review Manager: TBD
+* Status: Implemented [PR #70062](https://github.com/apple/swift/pull/70062)
+* Review: TBD
+
+## Introduction
+
+With the introduction of Swift Concurrency and the concept of tasks to the language, Swift also gained the ability to
+associate values with the structured task hierarchy thanks to [Task Local Values (SE-0311)](https://github.com/apple/swift-evolution/blob/main/proposals/0311-task-locals.md).
+
+Task local values are a **safe** and powerful alternative to what in traditional thread-based concurrency systems
+would be handled by thread-local values, however they are also automatically inherited by child tasks created
+by the task which bound such value, and are automatically cleaned up when the task or scope completes.
+
+This proposal aims to introduce "unsafe" versions of the task local APIs in Swift.
+
+## Motivation
+
+Task local values achieve, in large part, their safety and _structured_ properties that work so well with
+Structured Concurrency by forcing the APIs to form scopes during which the value binding applies.
+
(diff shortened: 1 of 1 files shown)


proposals/NNNN-unsafe-task-locals.md
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func makeSushi() -> Sushi {
+    print("Preparing sushi for customer \(currentCustomer)")
+    // ... 
+}
+```
+
+This capability is used primarily to propagate additional metadata throughout a chain of calls, 
+without which the system still would work fine, but thanks to this metadata we can provide
```
slashmo, line 46:
```suggestion
without which the system would still work fine, but thanks to this metadata we can provide
```

proposals/NNNN-unsafe-task-locals.md
```diff
@@ -0,0 +1,202 @@
...
+  }
+}
+```
+
+can be expressed using the new unsafe APIs as follows:
+
+```swift
+func order() async {
+  $wasabiPreference.unsafePushValue(.with)
+  defer { TL.$number.unsafePopValue() }
```
slashmo, line 90: Where is `TL.$number` coming from? Shouldn't this be `$wasabiPreference.unsafePopValue()`?

proposals/NNNN-unsafe-task-locals.md
```diff
@@ -0,0 +1,202 @@
...
+}
+```
+
+which expands to the same piece of code we've seen before, using the unsafe push and pop method calls:
+
+```swift
+@WasabiPreference(.with)
+func order() async {
+  //   $wasabiPreference.unsafePushValue(.with)
+  //   defer { TL.$number.unsafePopValue() }
```
slashmo, line 117: Same here, I'm confused about the `TL.$number`.

proposals/NNNN-unsafe-task-locals.md
```diff
@@ -0,0 +1,202 @@
...
+  /// used instead of this pair of unsafe calls.
+  ///
+  /// Must only be used in conjunction with the
+  /// ``unsafePushValue(_:file:line:)`` method, to remove any value binding the
+  /// prior has established.
+  ///
+  /// - SeeAlso: ``withValue(_:file:line:)``
+  /// - SeeAlso: ``unsafePushValue(_:file:line:)``
+  @inlinable
+  public func unsafePopValue(file: String = #fileID, line: UInt = #line)
```
slashmo, line 183: Do you also want to include the body of this method?

slashmo (approved):
LGTM, the naming looks like it's enough of a barrier to keep folks from using it unless absolutely necessary 👍

proposals/NNNN-unsafe-task-locals.md
ktoso: Nope, we're not showing implementation details -- it is swift runtime builtins

proposals/NNNN-unsafe-task-locals.md
```diff
@@ -0,0 +1,202 @@
...
+leading to incorrect value bindings. Or even worse attempt to pop a value stored in a parent task (which can happen
+by popping more values than a task has pushed). 
+
+> **WARNING:** Mis-use of these APIs will lead to incorrect behavior, crashes, or even undefined behavior, so they must be handled very carefully. Preferably, only in macro generated code which can _guarantee_ the correct balance of push/pop calls.
+
+The following code using `withValue`:
+
+```swift
+func order() async {
+  $wasabiPreference.withValue(.with) {
```
xwu, line 79: Sorry, what is `.with` here? It is not mentioned anywhere in the text or declared in the examples—is this some existing concurrency API?

slashmo: It's the value set for the `wasabiPreference` task-local within the closure scope. I suppose there are `.with` and `.without`, but I think this should be written down as part of the proposal. CC @ktoso

The `$taskLocal.withValue` API is the existing (safe) API to bind task-local values.

rjmccall:
The Language Steering Group is going to sit on this until we figure out in the function-body macros proposal whether this sort of thing really has to be necessary.  I think we're hoping that it doesn't.

ktoso:
Sounds good, agreed!

ktoso:
I think since we have done function body replacement macros now, this proposal is not necessary - going to close it.

PR #2223, opened 2023-12-02
SE-0412 revision 2

sophiapoirier:


proposals/0412-strict-concurrency-for-global-variables.md
```diff
@@ -39,9 +47,24 @@ There may be need in some circumstances to opt out of static checking to enable
...
+    return value
+  }
+  print(await task.value)
+}
+```
+
 Because `nonisolated` is a contextual keyword, there is ambiguity when using `nonisolated(unsafe)` on a separate line immediately preceding a top-level variable declaration in script mode as it could also be the invocation of a function named `nonisolated` with argument `unsafe`. This ambiguity can be resolved by favoring the interpretation of `nonisolated` as a keyword if it has a single unlabeled argument of `unsafe` and precedes a variable declaration.
 
-Imported C or C++ global variables, unless they are immutable and `Sendable`, are treated as if they have unsafely opted out of isolation checking, depending upon the developer to know how to use them safely. There remain tools for enforcing safety, such as isolating it to a global actor using for example `__attribute__((swift_attr("@MainActor")))`, or wrapping access within a safer API that declares the correct isolation or locks appropriately.
+Importing a module via `@preconcurrency import` suppresses any potential errors resulting from data isolation checking of imported global variables that lack explicit concurrency annotations.
```
hborla, line 65: You have an example of this in the motivation section, but I think it's worth specifying here in the detailed design that any use of a concurrency-unsafe global variable will produce a warning at the use site.

PR #2228, opened 2023-12-05
Rename to consuming parameter modifier

jcavar:
Rename old spelling of `consume` parameter modifier to `consuming`


rjmccall:
Thanks!

PR #2234, opened 2023-12-07
Package Manager Allow Targets to Depend on Products in the Same Package

stackotter:
This is a WIP proposal based off the discussion in [this forum thread](https://forums.swift.org/t/pitch-swiftpm-allow-targets-to-depend-on-products-in-the-same-package/57717/32). @tammofreese created [an initial implementation](https://github.com/tammofreese/swift-package-manager/tree/use_internal_product_as_target_dependency_experiment) a year ago which is a great way to test out the proposed changes (although it may differ from the proposal slightly, and requires updating to Swift 5.9).

Forum thread: https://forums.swift.org/t/pitch-swiftpm-allow-targets-to-depend-on-products-in-the-same-package/57717/32
WIP implementation: https://github.com/tammofreese/swift-package-manager/tree/use_internal_product_as_target_dependency_experiment


stackotter:
Thanks for the feedback! I agree 👍 I've moved the detail after the code block now

PR #2264, opened 2024-01-05
Add a proposal to generalize `AsyncSequence` and `AsyncIteratorProtocol`.

hborla:


proposals/NNNN-generalize-async-sequence.md
```diff
@@ -0,0 +1,311 @@
...
+```
+
+Note that the default implementation of `nextElement` necessarily violates `Sendable` checking in order to pass `self` from a possibly-isolated context to a `nonisolated` one. Though this is generally unsafe, this is how calls to `next()` behave today, so existing conformances will maintain the behavior they already have. Implementing `nextElement()` directly will eliminate the unsafety.
+
+To enable conformances of `AsyncIteratorProtocol` to only implement `nextElement()`, a default implementation is also provided for `next()`:
+
+```swift
+extension AsyncIteratorProtocol {
+  @available(SwiftStdlib 5.11, *)
+  public mutating func next() async throws(Failure) -> Element? {
```
tevelee, line 239: The original `next()` function is not using typed throws, so for this function to be a valid default implementation of the `AsyncIteratorProtocol`'s `next()` requirement, we need to lose the `Failure` type.
```suggestion
  public mutating func next() async throws -> Element? {
```

proposals/NNNN-generalize-async-sequence.md
```diff
@@ -0,0 +1,311 @@
...
+
+Some `AsyncSequence`s can throw during iteration, and others never throw. To enable callers to only require `try` when the given sequence can throw, `AsyncSequence` and `AsyncIteratorProtocol` use a bespoke `@rethrows` attribute on the protocol. The `@rethrows` attribute does not allow generic constraints over the `throws` effect, which has also [prevented `AsyncSequence` from adopting primary associated types](https://forums.swift.org/t/se-0346-lightweight-same-type-requirements-for-primary-associated-types/55869/70). Primary associated types on `AsyncSequence` would enable hiding concrete implementation details behind constrained opaque or existential types, such as in transformation APIs on `AsyncSequence`:
+
+```swift
+extension AsyncSequence {
+  // 'AsyncThrowingMapSequence' is an implementation detail hidden from callers.
+  public func map<Transformed>(
+    _ transform: @Sendable @escaping (Element) async throws -> Transformed
+  ) -> some AsyncSequence<Element, any Error> { ... }
```
luca-bernardi, line 51: I think the returned async sequence's element should be `Transformed`:

```suggestion
  ) -> some AsyncSequence<Transformed, any Error> { ... }
```

proposals/NNNN-generalize-async-sequence.md
```diff
@@ -0,0 +1,311 @@
...
+* Authors: [Doug Gregor](https://github.com/douggregor), [Holly Borla](https://github.com/hborla)
+* Review Manager: TBD
+* Status: **Awaiting implementation**
+* Implementation: https://github.com/apple/swift/pull/70635
+* Review: ([pitch](https://forums.swift.org/t/pitch-generalize-asyncsequence-and-asynciteratorprotocol/69283))
+
+## Introduction
+
+This proposal generalizes `AsyncSequence` in two ways:
+1. The bespoke `@rethrows` attribute is replaced with proper `throws` polymorphism via adoption of typed throws.
```
Jumhyn, line 13: I do wonder how much discussion of `@rethrows` is appropriate given that it's never been an official feature. Perhaps makes sense to mention as an aside below but I don't know if it warrants callout up-front?

proposals/NNNN-generalize-async-sequence.md
```diff
@@ -0,0 +1,311 @@
...
+
+### Adopting primary associated types
+
+The `Element` and `Failure` associated types are promoted to primary associated types. This enables using constrained existential and opaque `AsyncSequence` and `AsyncIteratorProtocol` types, e.g. `some AsyncSequence<Element, Never>` or `any AsyncSequence<Element, any Error>`.
+
+### Adopting isolated parameters
+
+The `nextElement()` requirement abstracts over actor isolation using [isolated parameters](/proposals/0313-actor-isolation-control.md). For callers to `nextElement()` that pass an iterator value that cannot be transferred across isolation boundaries under [SE-0414: Region based isolation](/proposals/0414-region-based-isolation.md), the call is only valid if it does not cross an isolation boundary. Whether `nextElement()` shares the isolation of the caller depends on the isolated argument value, with generalized rules for `nonisolated` and global-actor isolation described below.
+
+#### Generalized `isolated` parameters
```
Jumhyn, line 177: Just a nudge to remove this (and probably makes sense to add a cross-reference once we have a permalink in place)

proposals/NNNN-generalize-async-sequence.md
```diff
@@ -0,0 +1,311 @@
...
+
+## Detailed design
+
+### Adopting typed throws
+
+Concrete `AsyncSequence` and `AsyncIteratorProtocol` types determine whether calling `next()` can `throw`. This can be described in each protocol with a `Failure` associated type that is thrown by the `AsyncIteratorProtcol.nextElement()` requirement. Describing the thrown error with an associated type allows conformances to fulfill the requirement with a type parameter, which means that libraries do not need to expose separate throwing and non-throwing concrete types that otherwise have the same async iteration functionality.
+
+#### Error type inference from `for try await` loops
+
+The `Failure` associated type is only accessible at runtime in the Swift 5.11 standard library; code running against older standard library versions does not include the `Failure` requirement in the witness tables for `AsyncSequence` and `AsyncIteratorProtocol` conformances. This impacts error type inference from `for try await` loops.
```
Jumhyn, line 114: I think it would potentially make more sense to move this below a couple paragraphs, once we've introduced what the inference behavior actually is (i.e., as a caveat to the `MyAsyncIterator` example, and just before "When the throw error type of an `AsyncIteratorProtocol`..."

proposals/NNNN-generalize-async-sequence.md
```diff
@@ -0,0 +1,311 @@
...
+* the type coercion operator as (in the cases where it doesn't perform a dynamic bridging conversion); and
+implicit type conversions such as promotion to `Optional` type.
+
+#### Calling `nextElement()` in async `for-in` loops
+
+Desugared async `for-in` loops will call `AsyncIteratorProtocol.nextElement()` instead of `next()` when the context has appropriate availability, and pass in an isolated argument value of type `(any Actor)?`. The argument value always represents the isolation of the caller so that the call does not cross an isolation boundary:
+
+* if the caller's context is `nonisolated`, the argument value is `nil`;
+* if the caller's context is isolated to some global actor `T`, the argument value is `T.shared`;
+* if the caller's context has an isolated parameter `param`, the argument value is `param`;
```
Jumhyn, line 206: I suppose this technically covers actor methods but might be good to spell it out explicitly

proposals/NNNN-generalize-async-sequence.md
```diff
@@ -0,0 +1,311 @@
...
+      self = unsafeIterator
+      return element
+    } catch {
+      throw error as! Failure
+    }
+  }
+}
+```
+
+Note that the default implementation of `nextElement` necessarily violates `Sendable` checking in order to pass `self` from a possibly-isolated context to a `nonisolated` one. Though this is generally unsafe, this is how calls to `next()` behave today, so existing conformances will maintain the behavior they already have. Implementing `nextElement()` directly will eliminate the unsafety.
```
Jumhyn, line 232: Of course, calls to `next()` today produce a warning under strict concurrency checking, would this default end up swallowing those warnings?

proposals/NNNN-generalize-async-sequence.md
```diff
@@ -0,0 +1,311 @@
...
+
+### Associated type inference for `AsyncIteratorProtocol` conformances
+
+When an `AsyncIteratorProtocol`-conforming type provides a `nextElement` function, the `Failure` type is inferred based on whether (and what) `nextElement` throws using the rules described in [SE-0413](/proposals/0413-typed-throws.md).
+
+If the `AsyncIteratorProtocol`-conforming type uses the default implementation of `nextElement`, then the `Failure` associated type is inferred from the `next` function instead based on the following rules:
+
+* If `next()` throws nothing, `Failure` is inferred to `Never`.
+* If `next()` throws, `Failure` is inferred to `any Error`.
+* If `next()` rethrows, `Failure` is inferred to `T.Failure`, where `T` is the first type parameter with a conformance to either `AsyncSequence` or `AsyncIteratorProtocol`. If there are multiple such requirements, take the `errorUnion` of them all.
```
Jumhyn, line 260: Also here—if `rethrows` for `next()` isn't a 'real' thing, do we need to spell it out explicitly in this list?

proposals/NNNN-generalize-async-sequence.md
hborla: Swift 5.9 doesn't diagnose calls to `next()`. The call wasn't diagnosed until https://github.com/apple/swift/pull/67730 which I then suppressed in https://github.com/apple/swift/pull/70750 because there's not a good workaround, and I've never seen anybody actually observe a data race on the iterator. In any case, the warnings are already swallowed so this proposal doesn't make any changes there.

jumhyn-browser: Ah great! Saw the warning when building against ~nightly at work but didn't realize it had been suppressed

Jumhyn: Ah great! Saw the warning when building against ~nightly at work but didn't realize it had been suppressed

proposals/NNNN-generalize-async-sequence.md
hborla: I think we can now remove this whole list and say `#isolation` is used as the argument value. This list effectively describes what `#isolation` expands to, which is covered in the actor isolation inheritance proposal.

proposals/NNNN-generalize-async-sequence.md
hborla: I changed this to just say proper `throws` polymorphism is accomplished with typed throws adoption.

proposals/NNNN-generalize-async-sequence.md
hborla: Oh hmm, I meant this as a general statement that doesn't just apply to error type inference in `for try await` loops. Maybe I should move this up into the previous section to describe why the `Failure` associated type has Swift 5.11 availability. Then this section can say that because `Failure` is only available in Swift 5.11, thrown error type inference depends on availability context. Thoughts?

Jumhyn: Yes that reorganization makes sense to me!

PR #2273, opened 2024-01-13
Actor isolation inheritance.

hborla:
This is the same as https://github.com/apple/swift-evolution/pull/2210 with `@inheritsIsolation` moved to future directions.

### DIFF
--- a/proposals/NNNN-inheritance-of-actor-isolation.md
+++ b/proposals/NNNN-inheritance-of-actor-isolation.md
@@ -353,7 +353,7 @@ actor MyActor {
   var counter = Counter()
 }
 
-extenstion MyActor {
+extension MyActor {
   func testActor(other: MyActor) {
     // allowed
     await counter.incrementAndSleep(isolation: self)
--- a/proposals/NNNN-inheritance-of-actor-isolation.md
+++ b/proposals/NNNN-inheritance-of-actor-isolation.md
@@ -5,7 +5,7 @@
 * Review Manager: TBD
 * Status: **Awaiting implementation**
 * Implementation: TBD
-* Review: ([pitch](https://forums.swift.org/...))
+* Review: ([pitch](https://forums.swift.org/t/pitch-inheriting-the-callers-actor-isolation/68391))  
 
 **This is a draft design which has not been implemented.  It's very
 possible that some of the ideas in it simply do not work.**
--- a/proposals/NNNN-inheritance-of-actor-isolation.md
+++ b/proposals/NNNN-inheritance-of-actor-isolation.md
@@ -343,7 +343,7 @@ extension Counter {
   /// forbidden because it requires sharing a non-Sendable value
   /// between concurrency domains.  Inheriting isolation makes it
   /// okay.  This is a contrived example chosen for its simplicity.
-  func incrementAndSleep(isolation: (any Actor)?) async {
+  func incrementAndSleep(isolation: isolated (any Actor)?) async {
     count += 1
     await Task.sleep(nanoseconds: 1_000_000)
   }
--- a/proposals/NNNN-inheritance-of-actor-isolation.md
+++ b/proposals/NNNN-inheritance-of-actor-isolation.md
@@ -1,0 +1,653 @@
+# Inheritance of actor isolation
+
+* Proposal: [SE-NNNN](NNNN-inheritance-of-actor-isolation.md)
+* Authors: [John McCall](https://github.com/rjmccall)
+* Review Manager: TBD
+* Status: **Awaiting implementation**
+* Implementation: TBD
+* Review: ([pitch](https://forums.swift.org/...))
+
+**This is a draft design which has not been implemented.  It's very
+possible that some of the ideas in it simply do not work.**
+
+[SE-0302]: https://github.com/apple/swift-evolution/blob/main/proposals/0302-concurrent-value-and-concurrent-closures.md
+[SE-0304-propagation]: https://github.com/apple/swift-evolution/blob/main/proposals/0304-structured-concurrency.md#actor-context-propagation
+[SE-0306]: https://github.com/apple/swift-evolution/blob/main/proposals/0306-actors.md
+[SE-0313]: https://github.com/apple/swift-evolution/blob/main/proposals/0313-actor-isolation-control.md
+[SE-0316]: https://github.com/apple/swift-evolution/blob/main/proposals/0316-global-actors.md
+[SE-0338]: https://github.com/apple/swift-evolution/blob/main/proposals/0338-clarify-execution-non-actor-async.md
+[SE-0392]: https://github.com/apple/swift-evolution/blob/main/proposals/0392-custom-actor-executors.md
+
+## Introduction
+
+Under Swift's [actors design][SE-0306], every function in Swift has
+an actor isolation: it is either isolated to some specific actor or
+non-isolated.  It is sometimes useful to be able to give a function
+the same actor isolation as its caller, either to give it access to
+actor-isolated data or just to avoid unnecessary suspensions.  This
+proposal allows `async` functions to opt in to this behavior.
+
+## Motivation
+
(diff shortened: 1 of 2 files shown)


proposals/NNNN-inheritance-of-actor-isolation.md
```diff
@@ -0,0 +1,551 @@
...
+isolation checking is performed separately from type-checking.
+
+## ABI compatibility
+
+This proposal does not change how any existing code is compiled.
+
+## Implications for adoption
+
+This proposal does not add any new types and does not require new
+runtime or library support.  It can be implemented purely in the compiler.
```
xwu, line 400: As noted in the PR for `#isolation`, when implemented as an expression macro, it would tie this feature (and hence, any async features that depend on it) to platform support for macros. Ought we to consider platform bring-up implications here?

proposals/NNNN-inheritance-of-actor-isolation.md
```diff
@@ -0,0 +1,551 @@
...
+function types.  With such a feature, you could declare a value to have
+type, say, `@isolated(myActor) () -> ()`, where `myActor` is a `let`
+constant in the local scope.  This kind of value dependence, however,
+is a large step in complexity for a type system, and it's not a likely
+path for Swift in the foreseeable future.
+
+A more promising approach would be to allow the isolation to be
+statically erased but still make it dynamically recoverable by carrying
+it along in the function value, essentially as an extra value of type
+`(any Actor)?`.  This would look something like `@isolated () -> ()`,
```
xwu, line 492: Drafting nit: "This" here, grammatically, refers to the preceding "value of type `(any Actor)?`," which clearly cannot be anything like `@isolated () -> ()`. Could the subject of this sentence be spelled out concretely?

proposals/NNNN-inheritance-of-actor-isolation.md
```diff
@@ -0,0 +1,551 @@
...
+nice for functions like `sequentialMap` that should probably be isolated
+not to their *caller* but to the *function they've been passed*.
+
+## Alternatives Considered
+
+### Allowing isolation to `SerialExecutor` types
+
+This proposal observes that it is more efficient to pass down an
+`UnownedSerialExecutor` value instead of an actor reference and suggests
+that `@inheritsIsolation` be implemented this way.  However, if a function
```
xwu, line 507: Now that `@inheritsIsolation` is moved to "Future directions" without a strawman syntax, this is now the first reference to that attribute; probably just strike this second half of the sentence? Some of the following text may also need to be reworked (for example, there is nothing _but_ the explicit isolation-polymorphism pattern in this proposal now, so the part of the following sentence that says "_if_ a function wants to use" this pattern is inoperative).

proposals/NNNN-inheritance-of-actor-isolation.md
```diff
@@ -0,0 +1,551 @@
...
+  be `Sendable`.  Collections are generally `Sendable` if their
+  element types are `Sendable`, so in short, this restriction
+  prevents `sequentialMap` from being used from an actor to work
+  on a collection of values isolated to the actor.
+
+- Finally, the `transform` argument to the call must be `Sendable`.
+  Functions are generally `Sendable` if they don't capture any
+  `Sendable` values, so in short, this restriction prevents
+  `sequentialMap` from using any other data that's isolated to
+  the actor.
```
PsychoH13, line 138: I think there's a typo here, I'm not 100% sure about this but I'm pretty sure that if a function captured only sendable values (or capture nothing) the function would be sendable as well. So shouldn't this say this instead?
```suggestion
- Finally, the `transform` argument to the call must be `Sendable`.
  Functions are generally `Sendable` if they don't capture any
  non-`Sendable` values, so in short, this restriction prevents
  `sequentialMap` from using any other data that's isolated to
  the actor.
```

proposals/NNNN-inheritance-of-actor-isolation.md
```diff
@@ -0,0 +1,551 @@
...
+  protocol requirement or the concrete declaration which provides it
+  in `T`'s conformance to `GlobalActor`.
+
+An expression is a non-optional derivation of an isolated parameter
+`param` if it is:
+- `param?` (the optional-chaining operator);
+- `param!` (the force-unwrapping operator); or
+- a reference to a *non-optional binding* of `param`, i.e. a `let`
+  constant initialized by a successful pattern-match which removes
+  the optionality from `param`, such as `ref` in `if let ref = param`.
```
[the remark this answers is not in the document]
hborla: No, because the derivation must be non-optional. You might imagine supporting something like this:

```swift
func test(param: isolated (any Actor)?) {
  if let value = param ?? nil {
    // 'value' could be considered a non-optional derivation of 'param' here
  }
}
```

But the proposal does not support looking through any operators other than `?` and `!`.

proposals/NNNN-inheritance-of-actor-isolation.md
hborla: You're right, thank you! I've reworded this, is it clearer now?

proposals/NNNN-inheritance-of-actor-isolation.md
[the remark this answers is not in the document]
xwu: Sorry, poorly phrased; consider the question to apply to line 249—

Given the following:

```swift
func f(_ param: isolated (any Actor)?) {
  g(<<< expression >>>)
}
func g(_ param: isolated (any Actor)?) { }
```

...the proposal states that `g` shares an isolation context if `<<< expression >>>` is `param` or `param!`, the latter because it is a "non-optional derivation" which is then implicitly promoted back to an `Optional` type. For completeness (cf. note 5 about pointlessness), what about `param ?? nil`.

proposals/NNNN-inheritance-of-actor-isolation.md
```diff
@@ -0,0 +1,551 @@
...
+  Functions are generally `Sendable` if they don't capture any
+  `Sendable` values, so in short, this restriction prevents
+  `sequentialMap` from using any other data that's isolated to
+  the actor.
+
+In summary, these restrictions unnecessarily limit the capability of
+the API when used from an isolated context.[^4]  Furthermore, even
+if the API is usable (e.g. because all the types involved are
+`Sendable`), it may be unexpectedly inefficient if, say, the
+function argument is an isolated function.
```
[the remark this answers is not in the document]
xwu: Super!

proposals/NNNN-inheritance-of-actor-isolation.md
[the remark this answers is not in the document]
hborla: Ah, I see your point, thank you for elaborating! Still the answer is that only the forms mentioned in the proposal are supported, so that does not include `param ?? nil`. I'm reluctant to start adding other operators to the list, because this will not generalize to arbitrary operators and functions. I think whether or not that's the right choice is fair to raise during the review.

hborla:
@xwu I believe this is ready for review. The toolchains were built in https://github.com/apple/swift/pull/71143

* Linux: https://ci.swift.org/job/swift-PR-toolchain-Linux/713/
* Windows: https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/1121/
* macOS: https://ci.swift.org/job/swift-PR-toolchain-macos/1041/

xwu:
@hborla Just to check—the toolchain in question has the featured gated under `OptionalIsolatedParameters` and doesn't include support for `#isolation` as the default argument of a function parameter, correct?

hborla:
> @hborla Just to check—the toolchain in question has the featured gated under OptionalIsolatedParameters and doesn't include support for #isolation as the default argument of a function parameter, correct?

Yes, that's correct!

xwu:
Good to know! The Linux job at the URL provided looks like it's missing a link to the build artifacts (or at least I can't find it with some cursory digging)—wonder if there's another way to get at it.

PR #2281, opened 2024-01-24
[SE-0413] Fix minor typo

jeonghi:
1) Corrected `themselve` to `themselves` to fix a spelling error.

2) Modified the sentence to convey the intended meaning more clearly. 
`In general we want to add the possibility to use throws with a single, specific error type.` -> `In general, we want to add the possibility of using throws with a single, specific error type.` 

3) Removed the duplicated `a` to fix the grammar issue.

4) Corrected `an function` to `a function` to fix a grammar issue.

### DIFF
--- a/proposals/0413-typed-throws.md
+++ b/proposals/0413-typed-throws.md
@@ -9,7 +9,7 @@
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


DougGregor (approved):
Thank you!

PR #2289, opened 2024-01-27
Update status of SE-0421 to 'Active review'

dempseyatgithub:
The active review of SE-0421 began yesterday but still displays as 'scheduled' on the evolution dashboard.

(Review announcement on Swift forums: https://forums.swift.org/t/se-0421-generalize-effect-polymorphism-for-asyncsequence-and-asynciteratorprotocol/69662)


DougGregor (approved):
Thank you!

PR #2294, opened 2024-01-30
Custom isolation checking for SerialExecutor

ktoso:
Work in progress, PR for discussion of draft.

This proposal aims to provide customization points for executor isolation "checks".

### DIFF
--- a/proposals/NNNN-custom-isolation-checking-for-serialexecutor.md
+++ b/proposals/NNNN-custom-isolation-checking-for-serialexecutor.md
@@ -9,31 +9,24 @@
 
 ## Introduction
 
-Swift introduced custom actor executors in [SE-0392: Custom Actor Executors](https://github.com/apple/swift-evolution/blob/main/proposals/0392-custom-actor-executors.md), and ever since allowed further customization of isolation and execution semantics of actors.
-
-This proposal also introduced a family of assertion and assumption APIs which are able to dynamically check the isolation of a currently executing task. These APIs are:
-
-- Asserting isolation context:
-  - [`Actor/assertIsolated(_:file:line:)`](https://developer.apple.com/documentation/swift/actor/assertisolated(_:file:line:))
-  - [`Actor/preconditionIsolated(_:file:line:)`](https://developer.apple.com/documentation/swift/actor/preconditionisolated(_:file:line:))
-  - [`DistributedActor/assertIsolated(_:file:line:)`](https://developer.apple.com/documentation/distributed/distributedactor/preconditionisolated(_:file:line:))
-  - [`DistributedActor/preconditionIsolated(_:file:line:)`](https://developer.apple.com/documentation/distributed/distributedactor/preconditionisolated(_:file:line:))
-- Assuming isolation context, and obtaining an `isolated actor` reference of the target actor
-  - [`Actor/assumeIsolated(_:file:line:)`](https://developer.apple.com/documentation/swift/actor/assumeisolated(_:file:line:))
-  - [`DistributedActor/assumeIsolated(_:file:line:)`](https://developer.apple.com/documentation/distributed/distributedactor/assumeisolated(_:file:line:))
+[SE-0392 (Custom Actor Executors)](https://github.com/apple/swift-evolution/blob/main/proposals/0392-custom-actor-executors.md) added support for custom actor executors, but its support is incomplete. Safety checks like [`Actor.assumeIsolated`](https://developer.apple.com/documentation/swift/actor/assumeisolated(_:file:line:)) work correctly when code is running on the actor through a task, but they don't work when code is scheduled to run on the actor's executor through some other mechanism. For example, if an actor uses a serial `DispatchQueue` as its executor, a function dispatched _directly_ to the queue with DispatchQueue.async cannot use `assumeIsolated` to assert that the actor is currently isolated. This proposal fixes this by allowing custom actor executors to provide their own logic for these safety checks.
 
 ## Motivation
 
-All the above mentioned APIs rely on an internal capability of the Swift concurrency runtime to obtain the "current serial executor", and compare it against the expected executor. Additional comparison modes such as "complex equality" are also supported, which help executors that e.g. share a single thread across multiple executor instances to still be able to correctly answer the "are we on the same executor?" question when different executor *instances* are being compared, however in reality they utilize the same threading resource.
+The Swift concurrency runtime already provides means of dynamically tracking and checking the tasks's current executor, and APIs like `assertIsolated` and `assumeIsolated` are built on top of that functionality.
 
-The proposal did not account for the situation in which the Swift concurrency runtime has no notion of "current executor" though, causing the following situation to -- perhaps surprisingly -- result in runtime crashes reporting an isolation violation, while in reality, no such violation takes place in the following piece of code:
+In those APIs the expected executor is compared with the "current" executor of a current task which is tracked by the runtime. If the code calling the comparison logic is not running within a Swift concurrency task, the comparison fails, as there are no two executors that can be compared.
+
+This logic is not sufficient to be able to handle the situation in which code is runnin on the correct queue or thread, and mutual exclusion is properly ensured, however, the calling code is not being called from a task and therefore the check fails unexpectedly. The following example demonstrates such situation:
 
 ```swift
 import Dispatch
 
 actor Caplin {
   let queue: DispatchSerialQueue(label: "CoolQueue")
-  
+
+  var num: Int // actor isolated state
+
   // use the queue as this actor's `SerialExecutor`
   nonisolated var unownedExecutor: UnownedSerialExecutor {
     queue.asUnownedSerialExecutor()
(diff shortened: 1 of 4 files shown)


proposals/NNNN-custom-isolation-checking-for-serialexecutor.md
```diff
@@ -0,0 +1,190 @@
...
+  /// to prove that the calling context can be safely assumed to be the same isolation 
+  /// context as represented by this ``SerialExecutor``.
+  ///
+  /// A default implementation is provided that unconditionally causes a fatal error.
+  func checkIsolation()
+}
+
+extension SerialExecutor {
+  public func checkIsolation() {
+    fatalError("Incorrect actor executor assumption, expected: \(self)")
```
FranzBusch, line 86: Can we change this fatal error here since isolation checking is not only for actor executors but can also be done for task executors going forward if they are also `SerialExecutors`.

ktoso: isolation is specifically handled by serial executors; not sure what you mean, should we discuss on the forums with some examples? https://forums.swift.org/t/pitch-custom-isolation-checking-for-serialexecutor/69786/

proposals/NNNN-custom-isolation-checking-for-serialexecutor.md
```diff
@@ -0,0 +1,190 @@
...
+This proposal also introduced a family of assertion and assumption APIs which are able to dynamically check the isolation of a currently executing task. These APIs are:
+
+- Asserting isolation context:
+  - [`Actor/assertIsolated(_:file:line:)`](https://developer.apple.com/documentation/swift/actor/assertisolated(_:file:line:))
+  - [`Actor/preconditionIsolated(_:file:line:)`](https://developer.apple.com/documentation/swift/actor/preconditionisolated(_:file:line:))
+  - [`DistributedActor/assertIsolated(_:file:line:)`](https://developer.apple.com/documentation/distributed/distributedactor/preconditionisolated(_:file:line:))
+  - [`DistributedActor/preconditionIsolated(_:file:line:)`](https://developer.apple.com/documentation/distributed/distributedactor/preconditionisolated(_:file:line:))
+- Assuming isolation context, and obtaining an `isolated actor` reference of the target actor
+  - [`Actor/assumeIsolated(_:file:line:)`](https://developer.apple.com/documentation/swift/actor/assumeisolated(_:file:line:))
+  - [`DistributedActor/assumeIsolated(_:file:line:)`](https://developer.apple.com/documentation/distributed/distributedactor/assumeisolated(_:file:line:))
```
rjmccall, line 23: I don't think you really need this exhaustive list of APIs; you could just say something like ", such as `assumeIsolated` and `preconditionIsolated`".

The break here between sections is awkward.  This is all really a detailed breakdown of the problem and belongs in Motivation; the Introduction is supposed to be more of an "elevator pitch".  I would suggest something like this:

> SE-0392 added support for custom actor executors, but its support is incomplete.  Safety checks like `assumeIsolated` work correctly when code is running on the actor through a `Task`, but they don't work when code is scheduled to run on the actor's executor through some other mechanism.  For example, if an actor uses a serial `DispatchQueue` as its executor, a function dispatched to the queue with `DispatchQueue.async` cannot use `assumeIsolated` to assert that the actor is currently isolated.  This proposal fixes this by allowing executors to provide custom logic for these safety checks.

proposals/NNNN-custom-isolation-checking-for-serialexecutor.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Motivation
+
+All the above mentioned APIs rely on an internal capability of the Swift concurrency runtime to obtain the "current serial executor", and compare it against the expected executor. Additional comparison modes such as "complex equality" are also supported, which help executors that e.g. share a single thread across multiple executor instances to still be able to correctly answer the "are we on the same executor?" question when different executor *instances* are being compared, however in reality they utilize the same threading resource.
+
+The proposal did not account for the situation in which the Swift concurrency runtime has no notion of "current executor" though, causing the following situation to -- perhaps surprisingly -- result in runtime crashes reporting an isolation violation, while in reality, no such violation takes place in the following piece of code:
```
rjmccall, line 29: I think you can tell a somewhat more focused narrative here.  I'd suggest a structure something like this:
- The runtime already needs to dynamically track a task's current executor.
- `assumeIsolated` etc. are primarily built on top of that.
- That's good enough for uses from tasks, like when a task calls a synchronous non-isolated function.
- It's not good enough when code can be scheduled on the executor in other ways.
- The actor is still isolated in these other cases, but the runtime doesn't know about it because the scheduling didn't come through the runtime.
- This was a common and important enough situation that we added a special case for `MainActor`.
- But it also comes up with `DispatchQueue` and other serial executors, and these executors generally offer their own ways to assert that you're on the right executor.
- It makes sense to generalize that by allowing the executor to provide its own logic for safely asserting the precondition.

I don't think stuff like complex equality is really important to include.

ktoso: That sounds good, I'll rephrase the narrative -- it's good to compare this to our special casing of the main actor 👍

proposals/NNNN-custom-isolation-checking-for-serialexecutor.md
```diff
@@ -0,0 +1,182 @@
...
+- obtain current executor
+  - if no current executor exists, ​​use heurystics to detect the "main actor" executor 
+    - These heurystics could be removed by using this proposal's `checkIsolation()` API, however we'll first need to expose the MainActor's SerialExecutor as a global property which this proposal does not cover. Please see **Future Directions** for more discussion of this topic.
+  - if a current executor exists, perform basic object comparison between them
+- if unable to prove the executors are equal:
+  - compare the executors using "complex equality" (see [SE-0392: Custom Actor Executors](https://github.com/apple/swift-evolution/blob/main/proposals/0392-custom-actor-executors.md) for a detailed description of complex exector equality)
+- if still unable to prove the executors are equal:
+  - :arrow_right: call the expected executor's `checkIsolation()` method
+
+The last step of this used to be just to unconditionally fail the comparison, leaving no space for an executor to take over and use whatever it's own tracking -- usually expressed using thread-locals the executor sets as it creates its own worker thread -- to actually save the comparison from failing.
```
rjmccall, line 100:
```suggestion
The last step of this used to be just to unconditionally fail the comparison, leaving no space for an executor to take over and use whatever its own tracking -- usually expressed using thread-locals the executor sets as it creates its own worker thread -- to actually save the comparison from failing.
```

proposals/NNNN-custom-isolation-checking-for-serialexecutor.md
```diff
@@ -0,0 +1,182 @@
...
+// Dispatch 
+
+extension DispatchSerialQueue { 
+  public func checkIsolated(message: String) {
+    dispatchPrecondition(condition: .onQueue(self)) // existing Dispatch API
+  }
+}
+```
+
+Other executors would have the same capability, if they used some mechanisms to identify their own worker threads.
```
rjmccall, line 114:
```suggestion
An executor that wishes to take advantage of this proposal will need to have some mechanism to identity its active worker thread.  If that's not possible or desired, the executor should leave the default implementation (that unconditionally crashes) in place.
```

proposals/NNNN-custom-isolation-checking-for-serialexecutor.md
```diff
@@ -0,0 +1,182 @@
...
+
+### Impact on async code and isolation assumtions
+
+The `assumeIsolated(_:file:line:)` APIs purposefully only accept a **synchronous** closure. This is correct, and with the here proposed additions, it remains correct -- we may be executing NOT inside a Task, however we may be isolated and can safely access actor state.
```
rjmccall, line 118:
```suggestion
The `assumeIsolated(_:file:line:)` APIs purposefully only accept a **synchronous** closure. This is correct, and it remains correct with these proposed additions.  An isolation check on an executor ensures that any actor using the executor is synchronously isolated, and the closure provided to `assumeIsolated` will execute prior to any possible async suspension.  This is what makes it safe to access actor-isolated state within the closure.
```

proposals/NNNN-custom-isolation-checking-for-serialexecutor.md
```diff
@@ -0,0 +1,182 @@
...
+    return self.assumeIsolated { // () throws -> Int
+      // suspensions are not allowed in this closure.
+      
+      self.number // we are guaranteed to be isolated on this actor; read is safe
+    }
+  }
+      
+```
+
+As such, there is no negative impact on the correctness of these APIs.
```
rjmccall, line 138:
```suggestion
As such, there is no negative impact on the correctness of these APIs.

Asynchronous functions should not use isolation checking.  Isolation
checking is useful in synchronous functions because they naturally inherit
execution properties like their caller's isolation without disturbing it.
A synchronous function may be formally non-isolated and yet actually
run in an isolated context dynamically.  This is not true for asynchronous
functions, which switch to their formal isolation on entry without regard
to their caller's isolation.  If an asynchronous function is not formally
isolated to an actor, its execution will never be dynamically in an
isolated context, so there's no point in checking for it.
```

proposals/NNNN-custom-isolation-checking-for-serialexecutor.md
```diff
@@ -0,0 +1,182 @@
...
+}
+```
+
+## Detailed design
+
+This proposal adds another customization point to the Swift concurrency runtime that hooks into isolation context comparison mechanisms used by `assertIsolated`, `preconditionIsolated`, `assumeIsolated` as well as implicitly injected assertions used in `@preconcurrency` code.
+
+### Extended executor comparison mechanism
+
+With this proposal, the logic for checking if the "current" executor is the same as the "expected" executor changes becomes as follows:
```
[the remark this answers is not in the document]
ktoso: It'll be extra work to handle the main actor's executor, but we should do it in the future; I'll pretend we can do so.

proposals/NNNN-custom-isolation-checking-for-serialexecutor.md
ktoso:
```suggestion
As such, there is no negative impact on the correctness of these APIs.

Asynchronous functions should not use dyanamic isolation checking.  Isolation
checking is useful in synchronous functions because they naturally inherit
execution properties like their caller's isolation without disturbing it.
A synchronous function may be formally non-isolated and yet actually
run in an isolated context dynamically.  This is not true for asynchronous
functions, which switch to their formal isolation on entry without regard
to their caller's isolation.  If an asynchronous function is not formally
isolated to an actor, its execution will never be dynamically in an
isolated context, so there's no point in checking for it.
```

added "dynamic", happy to take this edit otherwise

PR #2308, opened 2024-02-06
Synchronous Mutual Exclusion Lock

Azoy:
This proposal introduces a new type in the standard library's synchronization module, `Mutex`.

### DIFF
--- a/proposals/nnnn-mutex.md
+++ b/proposals/nnnn-mutex.md
@@ -90,7 +90,7 @@ public struct Mutex<Value: ~Copyable>: ~Copyable {
   /// Initializes a value of this mutex with the given initial state.
   ///
   /// - Parameter initialValue: The initial value to give to the mutex.
-  public init(_: consuming Value)
+  public init(_: transferring consuming Value)
   
   /// Attempts to acquire the lock and then calls the given closure if
   /// successful.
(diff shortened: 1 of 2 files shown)


proposals/nnnn-mutex.md
```diff
@@ -0,0 +1,480 @@
...
+let globalCache = Mutex<[MyKey: MyValue]>([:])
+```
+
+## Detailed design
+
+### Underlying System Mutex Implementation
+
+The `Mutex` type proposed is a wrapper around a platform's implementation.
+
+* macOS, iOS, watchOS, tvOS, visionOS:
```
Joannis, line 48: What's the behaviour when compiling for WASM? Will Mutex be available on Embedded?

Azoy: WASM has support for atomic operations and even a `wait` and `wake` so we can reasonably build something like Linux's implementation on WASM. I'm unsure if those features are only available for X version of WASM however. Embedded doesn't include the Synchronization module at all right now, so Mutex will not be available yet/ever.

Azoy: It looks like this support is implemented in some places, but the actual proposal hasn't been accepted yet...?

Azoy: Ah looks like it's in phase 4 of the proposal which is standardize the thing https://github.com/WebAssembly/proposals

Azoy: @MaxDesiatov points out that this feature is implemented in all browsers at this point though https://webassembly.org/features/

MaxDesiatov: cc @kateinoigakukun

kateinoigakukun: LLVM also implements the [`*.atomic.*` instructions](https://github.com/WebAssembly/threads/blob/main/proposals/threads/Overview.md) in their backend and it's enabled only when a specific CPU feature (`-matomics`) is enabled. So for both Embedded/non-Embedded Wasm targets, it would be best to have `Mutex` implementation when the feature is enabled. (Also if we can provide no-op implementation when the feature is not enabled, it can reduce a large amount of `#if` in the use-site)

We can have separate Swift SDKs for each feature set configuration, so we don't need to care Wasm's versioning here.

Anyway, I can help enable this module for WebAssembly targets.

proposals/nnnn-mutex.md
```diff
@@ -0,0 +1,348 @@
+# Synchronous Mutual Exclusion Lock 🔒
+
+* Proposal: [SE-NNNN](NNNN-mutex.md)
+* Author: [Alejandro Alonso](https://github.com/Azoy)
+* Review Manager: TBD
```
stephentyrone, line 5:
```suggestion
* Review Manager: [Stephen Canon](http://github.com/stephentyrone)
```

PR #2314, opened 2024-02-09
[BitwiseCopyable] Pitch.

nate-chandler:
A new marker protocol to identify types that can be trivially copied, moved, and destroyed.

### DIFF
--- a/proposals/nnnn-bitwise-copyable.md
+++ b/proposals/nnnn-bitwise-copyable.md
@@ -300,7 +300,7 @@ The following types in the standard library will gain the `BitwiseCopyable` cons
   - `UInt8.Words`, `UInt16.Words`, `UInt32.Words`, `UInt64.Words`, `UInt.Words`
   - `Int8.Words`, `Int16.Words`, `Int32.Words`, `Int64.Words`, `Int.Words`
 - The fixed-precision floating-point types:
-  - `Float`, `Double`, `Float80`
+  - `Float`, `Double`, `Float16`, `Float80`
   - `FloatingPointSign`, `FloatingPointClassification`
 - The family of `SIMDx<Scalar>` types
 - The family of unmanaged pointer types:
--- a/proposals/nnnn-bitwise-copyable.md
+++ b/proposals/nnnn-bitwise-copyable.md
@@ -1,0 +1,270 @@
+# BitwiseCopyable
+
+* Proposal: [SE-NNNN](NNNN-filename.md)
+* Authors: [Kavon Farvardin](https://github.com/kavon), [Guillaume Lessard](https://github.com/glessard), [Nate Chandler](https://github.com/nate-chandler), [Tim Kientzle](https://github.com/tbkka)
+* Review Manager: TBD
+* Implementation: On `main` gated behind `-enable-experimental-feature BitwiseCopyable`
+
+<!-- *During the review process, add the following fields as needed:*
+
+* Implementation: [apple/swift#NNNNN](https://github.com/apple/swift/pull/NNNNN) or [apple/swift-evolution-staging#NNNNN](https://github.com/apple/swift-evolution-staging/pull/NNNNN)
+* Decision Notes: [Rationale](https://forums.swift.org/), [Additional Commentary](https://forums.swift.org/)
+* Bugs: [SR-NNNN](https://bugs.swift.org/browse/SR-NNNN), [SR-MMMM](https://bugs.swift.org/browse/SR-MMMM)
+* Previous Revision: [1](https://github.com/apple/swift-evolution/blob/...commit-ID.../proposals/NNNN-filename.md)
+* Previous Proposal: [SE-XXXX](XXXX-filename.md) -->
+
+## Introduction
+
+We propose a new marker protocol `BitwiseCopyable` that can be conformed to by types that can be moved or copied with direct calls to `memcpy` and which require no special destroy operation[^1].
+When compiling generic code with such constraints, the compiler can emit these efficient operations directly, only requiring minimal overhead to look up the size of the value at runtime.
+Alternatively, developers can use this constraint to selectively provide high-performance variations of specific operations, such as bulk copying of a container.
+
+[^1]: The term "trivial" is used in [SE-138](0138-unsaferawbufferpointer.md) and [SE-0370](0370-pointer-family-initialization-improvements.md) to refer to types with the property above. The discussion below will explain why certain generic or resilient types that are trivial will not in fact be `BitwiseCopyable`.
+
+## Motivation
+
+Swift can compile generic code into an unspecialized form in which the compiled function receives a value and type information about that value.
+Basic operations are implemented by the compiler as calls to a table of "value witness functions."
+
+This approach is flexible, but can represent significant overhead.
+For example, using this approach to copy a buffer with a large number of `Int` values requires a function call for each value.
+
+Constraining the types in generic functions to `BitwiseCopyable` allows the compiler (and in some cases, the developer) to instead use highly efficient direct memory operations in such cases.
+
+The standard library already contains many examples of functions that could benefit from such a concept, and more are being proposed:
+
+The `UnsafeMutablePointer.initialize(to:count:)` function introduced in [SE-0370](0370-pointer-family-initialization-improvements.md) could use a bulk memory copy whenever it statically knew that its argument was `BitwiseCopyable`.
+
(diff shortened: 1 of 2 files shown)


proposals/nnnn-bitwise-copyable.md
```diff
@@ -0,0 +1,270 @@
...
+* Integer types
+* Floating point types
+* SIMD types
+* Pointer types
+* `Unmanaged`
+* `Optional`
+
+For an exhaustive list, see the [appendix](#all-stdlib-conformers).
+
+### Additional BitwiseCopyable types
```
allevato, line 83: During discussion, a couple other cases came up that we think would be worth calling out explicitly:

* Do raw-value `enum`s get `BitwiseCopyable` inferred if its raw value type is `BitwiseCopyable`? I think there are a couple possible answers here: either "yes", if it would be inferred for the same enum without raw values or associated values since it's a simple ordinal, or "no" if it wouldn't also be inferred for that case. (That is, having a raw value doesn't affect the Swift layout of the enum, so there should be no difference in `BitwiseCopyable` inference from the non-raw-value case.)

* Do imported C structs get `BitwiseCopyable`? If not, is there a way to add it (either retroactively, or via API Notes or `__swift_attr__` in the header)?

proposals/nnnn-bitwise-copyable.md
```diff
@@ -0,0 +1,270 @@
...
+
+## Acknowledgments
+
+This proposal has benefitted from discussions with John McCall, Joe Groff, Andrew Trick, Michael Gottesman, and Arnold Schwaigofer.
+
+## Appendix: Standard library conformers<a name="all-stdlib-conformers"/>
+
+The following protocols in the standard library will gain the `BitwiseCopyable` constraint:
+
+- `FixedWidthInteger`
```
allevato, line 228: The language steering group discussed these proposed conformances and we feel that `FixedWidthInteger` shouldn't be in this list. While the concrete types in stdlib that conform to it are certainly bitwise copyable, there are no semantic requirements of the protocol itself that _require_ bitwise copyability.

The summary of `FixedWidthInteger` from the documentation is "An integer type that uses a fixed size for every instance"; while it's unlikely, an implementation _could_ do something a bit tortured like store a pointer to heap storage that contains the actual value. It was felt that the term "fixed width" here refers to the range of values it can represent, rather than the internal representation.

Would you agree/disagree with that?

proposals/nnnn-bitwise-copyable.md
```diff
@@ -0,0 +1,270 @@
...
+- `Optional<T>` when `T` is `BitwiseCopyable`
+- The fixed-precision integer types:
+  - `Bool`
+  - `Int8`, `Int16`, `Int32`, `Int64`, `Int`
+  - `UInt8`, `UInt16`, `UInt32`, `UInt64`, `UInt`
+  - `StaticBigInt`
+  - `UInt8.Words`, `UInt16.Words`, `UInt32.Words`, `UInt64.Words`, `UInt.Words`
+  - `Int8.Words`, `Int16.Words`, `Int32.Words`, `Int64.Words`, `Int.Words`
+- The fixed-precision floating-point types:
+  - `Float`, `Double`, `Float80`
```
allevato, line 244: Can you add `Float16` here as well?

proposals/nnnn-bitwise-copyable.md
```diff
@@ -0,0 +1,270 @@
...
+  - `UnsafeBufferPointer`, `UnsafeMutableBufferPointer`
+  - `UnsafeRawBufferPointer`, `UnsafeMutableRawBufferPointer`
+  - `Unmanaged`
+  - `CVaListPointer`
+- Some types related to collections
+  - `EmptyCollection`
+  - `UnsafeBufferPointer.Iterator`, `UnsafeRawBufferPointer.Iterator`, `EmptyCollection.Iterator`
+  - `String.Index`, `CollectionDifference.Index`
+- Some types related to unicode
+  - `Unicode`
```
allevato, line 260: We considered some of the uninhabited types here and we feel that we need to be more selective about which get the conformance, even though they're trivially bitwise copyable by virtue of being uninhabited and thus zero-sized. Rather than looking at just the size of the type, should we consider intent—as in, would a user reasonably write a generic algorithm constrained by `BitwiseCopyable` that might involve one of these types?

We decided that the answer is probably **no** for `Unicode`, `MemoryLayout`, and `CommandLine`, as those are purely namespaces for other declarations. But on the flipside, the answer is probably **yes** for `Never` and `SystemRandomNumberGenerator`; `Never` does occur in generic contexts and we have precedence for adding other conformances to it for that purpose. Likewise, despite being zero-size, users do create values of `SystemRandomNumberGenerator` so the conformance makes sense there.

allevato (commented):
Hi all! I'll be review managing this proposal. I've left some comments based on our discussion in the language steering group.

proposals/nnnn-bitwise-copyable.md
tbkka: I would defer to @stephentyrone on this.  It doesn't seem to make a practical difference, since we're explicitly notating every existing fixed-width integer type.

nate-chandler:
Thanks for sharing the feedback @allevato.  There should be an updated draft to address these points in the next day or so.

proposals/nnnn-bitwise-copyable.md
nate-chandler: Updated the implementation and proposal accordingly.  And added a few words to describe the inference suppression mechanism.

proposals/nnnn-bitwise-copyable.md
nate-chandler: Added.

proposals/nnnn-bitwise-copyable.md
nate-chandler: Added a footnote for raw-value enums.  Inference is based on the enum itself, not the `RawValue` type witness of its synthesized `RawRepresentable` conformance.

Added a section on imported C and C++ types.  Conformance is inferred unless the fields are not representable in Swift (or in the C++ case if the type is non-trivial).  Conformance can be forced even when not inferred via `__attribute__((__swift_attr__("_BitwiseCopyable")))`.

proposals/nnnn-bitwise-copyable.md
```diff
@@ -0,0 +1,270 @@
...
+As a convenience, unconditional conformances will be inferred for structs and enums much of the time.
+When the module containing the type is built, if all of the type's fields are `BitwiseCopyable`, the compiler will generate a conformance for it to `BitwiseCopyable`.
+
+For generic types, a conformance will only be inferred if its fields unconditionally conform to `BitwiseCopyable`.
+In the `RegularBox` example above, a conditional conformance will not be inferred.
+If this is desired, the developer can explicitly write the conditional conformance.
+
+### Inference for types in evolving libraries
+
+This does not apply to public (or `@usableFromInline`) types defined within a module built with library evolution.
```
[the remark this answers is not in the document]
nate-chandler: Rewrote this section in terms of "exported" types.

proposals/nnnn-bitwise-copyable.md
nate-chandler: This remains unchanged for now, pending comment.

allevato: @stephentyrone Can you comment quickly on this, please?

stephentyrone: What Tony said. Every FixedWidthInteger type in the stdlib is bitwise copyable, but it is technically possible for someone to implement a fixed-width integer type that is not.

nate-chandler: Removed.

proposals/nnnn-bitwise-copyable.md
```diff
@@ -0,0 +1,356 @@
...
+## Acknowledgments
+
+This proposal has benefitted from discussions with John McCall, Joe Groff, Andrew Trick, Michael Gottesman, and Arnold Schwaigofer.
+
+## Appendix: Standard library conformers<a name="all-stdlib-conformers"/>
+
+The following protocols in the standard library will gain the `BitwiseCopyable` constraint:
+
+- `_Pointer`
+- `SIMDStorage`, `SIMDScalar`, `SIMD`
```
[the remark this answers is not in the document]
allevato: When we discussed those, there was a sense that the SIMD protocols were less likely to be used in a way that wasn't `BitwiseCopyable` than `FixedWidthInteger`, though it's kind of a fine line.

Let's keep things as they are currently based on the steering group's original feedback, and if there are other protocols/types that folks want to discuss further, we can use the proposal review thread for those.

allevato (approved):
Thank you for the updates! Getting ready to merge and start the review.

PR #2317, opened 2024-02-14
Partial Consumption of Noncopyable Values

nate-chandler:


proposals/nnnn-partial-consumption.md
```diff
@@ -0,0 +1,260 @@
...
+  }
+}
+```
+
+Partial consumption relies on viewing a struct as just a combination of values, though.
+This view is what permits the refinement of the check that the whole noncopyable aggregate is consumed on each path
+to checking that each noncopyable field of the aggregate is consumed exactly once.
+
+Because, as this example illustrates, the layout of a resilient struct may change,
+partial consumption isn't permitted for a noncopyable resilient value used outside the resilience domain which defines the type.
```
beccadax, line 148: I assume this restriction does not apply to `@frozen` types even if they come from another resilience domain?

tbkka: Yes.

proposals/nnnn-partial-consumption.md
```diff
@@ -0,0 +1,260 @@
...
+This remains true when a value is partially consumed:
+
+```swift
+extension Named {
+  consuming func unpack() {
+    takeString(name)
+    takeString(name)
+    takeUnique(unique)
+  }
+}
```
xwu, line 177: What about:

```
takeUnique(unique)
takeString(name)
takeString(name)
```

nate-chandler: Yes, that too is legal.

xwu:
Thanks @nate-chandler!

We discussed this proposal within the steering group, and I believe Joe and John shared some of their thoughts on the pitch thread.

There was also a bit of feedback earlier in the pitch regarding non-frozen `public` types being resilient and therefore not partially consumable in library evolution mode but not otherwise—the discussion within the steering group was much along the same lines as @allevato detailed for the `BitwiseCopyable` pitch, in that I think we'd prefer to avoid more behaviors being conditional on library evolution mode. (And also, if it could be spelled out how `package` works, that would be great too.)

nate-chandler:
@xwu The proposal has been updated with the following changes:
- Legality of partial consumption is based on whether types are imported from a different module rather than on a compiler flag (-enable-library-evolution).  Partial consumption of `@frozen` types remains legal.
- Partial consumption within deinit is added.
- Explicit consumption of noncopyable fields is added.
- A future direction for explicit consumption of copyable fields is added.
- A future direction for partial consumption of copyable values is added.

proposals/nnnn-partial-consumption.md
```diff
@@ -0,0 +1,334 @@
...
+
+It is currently not straightforward to write a function that forms a new `Pair` with the values reversed.
+For example, the following code is not currently allowed:
+
+```swift
+extension Pair {
+  consuming func swap() -> Pair {
+    return Pair(
+      first: first, // error: cannot partially consume 'self'
+      second: second // error: cannot partially consume 'self'
```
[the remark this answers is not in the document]
nate-chandler: Thanks!

PR #2324, opened 2024-02-16
Change reference from Swift 5.11 to Swift Next

dempseyatgithub:
Change reference from 'Swift 5.11' to 'Swift Next' to  prevent community confusion viewing the SE dashboard until all proposals are updated from Swift Next.


DougGregor (approved):
Thanks!

PR #2335, opened 2024-02-21
Normalize implementation versions for past major versions to match Swift 6.0

dempseyatgithub:
This PR normalizes the implementation versions for past major versions to match the LSG decision to use '6.0' for the upcoming release, to differentiate from Swift language modes.

I went back to original release announcement blog posts which list proposals implemented to verify that these proposals were part of the appropriate '.0' release.

Although a large number of files change, the changes are all one of the following:
`(Swift 3)` -> `(Swift 3.0)`
`(Swift 4)` -> `(Swift 4.0)`
`(Swift 5)` -> `(Swift 5.0)`

This change will make the version filter interface on the Swift Evolution dashboard consistent, instead of mixing major release versions with and without a '.0'.

I have put up a PR for swift.org to update the list of filters. https://github.com/apple/swift-org-website/pull/557

From @rjmccall in https://github.com/apple/swift-evolution/pull/2328:
> The LSG decided that we'd like to use 6.0 going forward to avoid confusion with Swift language modes. Updating the existing proposals is the right thing to do. Unfortunately, that'll require checking to make sure that the .0 is actually right; if you'd like to take that on, you're welcome to.
> 
> I'm not sure how to update the filter list on the dashboard; we don't usually maintain that.

### DIFF
--- a/proposals/0002-remove-currying.md
+++ b/proposals/0002-remove-currying.md
@@ -2,7 +2,7 @@
 
 * Proposal: [SE-0002](0002-remove-currying.md)
 * Author: [Joe Groff](https://github.com/jckarter)
-* Status: **Implemented (Swift 3)**
+* Status: **Implemented (Swift 3.0)**
 * Implementation: [apple/swift@983a674](https://github.com/apple/swift/commit/983a674e0ca35a85532d70a3eb61e71a6d024108)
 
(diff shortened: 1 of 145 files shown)


rjmccall:
Thanks!  I really appreciate you going through all of these.

PR #2339, opened 2024-02-23
Add draft proposal for transferring parameters and results.

hborla:


proposals/NNNN-transferring-parameters-and-results.md
```diff
@@ -0,0 +1,320 @@
...
+
+The "transferred parameter" behavior of actor initializers is a generally
+useful concept, but it is not possible to explicitly specify that functions
+and methods can transfer away specific parameter values. Consider the following
+code that uses `CheckedContinuation`:
+
+```swift
+@MainActor var mainActorState: NonSendable?
+
+nonisolated func test() async {
```
gottesmm, line 97: nonisolated isn't needed here.

proposals/NNNN-transferring-parameters-and-results.md
```diff
@@ -0,0 +1,320 @@
...
+  let ns = await withCheckedContinuation { continuation in
+    Task { @MainActor in
+      let ns = NonSendable()
+      // Oh no! 'NonSendable' is passed from the main actor to a
+      // nonisolated context here!
+	  continuation.resume(returning: ns)
```
gottesmm, line 103: Indentation fix needed. I do like the Oh no though. = ). Made me giggle a little.

proposals/NNNN-transferring-parameters-and-results.md
```diff
@@ -0,0 +1,320 @@
...
+* Implementation: On `main` gated behind `-enable-experimental-feature TransferringArgsAndResults`
+* Previous Proposal: [SE-0414: Region-based isolation](/proposals/0414-region-based-isolation.md)
+* Review: ([pitch](https://forums.swift.org/...))
+
+
+## Introduction
+
+This proposal extends region isolation to enable an explicit `transferring`
+annotation to denote when a parameter or result value is required to be in a
+disconnected region, allowing the callee or the caller, respectively, to
```
gottesmm, line 16: I am a little worried that this is going to sound to people like these are disconnected parameters (which is an extension as we have spoken about). I think it would be clearer if you included verbiage that says that the values need to be disconnected upon function start in the callee and function return in the caller. Then it is really clear that we are allowing for the transferring values to be merged into other regions. That being said, I am willing to be convinced that this is too much detail here.

proposals/NNNN-transferring-parameters-and-results.md
```diff
@@ -0,0 +1,320 @@
...
+@MainActor func main(ns: NonSendable) {}
+
+func tryTransfer(ns: NonSendable) async {
+  // error: task isolated value of type 'NonSendable' transferred to
+  // main actor-isolated context
+  await main(ns: ns)
+}
+```
+
+However, for actor initializers, parameters are always considered to be
```
gottesmm, line 40: This is true for all actor methods. I would just tweak it a little:

```
However for actor methods, parameters are always considered to be transferred into the actor's region since the actor method's body could modify the actor's state using the parameter.
```

Just misc wordsmith.

proposals/NNNN-transferring-parameters-and-results.md
```diff
@@ -0,0 +1,320 @@
...
+  // init. Later uses in caller could race with uses on the actor
+  let myActor = MyActor(ns: ns)
+
+  print(ns) // note: note: access here could race
+}
+```
+
+In the above code, if the `ns` local variable in the `transfer` function were
+instead a function parameter, it would be invalid to transfer `ns` into
+`myActor`'s region because the caller of `transfer()` may use the argument
```
gottesmm, line 72: Great!

proposals/NNNN-transferring-parameters-and-results.md
```diff
@@ -0,0 +1,320 @@
...
+	  continuation.resume(returning: ns)
+
+      // Save 'ns' to main actor state for concurrent access later on
+      mainActorState = ns
+	  }
+	}
+
+    // 'ns' and 'mainActorState' are now the same non-Sendable value;
+    // concurrent access is possible!
+	ns.mutate()
```
gottesmm, line 112: Great example!

proposals/NNNN-transferring-parameters-and-results.md
```diff
@@ -0,0 +1,320 @@
...
+}
+```
+
+In the above code, the closure argument to `withCheckedContinuation` crosses
```
gottesmm, line 116: Might be helpful to turn this into a list. Not sure though, just food for thought.

proposals/NNNN-transferring-parameters-and-results.md
```diff
@@ -0,0 +1,320 @@
...
+an isolation boundary to get onto the main actor, creates a non-`Sendable`
+value, then resumes the continuation with that non-`Sendable` value. The
+non-`Sendable` value is then returned to the original `nonisolated` context,
+thus crossing an isolation boundary. Because `resume(returning:)` does not
+impose a `Sendable` requirement on its argument, this code does not produce any
+data-race safety diagnostics, even under `-strict-concurrency=complete`.
+
+Requiring `Sendable` on the parameter type of `resume(returning:)` is a harsh
+restriction, and it's safe to pass a non-`Sendable` value as long as the value
+is in a disconnected region.
```
gottesmm, line 126: I would add to the end here:

as long as we do not use the disconnected value later.

proposals/NNNN-transferring-parameters-and-results.md
```diff
@@ -0,0 +1,320 @@
...
+
+## Proposed solution
+
+This proposal enables explicitly specifying parameter and result values as
+being transferred over an isolation boundary using a modifier:
```
gottesmm, line 131: I think you should mention transferring here.

proposals/NNNN-transferring-parameters-and-results.md
```diff
@@ -0,0 +1,320 @@
...
+  await acceptTransfer(ns)
+
+  // note: access here could race
+  print(ns)
+}
+```
+
+What the callee does with the argument value is opaque to the caller; the
+callee may transfer the value away, or it may merge the value to the isolation
+region of one of the other parameters.
```
gottesmm, line 168: Nice!

proposals/NNNN-transferring-parameters-and-results.md
```diff
@@ -0,0 +1,320 @@
...
+  }
+}
+```
+
+The caller of a function returning a `transferring` result can assume the value
+is in a disconnected region, enabling non-`Sendable` result values to cross an
+actor isolation boundary:
+
+```swift
+nonisolated func f(s: S) async {
```
gottesmm, line 195: nonisolated isn't needed here.

proposals/NNNN-transferring-parameters-and-results.md
```diff
@@ -0,0 +1,320 @@
...
+  let ns = s.getNonSendable() // okay; 'ns' is in a disconnected region
```
gottesmm, line 196: I would also show that you can then transfer it further here.

proposals/NNNN-transferring-parameters-and-results.md
```diff
@@ -0,0 +1,320 @@
...
+    return ns
+  }
+}
+```
+
+### Ownership convention for `transferring` parameters
+
+When a call passes an argument to a `transferring` parameter, the caller cannot
+use the argument value again after the callee returns. By default `transferring`
+on a function parameter implies that the callee consumes the parameter, but it does
```
gottesmm, line 287: Can you make it clearer here that one can assign into a transferring parameter (that is... it is var like). I have a feeling that readers will think that consumes means +1. Maybe something like:

By default `transferring` on a function parameter implies that the callee consumes the parameter. Inside the callee, the function parameter can be re-initialized like a consuming parameter albeit without implying no implicit copy semantics.

proposals/NNNN-transferring-parameters-and-results.md
```diff
@@ -0,0 +1,320 @@
...
+not imply no implicit copying semantics. `transferring` may also be used with an
+explicit `consuming` or `borrowing` ownership modifier.
```
gottesmm, line 289: I would just make this a little clearer by explicitly saying that these two let one opt into no implicit copy semantics. Maybe something like:

`transferring` may be paired with an explicit `consuming` to opt into no implicit copy semantics.

Borrowing may need a little more exposition since one can pass it as a weakly transferred thing, but one cannot transfer it by passing it to another bare transferring parameter due to the no implicit copy semantics.

proposals/NNNN-transferring-parameters-and-results.md
```diff
@@ -0,0 +1,320 @@
...
+based on language mode allows `transferring` to be adopted in existing
+Concurrency APIs including `CheckedContinuation`.
+
+## ABI compatibility
+
+This proposal does not change how any existing code is compiled.
+
+## Implications on adoption
+
+Adding `transferring` to a parameter is more restrictive at the caller, and
```
FranzBusch, line 306: Can we make sure the API breaking checker is diagnosing this correctly?

proposals/NNNN-transferring-parameters-and-results.md
hborla: I know, I only added it to make it obvious that there's an isolation boundary crossed

proposals/NNNN-transferring-parameters-and-results.md
hborla: Updated to say the "disconnected region" requirement only applies at the function boundary

proposals/NNNN-transferring-parameters-and-results.md
```diff
@@ -0,0 +1,357 @@
+# `transferring` isolation regions of parameter and result values
+
+* Proposal: [SE-NNNN](NNNN-transferring-parameters-and-results.md)
+* Authors: [Michael Gottesman](https://github.com/gottesmm), [Holly Borla](https://github.com/hborla), [John McCall](https://github.com/rjmccall)
+* Review Manager: TBD
+* Status: **Awaiting review**
+* Implementation: On `main` gated behind `-enable-experimental-feature TransferringArgsAndResults`
+* Previous Proposal: [SE-0414: Region-based isolation](/proposals/0414-region-based-isolation.md)
+* Review: ([pitch](https://forums.swift.org/...))
```
beccadax, line 9: Nees a valid pitch thread link.

PR #2343, opened 2024-02-25
[SE-0419] Fix link to pitch thread

ole:
The link to the pitch thread pointed to a post in the middle of the thread. It should point to the start.


DougGregor (approved):
Thanks!

PR #2344, opened 2024-02-26
Allow trailing comma in comma-separated lists

mateusrodriguesxyz:
https://forums.swift.org/t/pitch-allow-trailing-comma-in-tuples-arguments-and-if-guard-while-conditions/70170


jessbennett:
Where are your GitHub checks?

proposals/proposals/nnnn-trailing-comma-tuples-arguments-conditions.md
```diff
@@ -0,0 +1,200 @@
...
+With trailing comma:
+```diff
+foo(
+  a: Int,
++  b: Int,
+)
+```
+
+> [!NOTE]  
+> A similar proposal was [rejected](https://forums.swift.org/t/rejected-se-0084-allow-trailing-commas-in-parameter-lists-and-tuples/2777) back in 2016 for Swift 3. It's been 8 years since that, the swift language has evolved a lot, some changes highlighted above as motivation, and the code style that "puts the terminating right parenthesis on a line following the arguments to that call" has been widely adopted by community, swift standard library codebase, swift-format, docc documentation and Xcode. Therefore, not encourage or endorse this code style doesn't hold true anymore nor is a reason for rejection.
```
xwu, line 104: SE-0084 explicitly proposed _not_ allowing trailing comma in single-element tuples.

So that this proposal text meets the standard of describing the solution with sufficient detail that someone else can implement it just by reading the text, I think this proposal should spell out explicitly the grammar changes proposed, which includes calling out whether trailing comma is allowed in single-element tuples, as well as zero-element tuples and parameter lists—e.g., is `(,)` equivalent to `()` (aka `Void`)?

proposals/proposals/nnnn-trailing-comma-tuples-arguments-conditions.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Source compatibility
+
+This change is purely additive and has no impact on existing code.
+
+## Future directions
+
+### Allow trailing comma anywhere there's a comma-separated list
+
+Although this proposal focuses on the most requested use cases for trailing comma, there's other places with comma-separated list and the restriction could be consistently lifted for all of these.
```
xwu, line 157: Missing from this list are variable bindings, e.g., `var a, b, c`.

Also worth calling out explicitly (in the proposed solution) is whether this proposal will support `var (a, b,) = (1, 2,)`, or whether only the right-hand side is allowed to have a trailing comma—in other words, whether a tuple pattern counts as a "tuple" under the proposed solution.

proposals/proposals/nnnn-trailing-comma-tuples-arguments-conditions.md
```diff
@@ -0,0 +1,200 @@
...
+#### Switch Case
+
+```swift
+switch number {
+    case 1, 2, 3,:
+        ...
+    default:
+        ...
+}
+```
```
xwu, line 179: One of the justifications offered (which you should feel free to include in the text here) for expanding support for trailing commas is growing developer expectation, with many popular languages offering support. As folks have replied on the pitch thread, many languages either choose to stop at arrays/lists or support trailing commas basically everywhere that commas are used as delimiters.

Is there a principle or justification that underlies this proposal expanding support to include tuples and arguments, and control flow conditions _other than_ `switch`, but not these future directions? If so, can you articulate it here?

proposals/proposals/nnnn-trailing-comma-tuples-arguments-conditions.md
```diff
@@ -0,0 +1,200 @@
...
+   condition2,
+   condition3,
+{
+                    
+}
+```
+
+## Source compatibility
+
+This change is purely additive and has no impact on existing code.
```
xwu, line 151: This, or the above section, might be a good place to include your findings regarding the implications for parsing condition lists that you've mentioned in the pitch thread. Even if how they're parsed is formally never ambiguous to the compiler, if there are scenarios in which a reader could be confused/misled due to how trailing commas are parsed in condition lists, it'd be good to record for posterity what the intended behavior is.

proposals/proposals/nnnn-trailing-comma-tuples-arguments-conditions.md
mateusrodriguesxyz: I think I'm pro adding everywhere that is possible but I chose the ones that I personally considered more demanded for scope reason. I did considered add more during the implementation review but @ahoppen rightly [pointed](https://github.com/apple/swift-syntax/pull/2515#issuecomment-2006951192) that the review had already dragged too long so we went with just the proposed here. And to be honest, this has been more time-consuming than I expected so if possible I would prefer to not expand the proposal. If this is approved and there is a desire to expand, I would be willing to make another proposal expanding to more places.

mateusrodriguesxyz:
@xwu I have updated proposed solution and source compatibility sections to address your feedbacks.

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+
+```swift
+let rank = [
+  "Player 1",
+  "Player 3",
+  "Player 2",
+]
+```
+
+Using trailing commas makes it easy to add, remove, reorder and comment in/out elements, without the need to add or delete the comma while doing any of these manipulations.
```
xwu, line 28:
```suggestion
Swift's support for trailing commas in array and dictionary literals makes it as easy to append, remove, reorder, or comment out the last element as any other element.
```

To be precise, the benefits have to do with the _last_ element of a list, not other elements. So perhaps something like this? Also, it is a bit pedantic, but one can try to _use_ a trailing comma but if it's not supported then it's not easy...

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+```
+
+`frame(width:)`, `frame(height:)`, `frame(width:alignment:)`, `frame(height:alignment:)`, `frame(width:height:)`, `frame(width:height:alignment:)` are all valid calls but you can't easily switch between `frame(width:)` and `frame(width:alignment:)` by commenting in/out `alignment` without adding/removing the trailing comma.
+
+```swift
+.frame(
+    width: 500,
+//    alignment: .leading
+) // ❌ Unexpected ',' separator
+```
```
xwu, line 47: Suggestion—I'd start the shift from your first example of something already supported (array literals) to the present motivation with a sentence to introduce this shift, like: "Other comma-separated lists in the language could also benefit from the flexibility enabled by trailing commas." You have a sentence at the end of this section that probably would work here.

It's permitted certainly to mention SwiftUI as an example, but if there's an example that applies more broadly than Apple platforms and app development, it's arguably preferable. For example, you could choose an API from the standard library, swift-algorithms, etc.

If you choose a well known enough API, you also wouldn't have to provide the declaration and list all the overloads: just the call site and the error that results would be sufficient.

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+
+The introduction of [parameter packs](https://github.com/apple/swift-evolution/blob/main/proposals/0393-parameter-packs.md) allows more APIs that are comma-separated lists at call site and would benefit from trailing comma.
```
xwu, line 49: Before this sentence, I would move the acknowledgment you have written below about SE-0084. You can write that a narrower proposal was reviewed and rejected, but since that time the language has evolved substantially that change the basis for rejection.

That would include not just evolution in prevailing code style, but also the point here that we have introduced parameter packs, but also the following point that we have introduced code generation tools.

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+extension [S] {
+    func sorted<each T: Comparable>(_ keyPath: repeat KeyPath<S, each T>) { }
+}
+
+arrayOfS.sorted(
+  \.a, 
+  \.b,
+//  \.c
+) // ❌ Unexpected ',' separator
+```
```
xwu, line 61: It isn't necessary to demonstrate how parameter packs work in this proposal, I think.

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+
+Since [#21381](https://github.com/apple/swift/pull/21381) has been merged back in 2019 **enum associated values** supports default values and are a good fit for trailing comma as well.
```
xwu, line 63: I don't understand this point. Why do default values for enum associated values need to be called out here?

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+
+```swift
+if 
+   condition1,
+   condition2,
+//   condition3
+{ // ❌ Cannot convert value of type '() -> ()' to expected condition type 'Bool'
+                    
+} // ❌ Expected '{' after 'if' condition
+```
```
xwu, line 75: I don't think this needs to be called out separately. It starts to read like an incomplete version of the proposed solution.

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+
+These benefits can be extended to other comma-separated lists found in the Swift language.
+
+### Code Generation
+
+**Plugins** and **Macros** have made it possible to generate code using swift and trailing comma would allow generate lists without worrying about a special condition for the last element.
```
xwu, line 81: See comment above.

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+
+### Code Diff
+
+A tangential motivation is that trailing comma makes version-control diffs cleaner.
```
xwu, line 85: If it is tangential, maybe better to leave it out? If we make editing less messy, then clearly the diff will also be cleaner: they are two sides of the same coin and the point is made adequately above.

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+With trailing comma:
+```diff
+foo(
+  a: Int,
++  b: Int,
+)
+```
+
+> [!NOTE]  
+> A similar proposal was [rejected](https://forums.swift.org/t/rejected-se-0084-allow-trailing-commas-in-parameter-lists-and-tuples/2777) back in 2016 for Swift 3. It's been 8 years since that, the swift language has evolved a lot, some changes highlighted above as motivation, and the code style that "puts the terminating right parenthesis on a line following the arguments to that call" has been widely adopted by community, swift standard library codebase, swift-format, docc documentation and Xcode. Therefore, not encourage or endorse this code style doesn't hold true anymore nor is a reason for rejection.
```
xwu, line 104: See comment earlier about moving this reference up. Also, please link directly to the proposal and not just to the decision note.

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+
+## Proposed solution
+
+This proposal adds support for trailing comma for comma-separated lists, but only when there's a clear terminator. A enum case label list won't support, for example:
```
xwu, line 108:
```suggestion
This proposal adds support for trailing commas in comma-separated lists when there's a clear terminator, which are the following:
```

Please list those here; probably table format is fine. I don't think this section should start first with an example of what's _not_ supported although that should be included in the proposal, including the ambiguity that arises if it is supported.

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+
+```swift
+enum E {
+  case a, b, c, // ❌ Expected identifier after comma in enum 'case' declaration
+}
+```
+
+Trailing comma will be allowed in the following comma-separated lists:
```
xwu, line 116:
```suggestion
```

See comment above.

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+- Tuples and tuple patterns 
+
+```swift
+(1, 2,)
+let block: (Int, Int,) -> Void = { (a, b,) in  }
+let (a, b,) = (1, 2,)
+for (a, b,) in zip(s1, s2) { }
+```
+
+**Trailing comma will be allowed in single-element lists but not in zero-element lists.**
```
xwu, line 127: Proposals are expected to have a "Detailed design" section, and I think these details—which are extremely useful and needed!—about single-element tuples versus zero-element tuples go best there. The details should come with a rationale; in this case, supporting a zero-element list would require supporting _leading_ commas, which isn't what this proposal is about (nor probably could we do it).

I would suggest that details about places in the grammar that don't have a clear terminator and the parsing ambiguities we need to avoid by continuing not to support trailing commas there could go into the "Detailed design" section as well.

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+
+```swift
+(1,) // OK
+(,) // ❌ expected value in tuple
+```
+
+- Initializers, functions and associated values arguments and parameters
```
xwu, line 134: Sorry, there are two "ands" here—did you mean to comma-separate some of these? :wink:

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+
+```swift
+
+m[x, y,]
```
xwu, line 138: This looks like a subscript example, but subscripts are mentioned below, no here...

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+    case foo(a: Int, b: Int,)
+}
+```
+- Subscripts
+
+```swift
+let value = m[x, y,]
+```
+
+- `KeyPath` subscripts
```
xwu, line 156: They may be a totally distinct case within swift-syntax, but I think most Swift end users would expect subscripts (the bullet point above) to include key path subscripts, no?

To such a reader, this separate bullet point would seem redundant—perhaps this could be consolidated into one ("Subscripts, including key path subscripts" or something like that)?

proposals/nnnn-trailing-comma-lists.md
xwu: Also, it looks like you're showing only declarations, rather than call sites. Perhaps worth showing both kinds of examples.

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+
+- `KeyPath` subscripts 
+
+```swift
+let keyPath = \Foo.bar[x,y,]
+          
+f(\.[x,y,])
+```
+
+- Attributes arguments
```
xwu, line 164: Similar comment for this and the following. I understand that you've had to handle separate cases in the implementation, but to end users they're all argument lists, I think.

It would seem logical that use sites for expression macros including built-in ones such as `#colorLiteral(...)`, attached macros, property wrappers, and built-in attributes all would fit under one heading of "argument lists"—although of course it is important to spell it out.

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+
+```swift
+#foo(1, 2,)
+
+struct S {
+    #foo(1, 2,)
+}
+```
+
+- `if`, `guard` and `while` conditions
```
xwu, line 181:
```suggestion
- `if`, `guard` and `while` condition lists
```

I *think* that's what they're called, yes?

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+```swift
+switch number {
+    case 1, 2,:
+        ...
+    default:
+        ..
+}
+```
+
+- Closure capture list
```
xwu, line 200:
```suggestion
- Closure capture lists
```

You've used plural elsewhere, so this would just be for consistency.

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+
+```swift
+{ [a, b,] in }
+```
+
+- inheritance list
```
xwu, line 206:
```suggestion
- Inheritance clauses
```

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+
+```swift
+struct S: P1, P2, P3, { }
+```
+
+**Inheritance lists won't support trailing comma when there isn't a clear terminator, such as `associatedtype` in a protocol declaration:**
```
xwu, line 212:
```suggestion
**Inheritance lists won't support trailing commas when there isn't a clear terminator, such as in `associatedtype` declarations in a protocol:**
```

"Such as" worries me here: do you have a concrete list or description of all the places they're not supported?

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+}
+```
+
+- Generic parameters
+
+```swift
+struct S<T1, T2, T3,> { }
+```
+
+- Generic `where` clause list
```
xwu, line 225:
```suggestion
- Generic `where` clauses
```

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+
+```swift
+struct S<T1, T2, T3> where T1: P1, T2: P2, { }
+```
+
+**Generic `where` clause lists won't support trailing comma when there isn't a clear terminator, such as functions in protocol declaration:**
```
xwu, line 231:
```suggestion
**Generic `where` clause lists won't support trailing commas when there isn't a clear terminator, such as functions in protocol declarations:**
```

Same comment as above re "such as."

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+```swift
+protocol Foo {
+  func f<T1, T2>(a: T1, b: T2) where T1: P1, T2: P2, // ❌ Expected type
+}
+```
+
+- Availability spec list
```
xwu, line 238: As above re attributes and macros, `#available` and `#unavailable` argument lists would seem to fall into the same category as other argument lists for the end user.

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+
+```swift
+if #unavailable(iOS 15, watchOS 9,) { }
```
xwu, line 241: Point of clarification: since `*` is (required to be?) at the end of these lists if used, can you give an example (maybe best in the "Detailed design" section) regarding whether `#available(iOS 9999, *,)` will be supported—in other words, a trailing comma after `*`?

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+
+- String interpolation
+
+```swift
+let s = "\(1, 2,)"
+```
+
+## Source compatibility
+
+Although this change won't impact existing valid code it will change how some invalid codes are parsed. Consider the following:
```
xwu, line 252:
```suggestion
Although this change won't impact existing valid code it will change how some invalid code is parsed. Consider the following:
```

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+  condition1,
+  condition2,
+{ // ❌ Function produces expected type 'Bool'; did you mean to call it with '()'?
+  return true
+} 
+
+{ print("something") }
+```
+
+Currently the parser uses the last comma to determine that whatever follows is the last condition, so `{ return true }` is a condition and `{ print("something") }` is the if body.
```
xwu, line 265:
```suggestion
Currently the parser uses the last comma to determine that whatever follows is the last condition, so `{ return true }` is a condition and `{ print("something") }` is the `if` body.
```

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+To allow trailing comma the proposed solution is to change the parser to terminate de condition list before the first block that is a valid if body, so `{ return true }` will the parsed as the if body.
```
xwu, line 266:
```suggestion
With trailing comma support, the parser will terminate the condition list before the first block that is a valid `if` body, so `{ return true }` will the parsed as the `if` body and `{ print("something") }` will be an unused closure expression.
```

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+A different approach to address the exact same motivation is to allow the comma between two expressions to be elided when they are separated by a newline.
+
+```swift
+print(
+    "red"
+    "green"
+    "blue"
+)
+```
+This was even [proposed](https://forums.swift.org/t/se-0257-eliding-commas-from-multiline-expression-lists/22889/188) and returned to revision back in 2019.
```
xwu, line 292:
```suggestion
This was even [proposed](https://forums.swift.org/t/se-0257-eliding-commas-from-multiline-expression-lists/22889/188) and returned for revision back in 2019.
```

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+
+Even though both approach are not mutually exclusive, this proposal is about consistently extend an existing behavior in the language while eliding comma is a more serious change to the language.
```
xwu, line 294:
```suggestion
The two approaches are not mutually exclusive. There remain unresolved questions about how the language can accommodate elided commas, and adopting this proposal does not prevent that approach from being considered in the future.
```

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,294 @@
...
+} 
+
+{ print("something") } // ❌ Closure expression is unused
+```
+
+## Alternatives considered
+
+### Eliding commas
+
+A different approach to address the exact same motivation is to allow the comma between two expressions to be elided when they are separated by a newline.
```
xwu, line 283:
```suggestion
A different approach to address similar motivations is to allow the comma between two expressions to be elided when they are separated by a newline.
```

They don't have to be the "exact same" motivation, or at least it's not necessary for you to argue that they're the exact same motivations here.

xwu (commented):
Editorial suggestions. Feel free, of course, to adopt or reject any as the proposal author, but my hope is that the suggested changes help readers to provide useful feedback during review and better understand what's proposed and what's not.

mateusrodriguesxyz:
@xwu I've just commited a updated version of the proposal text addressing your suggestions.

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,232 @@
...
+let subsequences = numbers.split(
+	separator: 0,
+//        maxSplits: 1
+) ❌ Unexpected ',' separator
+```
+
+### The Language Evolved
+
+Back in 2016, a similar [proposal](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0084-trailing-commas.md) with a narrower scope was reviewed and rejected for Swift 3 but since that time the language has evolved substantially that changed the basis for rejection. The code style that "puts the terminating right parenthesis on a line following the arguments to that call" has been widely adopted by community, swift standard library codebase, swift-format, docc documentation and Xcode. Therefore, not encouraging or endorsing this code style doesn't hold true anymore.
+The language  has also seen the introduction of [parameter packs](https://github.com/apple/swift-evolution/blob/main/proposals/0393-parameter-packs.md), which allows more APIs that are comma-separated lists at call site and would benefit from trailing comma, and code generation tools like plugins and macros that with trailing comma support wouldn't have to worry about a special condition for the last element when generating comma-separated lists.
```
xwu, line 46:
```suggestion
The language has also seen the introduction of [parameter packs](https://github.com/apple/swift-evolution/blob/main/proposals/0393-parameter-packs.md), which enables APIs that are generic over variable numbers of type parameters, and code generation tools like plugins and macros that, with trailing comma support, wouldn't have to worry about a special condition for the last element when generating comma-separated lists.
```

Any function that takes more than one argument would have a comma-separated argument list; the salient point is that now these APIs can have _variable_ numbers of type parameters.

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,232 @@
...
+- Tuples and tuple patterns 
+
+```swift
+(1, 2,)
+let block: (Int, Int,) -> Void = { (a, b,) in  }
+let (a, b,) = (1, 2,)
+for (a, b,) in zip(s1, s2) { }
+```
+
+- Declaration and call site for arguments lists of initializers, functions, associated values, built-in attributes, expression macros, attached macros and availability spec.
```
xwu, line 61:
```suggestion
- Parameter and argument lists of initializers, functions, enum associated values, expression macros, attributes, and availability specs.
```

Calling out "built-in attributes" and "attached macros" would suggest that property wrappers (for example) aren't supported...and I assume they are, yes? If so, I suggest generalizing to "attributes."

Clarification: are built-in macros like `#colorLiteral` supported?

Another clarification (possibly most appropriate to mention in detailed design): if a built-in attribute or macro always supports only one argument, is a trailing comma supported? For example, is `#selector(MyViewController.tappedButton,)` valid?

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,232 @@
...
+The language  has also seen the introduction of [parameter packs](https://github.com/apple/swift-evolution/blob/main/proposals/0393-parameter-packs.md), which allows more APIs that are comma-separated lists at call site and would benefit from trailing comma, and code generation tools like plugins and macros that with trailing comma support wouldn't have to worry about a special condition for the last element when generating comma-separated lists.
+
+## Proposed solution
+
+This proposal adds support for trailing commas in comma-separated lists when there's a clear terminator, which are the following:
+
+- Tuples and tuple patterns
```
xwu, line 52:
```suggestion
- Tuples and tuple patterns.
```

(For consistency)

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,232 @@
...
+
+- String interpolation
+
+```swift
+let s = "\(1, 2,)"
+```
+
+## Detailed Design
+
+Trailing comma will be supported in comma-separated lists whenever there is a terminator clear enough that the parser can determine the end of the list. The terminator can be the symbols like `)`, `]`, `>`, `{` and `:`, a keyword like `where` or a pattern code like the body of a `if`, `guard` and `while` statement.
```
xwu, line 152:
```suggestion
Trailing commas will be supported in comma-separated lists whenever there is a terminator clear enough that the parser can determine the end of the list. The terminator can be the symbols like `)`, `]`, `>`, `{` and `:`, a keyword like `where` or a pattern code like the body of a `if`, `guard` and `while` statement.
```

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,232 @@
...
+
+Enum case label lists:
+
+```swift
+enum E {
+  case a, b, c, // ❌ Expected identifier after comma in enum 'case' declaration
+}
+```
+
+Inheritance clauses for associated type in a protocol declaration:
```
xwu, line 164:
```suggestion
Inheritance clauses for associated types in a protocol declaration:
```

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,232 @@
...
+
+```swift
+protocol Foo {
+  associatedtype T: P1, P2, // ❌ Expected type
+}
+```
+
+Generic `where` clauses for initializers and functions in protocol declaration:
```
xwu, line 172:
```suggestion
Generic `where` clauses for initializers and functions in a protocol declaration:
```

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,232 @@
...
+
+```swift
+protocol Foo {
+  func f<T1, T2>(a: T1, b: T2) where T1: P1, T2: P2, // ❌ Expected type
+}
+```
+
+Trailing comma will be allowed in single-element lists but not in zero-element lists, since the trailing comma is actually attached to the last element. Supporting a zero-element list would require supporting _leading_ commas, which isn't what this proposal is about.
```
xwu, line 180:
```suggestion
Trailing commas will be allowed in single-element lists but not in zero-element lists, since the trailing comma is actually attached to the last element. Supporting a zero-element list would require supporting _leading_ commas, which isn't what this proposal is about.
```

proposals/nnnn-trailing-comma-lists.md
```diff
@@ -0,0 +1,232 @@
...
+  condition1,
+  condition2,
+{ // ❌ Function produces expected type 'Bool'; did you mean to call it with '()'?
+  return true
+} 
+
+{ print("something") }
+```
+
+Currently the parser uses the last comma to determine that whatever follows is the last condition, so `{ return true }` is a condition and `{ print("something") }` is the `if` body.
```
xwu, line 203:
```suggestion
Currently the parser uses the last comma to determine that whatever follows is the last condition, so `{ return true }` is a condition and `{ print("something") }` is the `if` body.

```

xwu (commented):
Fantastic. Minor grammar edits and one point of clarification. I think otherwise we are ready to go.

proposals/nnnn-trailing-comma-lists.md
mateusrodriguesxyz: I hadn't thought about these cases but I tried and both `#colorLiteral` and `#selector` parsers with trailing comma. I think the c++ parser is really general about how it parses `#` expressions.

xwu: Sure, the important thing from a proposal perspective is that the text specifies the behavior, but we can clarify during or after review.

PR #2357, opened 2024-03-09
`@isolated(any)` function types

rjmccall:


proposals/NNNN-isolated-any-functions.md
```diff
@@ -0,0 +1,516 @@
...
+
+Function declarations and closures in Swift support three different
+forms of actor isolation:
+
+- They can be non-isolated, which means somewhat different things
+  depending on whether the function is synchronous or asynchronous.
+  This can be explicitly expressed with the `nonisolated` modifier,
+  but it is also the default if other rules don't apply.
+
+- They can be isolated to a specific [global actor][SE-0316] type.
```
ktoso, line 37: nit: doesn't render as link
```suggestion
- They can be isolated to a specific global actor ([SE-0316](https://github.com/apple/swift-evolution/blob/main/proposals/0316-global-actors.md)) type.
```

proposals/NNNN-isolated-any-functions.md
```diff
@@ -0,0 +1,516 @@
...
+calls within that context no longer cross isolation boundaries.
+Maybe this could also decompose the function value?
+
+### Adoption in task-creation routines
+
+There are a large number of functions in the standard library that create
+tasks:
+- `Task.init`
+- `Task.detached`
+- `TaskGroup.add`
```
ktoso, line 340: Similar to the other proposal I wrote I guess; if we want to list them all, we have to list them all: 

```suggestion
- `TaskGroup.addTask(UnlessCancelled)`
```

(or, just say "all")

proposals/NNNN-isolated-any-functions.md
```diff
@@ -0,0 +1,516 @@
...
+- `ThrowingTaskGroup.add`
```
ktoso, line 341:
```suggestion
- `ThrowingTaskGroup.addTask(UnlessCancelled)`
```

proposals/NNNN-isolated-any-functions.md
```diff
@@ -0,0 +1,516 @@
...
+- `DiscardingTaskGroup.add`
```
ktoso, line 342:
```suggestion
- `DiscardingTaskGroup.addTask(UnlessCancelled)`
```

proposals/NNNN-isolated-any-functions.md
```diff
@@ -0,0 +1,516 @@
...
+- `ThrowingDiscardingTaskGroup.add`
```
ktoso, line 343:
```suggestion
- `ThrowingDiscardingTaskGroup.addTask(UnlessCancelled)`
```

The `Throwing...` technically shall become deprecated/removed if we manage to make typed throws happen for them 🤔 There's another proposal in flight for this

proposals/NNNN-isolated-any-functions.md
```diff
@@ -0,0 +1,516 @@
...
+
+This proposal modifies all of these APIs so that the task function has
+`@isolated(any)` function type.  These APIs now all synchronously enqueue
+the new task directly on the appropriate executor for the task function's
+dynamic isolation.
+
+Swift reserves the right to optimize the execution of tasks to avoid
+"unnecessary" isolation changes, such as when an isolated `async` function
+starts by calling a function with different isolation.  In general, this
+includes optimizing where the task initially starts executing.  As an
```
ktoso, line 353: This is probably worthy of an example as that's the "we'll reorder things" case that people should watch out for.

Maybe a bit here?

```suggestion
includes optimizing where the task initially starts executing:

```swift
@MainActor class Helper { 
  let other: OtherActor
  
  func help() {
    Task { await other.work() } // *implicitly* starts on MainActor, however 
    Task { await other.work() } // this may be optimized away and start on `other` immediately
  }
\``` // FIXME: can't suggest change with a \``` inside it, remove the \

In this example, `Task` is implicitly inheriting the enclosing context's MainActor isolation, however the task immediately calls code on another actor. The Swift runtime reserves the right to optimize away the initial "starting the task" enqueue on the MainActor, and instead enqueue on the `other` immediately.

As an
```

Did I get the semantics right of what you're proposing here?

ktoso (approved):

proposals/NNNN-isolated-any-functions.md
rjmccall: Oops, thanks.  An earlier draft had a lot more links, and I guess I deleted all the ref definitions.  Fixed.

proposals/NNNN-isolated-any-functions.md
```diff
@@ -0,0 +1,561 @@
...
+
+```swift
+await operation(timePassed / overallDuration)
+```
+
+The isolation can be read using the special `isolation` property
+of these types:
+
+```swift
+func traverse(operation: isolated(any) (Node) -> ()) {
```
stzn, line 170: missing "@"?

```suggestion
func traverse(operation: @isolated(any) (Node) -> ()) {
```

proposals/NNNN-isolated-any-functions.md
```diff
@@ -0,0 +1,671 @@
...
+  - It is also the default if other rules don't apply.
+
+- They can be isolated to a specific [global actor][SE-0316] type.
+
+  - This can be explicitly expressed with a global actor attribute,
+    such as `@MainActor`.
+  - It can also be inferred from context in a number of ways, such as
+    if the function is a method of a type with that attribute (which
+    itself can be inferred in a number of ways).
+  - Additionally, a closure passed directly to the `Task` initializer
```
DougGregor, line 48: A small drafting suggestion: the bullets here are interspersing the kinds of isolation with the ways in which isolation is inferred, making it harder to keep both separate in the reader's mind. Perhaps lay out the kinds of isolation here, and move the "inference" bits until later?

proposals/NNNN-isolated-any-functions.md
```diff
@@ -0,0 +1,671 @@
...
+  specified isolation of `F`:
+
+  - If `F` has an `isolated` parameter, the conversion is invalid.
+  - Otherwise, the conversion is valid, and the dynamic isolation of
+    the resulting function is determined as follows:
+    - If the converted value is the result of an expression that is a
+      closure expression (including an implicit autoclosure), a function
+      reference, or a partial application of a method reference, the
+      resulting function is dynamically isolated to the isolation of the
+      function or closure.  This looks through non-instrumental differences
```
DougGregor, line 245: "non-instrumental differences" isn't a phrase I've ever seen us use. Perhaps "looks though syntax that has no effect on the semantics of the expression, such as parentheses"?

proposals/NNNN-isolated-any-functions.md
```diff
@@ -0,0 +1,671 @@
...
+to a derivation of the expression `f.isolation`.
+
+A context is isolated to a derivation of a expression `E` if it has an
+isolated capture that is initialized to a derivation of `E`.
+
+An expression is a derivation of `E` if:
+
+- it has the exact form of `E`;
+- it is a reference to a capture or immutable binding immediately
+  initialized with a derivation of `E`; or
```
DougGregor, line 327: IIRC, the current isolation analysis for isolated parameters does not look through an "immutable binding immediately initialized with a derivation of `E`", e.g.,

```swift
actor A {
  func g() { }
}

func f(actor: isolated A) {
  let a = actor
  a.g() // error: a is not considered to be in the same isolation domain, call needs to be asynchronous
}
```

If we want to generalize that, great, but we should do so consistently rather than make it a rule specific to `.isolation` on an `@isolated(any)` type.

proposals/NNNN-isolated-any-functions.md
```diff
@@ -0,0 +1,671 @@
...
+    print("waking")
+    operation() // <-- does not cross an isolation barrier and so is synchronous
+    print("finished")
+  }
+}
+```
+
+(This uses the `isolated` capture modifier proposed in the draft
+[closure isolation control][isolated-captures] proposal.  It would be
+possible but significantly more awkward to get this effect with that
```
DougGregor, line 350:
```suggestion
possible but significantly more awkward to get this effect without that
```

proposals/NNNN-isolated-any-functions.md
```diff
@@ -0,0 +1,671 @@
...
+`.isolation`.
+
+As an exception, in order to provide a primitive scheduling operation with
+stronger guarantees, Swift will always start a task function on the
+appropriate executor for its formal dynamic isolation unless:
+- it is non-isolated or
+- it comes from a closure expression that is only *implicitly* isolated
+  to an actor (that is, it has neither an explicit `isolated` capture
+  nor a global actor attribute).  This can currently only happen with
+  `Task {}`.
```
DougGregor, line 402: This came up in the [pitch thread](https://forums.swift.org/t/isolated-any-function-types/70562/8), and you [answered it there](https://forums.swift.org/t/isolated-any-function-types/70562/10). The proposal would be stronger if it included this justification.

DougGregor (commented):
Looking good; just a handful of comments.

DougGregor (approved):

PR #2360, opened 2024-03-13
Fix typo in SE-0110 link that causes 404 error

dempseyatgithub:
The proposal link for SE-0110 returns a 404 error.
This PR fixes the typo that causes the error.


DougGregor (approved):
Thanks!

PR #2365, opened 2024-03-16
Fixed typo in 0314-async-stream.md

malhal:
Fixed the typo and added note about a side effect

### DIFF
--- a/proposals/0314-async-stream.md
+++ b/proposals/0314-async-stream.md
@@ -114,8 +114,8 @@ extension QuakeMonitor {
       monitor.quakeHandler { quake in
         continuation.yield(quake)
       }
-      monitor.onTermination = { _ in 
-        monitor.stopMonitoring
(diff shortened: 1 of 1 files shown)


proposals/0314-async-stream.md
```diff
@@ -114,8 +114,8 @@ extension QuakeMonitor {
       monitor.quakeHandler { quake in
         continuation.yield(quake)
       }
-      monitor.onTermination = { _ in 
-        monitor.stopMonitoring
+      continuation.onTermination = { _ in 
+        monitor.stopMonitoring // also retains the monitor, preventing deinit
```
rjmccall, line 118: This comment doesn't seem like something that needs to be in a proposal; it's not explaining anything important to readers.
```suggestion
        monitor.stopMonitoring()
```

I assume we do need to call `stopMonitoring`, though.

PR #2366, opened 2024-03-17
Normalize markdown in SE-0119 to improve dashboard metadata extraction

dempseyatgithub:
Remove unnecessary HTML tags in SE-0119 that interfere with dashboard metadata extraction.

There is no change to content or appearance of paragraphs. Code voice in one paragraph now appears correctly as was seemingly intended.


rjmccall:
Thanks!

PR #2372, opened 2024-03-20
Add a proposal to improve usability of global-actor-isolated types.

simanerush:


proposals/NNNN-global-actor-isolated-types-usability.md
```diff
@@ -0,0 +1,267 @@
...
+class NonSendable {
+  func test() { ... }
+}
+
+@MainActor
+class IsolatedSubclass: NonSendable {
+  var mutable = 0
+  override func test() {
+    super.test()
+    mutable += 0 // error: Main actor-isolated property 'isolated' can not be referenced from a non-isolated context
```
mattmassicotte, line 241: Super minor, but I think this error should be `mutable` ?

proposals/NNNN-global-actor-isolated-types-usability.md
```diff
@@ -0,0 +1,267 @@
...
+  }
+
+  Task {
+    // error: capture of 'closure' with non-sendable type '@MainActor () -> Void' in a `@Sendable` closure
+    await closure()
+  }
+}
+```
+
+In the above code, the closure is global-actor-isolated, so it cannot be called concurrently. The compiler should be able to infer the `@Sendable` attribute. Because of the same reason, globally isolated closures should be allowed to capture non-`Sendable` values.
```
rjmccall, line 49:
```suggestion
Requiring both a global actor attribute and `@Sendable` creates an unfortunate pile-up of attributes, and it would be better for Swift to just infer `@Sendable` from the global actor attribute.

Because a globally-isolated closure cannot be called concurrently, it's safe for it to capture non-`Sendable` values even if it's implicitly `@Sendable`.  Such values just need to be transferred to the global actor's region (if they aren't there already).  This also applies to closures that are isolated to a specific actor reference.
```

proposals/NNNN-global-actor-isolated-types-usability.md
```diff
@@ -0,0 +1,267 @@
...
+}
+```
+
+In the above code, an instance of `Subclass` can be passed across isolation boundaries because `@MainActor` implies that the type is `Sendable`. However, `Subclass` inherits non-isolated, mutable state from the superclass, so this `Sendable` conformance allows smuggling unprotected shared mutable state across isolation boundaries to create potential for concurrent access. For this reason, the warning about adding isolation to a subclass was added in Swift 5.10, but this restriction could be lifted by instead preventing the subclass from being `Sendable`.
+
+## Proposed solution
+
+We propose that:
+
+- `Sendable` properties of a global-actor-isolated value type would be treated `nonisolated` as inferred within the module.
```
rjmccall, line 98:
```suggestion
- Stored properties of `Sendable` type in a global-actor-isolated value type can be declared as `nonisolated` without using `(unsafe)`.
- Stored properties of `Sendable` type in a global-actor-isolated value type are treated as `nonisolated` when used within the module or if the value type is `frozen`.
```

proposals/NNNN-global-actor-isolated-types-usability.md
```diff
@@ -0,0 +1,267 @@
...
+- `@Sendable` would be inferred for global-actor-isolated functions and closures. Additionally, globally isolated closures would be allowed to capture non-`Sendable` values.
```
rjmccall, line 99:
```suggestion
- `@Sendable` is inferred for global-actor-isolated functions and closures.
- Global-actor-isolated closures are allowed to capture non-`Sendable` values despite being `@Sendable`.
```

proposals/NNNN-global-actor-isolated-types-usability.md
```diff
@@ -0,0 +1,267 @@
...
+- The programmer would be able to suppress the automatic conformance inferred via the above rule using the new `@~Sendable` attribute. By analogy, introduce a new `~Sendable` protocol to indicate that a nominal type is not `Sendable`.
```
rjmccall, line 100: I'm not sure this carries its weight.  Do we have any use for this?

proposals/NNNN-global-actor-isolated-types-usability.md
```diff
@@ -0,0 +1,267 @@
...
+- Require the global-actor-isolated subclass of a `nonisolated`, non-`Sendable` to be non-`Sendable`.
```
rjmccall, line 101:
```suggestion
- A global-actor-isolated subclass of a non-isolated, non-`Sendable` class is allowed, but it must be non-`Sendable`.
```

proposals/NNNN-global-actor-isolated-types-usability.md
```diff
@@ -0,0 +1,267 @@
...
+}
+
+extension S: Equatable {
+  static nonisolated func ==(lhs: S, rhs: S) -> Bool {
+    return lhs.x == rhs.x // okay
+  }
+}
+```
+
+In the above code, `x` is implicitly `nonisolated` within the module. Under this proposal, `nonisolated` is inferred for within the module access of `Sendable` properties of a global-actor-isolated value type. This is data-race safe because the property belongs to a value type, meaning it will be copied every time it crosses an isolation boundary.
```
rjmccall, line 124: You don't need to re-justify things here that you already justified in the earlier sections, and you can let the section title introduce the section.

This paragraph is a little confusing.  You need to be clear that you're not talking about actually inferring `nonisolated` — you're saying that certain accesses will be allowed *as if* the property were declared `nonisolated`.  You should say that this rules only applies to stored properties, and you may need to discuss the impact on property wrappers and `lazy`.  You should mention that the same effect happens across modules if the type is `@frozen` (right?).

Please use both positive and negative examples of things that are now allowed and still not allowed.  I think you should include an example of a mutation of a property.

proposals/NNNN-global-actor-isolated-types-usability.md
```diff
@@ -0,0 +1,267 @@
...
+
+The programmer can still choose to mark the property `nonisolated` to allow synchronous access from outside the module. Requiring asynchronous access from outside the module preserves the ability for library authors to change a stored property to a computed property without breaking clients, and the library author may explicitly write `nonisolated` to opt-into synchronous access as part of the API contract.
```
rjmccall, line 126:
```suggestion
The programmer can still choose to explicitly mark a stored property `nonisolated` to allow synchronous access from outside the module. It is not necessary to use `nonisolated(unsafe)` if the property has `Sendable` type and the property is of a value type. Once added, `nonisolated` cannot later be removed without potentially breaking clients. The programmer can still convert the property to a computed property, but it has to be a `nonisolated` computed property.
```

rjmccall (commented):
First batch of feedback.

proposals/NNNN-global-actor-isolated-types-usability.md
simanerush: I have discussed this with Holly and we also feel like this change can be omitted because there's no obvious use case. We can note this in the Forums discussion and remove this change unless someone comes up with a use case.

proposals/NNNN-global-actor-isolated-types-usability.md
```diff
@@ -0,0 +1,267 @@
...
+extension S: Equatable {
+  static nonisolated func ==(lhs: S, rhs: S) -> Bool {
+    return lhs.x == rhs.x
+  }
+}
+```
+
+The above code is perfectly safe and should not require an unsafe opt-out. Since `S` is a value type and `x` is of a `Sendable` type `Int`, it should not be unsafe to declare `x` non-isolated, because when accessing `x` from different concurrency domains, we will be operating on a copy of the value type, and the result type is `Sendable` so it's safe to return the same value across different isolation domains. Because access to `x` across concurrency domains is always safe, `nonisolated` should be implicit within the module, similar to actor-isolated `let` constants with `Sendable` type.
+
+Next, under the current concurrency rules, globally isolated functions and closures do not implicitly conform to `Sendable`. This impacts usability, because these closures cannot themselves be captured by `@Sendable` closures, which makes them unusable with `Task`:
```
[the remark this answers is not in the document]
simanerush: I feel like we meant to highlight that if the programmer's intention was to make the function globally isolated, then we _should_ infer Sendable by extension of global isolation properties, but your suggestion makes perfect sense- if the programmer meant to make the function non-Sendable, then it turns out that it is unnecessary to mark it globally isolated. But this was true before this proposal, and with the suggestion, I'm not sure how to segue to the specific problem we're trying to address.

proposals/NNNN-global-actor-isolated-types-usability.md
simanerush: Thank you!! We are working on rewriting this paragraph.

I have some questions about the negative example: 

The mutation of a property is still not allowed, because the only way to do so is to use `inout`. As a consequence, the property becomes non-`Sendable` because it is now wrapped in `LValueType`, so our change has no effect. I don't think it's a good "negative" example just because of how `inout` works.

I could include an example where the mutated version of the value type is assigned into a new value and returned, but it would make it obvious that it was just copied. So I'm not sure whether such an example would be worth to include.

rjmccall: As I understand it, mutation should be a positive example.  This code should be allowed under the proposal:

```swift
@MainActor struct A {
  var x: Int
}

func foo() {
  var a = A(x: 5)
  a.x += 5
}
```

Currently it is not:

```text
red.swift:7:5: error: main actor-isolated property 'x' can not be mutated from a non-isolated context
1 │ @MainActor struct A {
2 │   var x: Int
  │       ╰─ note: mutation of this property is only permitted within the actor
3 │ }
4 │ 
5 │ func foo() {
  │      ╰─ note: add '@MainActor' to make global function 'foo()' part of global actor 'MainActor'
6 │   var a = A(x: 5)
7 │   a.x += 5
  │     ╰─ error: main actor-isolated property 'x' can not be mutated from a non-isolated context
8 │ }
```

simanerush: Under this proposal, this code still gives me an error
```
119 │ func foo() {
120 │   var a = A(x: 5)
121 │   a.x += 5
    │   ╰─ warning: non-sendable type '@lvalue A' passed in call to main actor-isolated property 'x' cannot cross actor boundary; this is an error in the Swift 6 language mode
122 │ }
123 │ 
```
This is because Sendable can't be inferred from LValue type

rjmccall: That certainly seems like something we should fix.

simanerush: Yeah, make LValue types conform to Sendable?

proposals/NNNN-global-actor-isolated-types-usability.md
```diff
@@ -0,0 +1,233 @@
...
+}
+
+extension S: Equatable {
+  static nonisolated func ==(lhs: S, rhs: S) -> Bool {
+    return lhs.x == rhs.x
+  }
+}
+```
+
+Because `S` is a value type and `x` has the `Sendable` type `Int`, it is never unsafe in itself to use `x` from different concurrency domains.  `x` is mutable, but since `S` is a value type, any mutation of `x` is always part of a mutation of the containing `S` value, and the concurrency model will prevent data races at that level without needing any extra rules for `x`.  If we do have concurrent accesses to `x` on the same `S` value, they must both be reads, and it's fine to have concurrent reads of the same value as long as it's `Sendable`.  So, first off, it should be possible to declare `x` as `nonisolated` without adding `(unsafe)`.
```
rjmccall, line 32:
```suggestion
However, there is nothing unsafe about treating `x` as nonisolated.  The general rule is that concurrency is safe as long as there aren't data races.  The type of `x` conforms to `Sendable`, and using a value of `Sendable` type from multiple concurrent contexts shouldn't ever introduce a data race, so any data race involved with an access to `x` would have to be on memory in which `x` is stored.  But `x` is part of a value type, which means any access to it is always also an access to the containing `S` value.  As long as Swift is properly preventing data races on that larger access, it's always safe to access the `x` part of it.  So, first off, there's no reason for Swift to require `(unsafe)` when marking `x` `nonisolated`.
```

proposals/NNNN-global-actor-isolated-types-usability.md
```diff
@@ -0,0 +1,233 @@
...
+
+We can do better than that, though.  The only problem with treating `x` as *implicitly* `nonisolated` is source and binary stability: someone could reasonably change `x` to be a computed property in the future, and the getter and setter for that might need to be global-actor-isolated.  Within the module, though, we know that hasn't happened.  So Swift should treat stored properties like `x` as implicitly `nonisolated` when they're used from the same module, or when the containing type is declared `frozen`.
```
rjmccall, line 34: Holly asked me to revert the suggestion to allow `@frozen`.

```suggestion
We can do better than that, though.  It should be possible to treat a `var` stored property of a global-actor value type as *implicitly* `nonisolated` under the same conditions that a `let` property can be.  A stored property from a different module can be changed to a computed property in the future, and those future computed accessors may need to be isolated to the global actor, so allowing access across module boundaries would not be okay for source or binary compatibility.  But within the module that defines the property, we know that hasn't happened, so it's fine to use a more relaxed rule.
```

proposals/NNNN-global-actor-isolated-types-usability.md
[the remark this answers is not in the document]
rjmccall: I've tweaked my suggestion.  I'd just make the example a positive example of a function that takes a `@MainActor` function and show how it's now possible to send it.

proposals/NNNN-global-actor-isolated-types-usability.md
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  Task {
+    // error: capture of 'closure' with non-sendable type '@MainActor () -> Void' in a `@Sendable` closure
+    await closure()
+  }
+}
+```
+
+Requiring both a global actor attribute and `@Sendable` creates an unfortunate pile-up of attributes, and it would be better to infer `@Sendable` from the global actor attribute.
```
rjmccall, line 51: My suggestion above merges this into the previous paragraph.

```suggestion
```

proposals/NNNN-global-actor-isolated-types-usability.md
```diff
@@ -0,0 +1,233 @@
...
+
+Because a globally-isolated closure cannot be called concurrently, it's safe for it to capture non-`Sendable` values even if it's implicitly `@Sendable`.  Such values just need to be transferred to the global actor's region (if they aren't there already).  This also applies to closures that are isolated to a specific actor reference.
```
rjmccall, line 53:
```suggestion
Because a globally-isolated closure cannot be called concurrently, it's safe for it to capture non-`Sendable` values even if it's implicitly `@Sendable`.  Such values just need to be transferred to the global actor's region (if they aren't there already).  The same logic also applies to closures that are isolated to a specific actor reference, although it isn't currently possible to write such a closure in a context that isn't isolated to that actor.
```

proposals/NNNN-global-actor-isolated-types-usability.md
```diff
@@ -0,0 +1,233 @@
...
+```
+
+In the above code, an instance of `Subclass` can be passed across isolation boundaries because `@MainActor` implies that the type is `Sendable`. However, `Subclass` inherits non-isolated, mutable state from the superclass, so this `Sendable` conformance allows smuggling unprotected shared mutable state across isolation boundaries to create potential for concurrent access. For this reason, the warning about adding isolation to a subclass was added in Swift 5.10, but this restriction could be lifted by instead preventing the subclass from being `Sendable`.
+
+## Proposed solution
+
+We propose that:
+
+- Stored properties of `Sendable` type in a global-actor-isolated value type can be declared as `nonisolated` without using `(unsafe)`.
+- Stored properties of `Sendable` type in a global-actor-isolated value type are treated as `nonisolated` when used within the module or if the value type is `frozen`.
```
rjmccall, line 103:
```suggestion
- Stored properties of `Sendable` type in a global-actor-isolated value type are treated as `nonisolated` when used within the module that defines the property.
```

proposals/NNNN-global-actor-isolated-types-usability.md
rjmccall: The analysis probably needs to think more carefully about `inout` parameters, but yes, something like that.

proposals/NNNN-global-actor-isolated-types-usability.md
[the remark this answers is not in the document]
simanerush: I see, thank you John!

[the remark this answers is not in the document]
simanerush: >  and show how it's now possible to send it.
The "Motivation" section of this proposal shows the current state of things, so do you think I should show how it's _impossible_ to send it in this section?
I can then reuse this new example in the "Detailed Design" (as it is currently done with the closure example) to show how it's possible to send the function under this proposal.

proposals/NNNN-global-actor-isolated-types-usability.md
```diff
@@ -0,0 +1,225 @@
...
+
+```swift
+func test(globallyIsolated: @escaping @MainActor () -> Void) {
+  Task {
+    await globallyIsolated() //okay
+  }
+}
+```
+
+The `globallyIsolated` closure in the above code is global-actor isolated via the `@MainActor`. Thus, it can never operate on the same reference concurrently at the same time, making it safe to be invoked from different isolation domains. This means that for such global-actor-isolated closures and functions, the `@Sendable` attribute is implicit.
```
rjmccall, line 141:
```suggestion
The `globallyIsolated` closure in the above code is global-actor isolated because it has the `@MainActor` attribute. Because it will always run isolated, it's fine for it to capture and use values that are isolated the same way. It's also safe to share it with other isolation domains because the captured values are never directly exposed to those isolation domains. This means that there's no reason not to always treat these functions as `@Sendable`.
```

proposals/NNNN-global-actor-isolated-types-usability.md
```diff
@@ -0,0 +1,225 @@
...
+* Erasing to an existential type based on conformances of the superclass type
+* Passing the isolated subclass as a generic argument to a type parameter that requires a conformance implemented by the superclass
+
+## Source compatibility
+
+The introduced changes are additive, so the proposal does not impact source compatibility.
+
+## ABI compatibility
+
+This proposal should have no impact on ABI compatibility.
```
rjmccall, line 216: I believe the global-actor-implies-`@Sendable` part of this proposal will change the ABI for any existing functions that take a non-`Sendable` but global-actor-isolated function.  I believe the idea is just that hopefully no such functions actually exist.

proposals/NNNN-global-actor-isolated-types-usability.md
```diff
@@ -0,0 +1,225 @@
...
+
+Matching the isolation of the superclass method is necessary because the superclass method implementation may internally rely on the static isolation, such as when hopping back to the isolation after any asynchronous calls, and because there are a variety of ways to call the subclass method that don't preserve its isolation, including:
+
+* Upcasting to the superclass type
+* Erasing to an existential type based on conformances of the superclass type
+* Passing the isolated subclass as a generic argument to a type parameter that requires a conformance implemented by the superclass
+
+## Source compatibility
+
+The introduced changes are additive, so the proposal does not impact source compatibility.
```
rjmccall, line 212: The global-actor-implies-`@Sendable` part of this proposal is not additive because it changes the interpretation of existing code.  A call to a function which takes a global-actor-isolated-but-non-`@Sendable` function can become ill-formed under this proposal.

rjmccall (changes requested):
You should talk to @hborla about the compatibility sections.

proposals/NNNN-global-actor-isolated-types-usability.md
simanerush: Yes. Me and @hborla have discussed the potential effects on the name mangling, so I will update this

proposals/NNNN-global-actor-isolated-types-usability.md
simanerush: Yeah, agree

rjmccall (approved):

PR #2375, opened 2024-03-25
SE-0364 is implemented in Swift 6.0

ikesyo:
Please correct me if I am wrong.


rjmccall:
I believe we generally don't update those things until after the release.

rjmccall:
I'm sorry, I should've checked that before replying.  We do do this ahead of time.

ikesyo:
Yeah, there are 9 proposals whose status is `Implemented (Swift 6.0)` at the moment 😄

PR #2378, opened 2024-03-26
SE-0427: A few minor clarifications

slavapestov:
- Motivate conditional conformance restrictions better
- Mention that you don't have to write ~Copyable if you're declaring a conditional conformance
- Better explain what's coming in the ABI section


proposals/0427-noncopyable-generics.md
```diff
@@ -405,35 +404,93 @@ Note that no `where` clause needs to be written, because by the rules above,
...
+   protocol Manager { associatedtype Resource: ~Copyable }
+   struct ManagerManager<T: Manager>: ~Copyable {}
+   extension ManagerManager: Copyable where T.Resource: Copyable {}  // error
+   ```
+   To check the conditional requirement `T.Resource: Copyable`, we would
+   need to resolve the `Resource` associated type from the conformance of `T` to `Manager`,
+   which again could be retroactive.
+
+As a consequence of existing language rules around synthesized conformances
+together the first restriction above, the default conformance on the type
```
tbkka, line 439: "... together **with** the ... "

PR #2389, opened 2024-04-04
SE-0400: State limitations

AnthonyLatsis:
cc @hborla

See https://github.com/apple/swift/issues/71578.

### DIFF
--- a/proposals/0400-init-accessors.md
+++ b/proposals/0400-init-accessors.md
@@ -491,6 +491,11 @@ init() {
 }
 ```
 
+### Restrictions
+
+A property with an `init` accessor can only be declared in the primary
+declaration of a type.
+
 ## Source compatibility
 
 `init` accessors are an additive capability with new syntax; there is no impact on existing source code.
--- a/proposals/0400-init-accessors.md
+++ b/proposals/0400-init-accessors.md
@@ -502,7 +502,10 @@ declaration of a type.
 
 ## ABI compatibility
 
-`init` accessors are only called from within a module, so they are not part of the module's ABI. In cases where a type's initializer is `@inlinable`, the body of an `init` accessor must also be inlinable.
+`init` accessors are an ABI-additive change; they are at most `internal` but can
+be ABI-public.
+Calling an `init` accessor from an `inlinable` type initializer requires that
+the `init` accessor is ABI-public.
 
 ## Implications on adoption
 


proposals/0400-init-accessors.md
```diff
@@ -491,13 +491,22 @@ init() {
 }
 ```
 
+### Restrictions
+
+* A property with an `init` accessor must be declared in the same source file 
+  as the respective type declaration.
+* An `init` accessor is always `fileprivate` and thus cannot be called from
+  within a different source file.
+
```
AnthonyLatsis, line 500: I am still trying to make sense of the first one (it was inferred from [this](https://forums.swift.org/t/pitch-init-accessors/64881/16) post). If an `init` accessor is always confined to its file, then, logically, it cannot break a memberwise initializer generated in another file that isn’t aware of its existence.

AnthonyLatsis:
@hborla ping

proposals/0400-init-accessors.md
```diff
@@ -491,13 +491,22 @@ init() {
...
 ## Source compatibility
 
 `init` accessors are an additive capability with new syntax; there is no impact on existing source code.
 
 ## ABI compatibility
 
-`init` accessors are only called from within a module, so they are not part of the module's ABI. In cases where a type's initializer is `@inlinable`, the body of an `init` accessor must also be inlinable.
+`init` accessors are only called from within a source file, so they are not part of the module's ABI.
```
hborla, line 507: I think the original text is correct. Even if you can only define init accessors in the primary declaration of a type, you can add an initializer in another source file that will call the init accessor:

```swift
// in FirstFile.swift

struct S {
  var _x: Int

  var x: Int {
    @storageRestrictions(initializes: _x)
    init {
      self._x = newValue
    }
    get { _x }
  }
}

// in SecondFile.swift

extension S {
  init() {
    self.x = 10
  }
}
```

proposals/0400-init-accessors.md
hborla: I think this should just say that init accessors can only be written in the primary declaration of a type.

AnthonyLatsis:
```suggestion
### Restrictions

A property with an `init` accessor can only be declared in the primary 
declaration of a type.

```

proposals/0400-init-accessors.md
```diff
@@ -491,13 +491,22 @@ init() {
...
+An `init` accessor is *not* inferred as, but must be, inlinable if called from
+an inlinable type initializer.
```
AnthonyLatsis, line 509:
```suggestion
`init` accessors can be ABI-public but are at most `internal`.
Calling an `init` accessor from an `inlinable` type initializer requires that
the `init` accessor is ABI-public.
```

proposals/0400-init-accessors.md
AnthonyLatsis: Thanks. We had a crash report with like the simplest example of an init accessor use across files, and I was not sure if this sentence really meant they could be `internal` because the forum topics appeared to settle on `fileprivate`. I could have just inferred it from the `inlinable` business that follows, really. But I still think this paragraph would benefit from a little more straightforwardness and accuracy (see the latest revision).

hborla (approved):
Sorry for the delay and thanks for the ping :)

PR #2395, opened 2024-04-10
[BitwiseCopyable] Post-review changes.

nate-chandler:


allevato (approved):
Thanks for these revisions! We discussed it this week in the language steering group and will start the second review shortly.

PR #2402, opened 2024-04-15
Correct the chronological order of proposal review discussions

dempseyatgithub:
The forum discussions listed in the "Review" heading field should be in chronological order.

This PR corrects the chronological order of proposals and, in those proposals, also normalizes naming, order, formatting to match the format and directions in the [profile template](https://github.com/apple/swift-evolution/blob/main/proposal-templates/0000-swift-template.md):

> - Review: ([pitch](https://forums.swift.org/...))

> `Review` is a history of all discussion threads about this proposal, in chronological order. Use these standardized link names: `pitch` `review` `revision` `acceptance` `rejection`. If there are multiple such threads, spell the ordinal out: `first pitch` `second review` etc.

This is in preparation for changes to the SE Dashboard metadata extractor where review discussions will be included.

Note to keep the size of this PR manageable, it only focuses on proposals with discussions in incorrect chronological order.


DougGregor (approved):
Thank you!

PR #2403, opened 2024-04-16
Update status of 0220-count-where.md

glessard:


glessard:
@benrimmington already did this in https://github.com/apple/swift-evolution/pull/2391, closing.

PR #2407, opened 2024-04-17
Normalize SE-0433 Review header field

dempseyatgithub:
Update heading field for SE-0433 to 'Review' and adjust formatting to match proposal template.


rjmccall:
Thanks!

PR #2413, opened 2024-04-22
Add missing Review field to SE-0432

dempseyatgithub:
SE-0432 is missing its Review header field and link to the review.

This PR also uses normalized naming of Previous Revision link as described in proposal template.


DougGregor (approved):
Thank you!

PR #2416, opened 2024-04-23
Remove upcoming feature flag from SE-0429

dempseyatgithub:
Remove erroneous upcoming feature flag from SE-0429 as per https://forums.swift.org/t/accepted-se-0429-partial-consumption-of-noncopyable-values/70972/3.

> MoveOnlyPartialConsumption is not an upcoming feature in the compiler implementation, and you're right that it doesn't need an upcoming feature flag.


xwu:
Thanks!

PR #2419, opened 2024-04-24
SE-0366: rename document for clarity

glessard:
SE-0366's proposal document is named after the original pitched version, which is quite different from what swift-evolution eventually settled upon. This renames the document so that it's easier to find by filesystem inspection.


rjmccall:
We don't rename proposal documents once committed, since it breaks existing links.

PR #2429, opened 2024-05-01
Revise SE-0431 to remove the `@Sendable` implication.

rjmccall:
Also, discuss some of the syntax alternatives we explored.


DougGregor (approved):
Thank you, John.

PR #2430, opened 2024-05-02
SE-0428: adjust macro name to SE accepted name: @Resolvable

ktoso:
The proposal was accepted under condition of renaming the macro name: https://forums.swift.org/t/accepted-with-modifications-se-0428-resolve-distributedactor-protocols/71366

Adjust the proposal to use the accepted name


xedin (approved):
cc @Jumhyn

PR #2433, opened 2024-05-03
[SE-0415] Narrow scope down to just "body" macros.

DougGregor:


proposals/0415-function-body-macros.md
```diff
@@ -123,12 +121,12 @@ When using the shorthand syntax for get-only properties, a function body macro c
 
 ### Declaring function body macros
 
-Function body macros are declared with the `body` or `preamble` role, which indicate that they can be attached to any kind of function, and can produce the contents of a function body. For example, here are declarations for the macros used above:
+Function body macros are declared with the `body`, which indicate that they can be attached to any kind of function, and can produce the contents of a function body. For example, here are declarations for the macros used above:
```
PsychoH13, line 124:
```suggestion
Function body macros are declared with the `body` role, which indicate that they can be attached to any kind of function, and can produce the contents of a function body. For example, here are declarations for the macros used above:
```

allevato (approved):

PR #2435, opened 2024-05-03
[transferring-params-and-args] Rename transferring -> sendable.

gottesmm:
Just an initial draft where I changed transferring -> sendable and added an alternatives considered section.

### DIFF
--- a/proposals/0430-transferring-parameters-and-results.md
+++ b/proposals/0430-transferring-parameters-and-results.md
@@ -3,7 +3,7 @@
 * Proposal: [SE-0430](0430-transferring-parameters-and-results.md)
 * Authors: [Michael Gottesman](https://github.com/gottesmm), [Holly Borla](https://github.com/hborla), [John McCall](https://github.com/rjmccall)
 * Review Manager: [Becca Royal-Gordon](https://github.com/beccadax)
-* Status: **Returned for Revision** 
+* Status: **Active Review (May 7...May 13, 2024)** 
 * Implementation: On `main` gated behind `-enable-experimental-feature TransferringArgsAndResults`
 * Previous Proposal: [SE-0414: Region-based isolation](/proposals/0414-region-based-isolation.md)
 * Review: ([pitch](https://forums.swift.org/t/pitch-transferring-isolation-regions-of-parameter-and-result-values/70240)) ([first review](https://forums.swift.org/t/se-0430-transferring-isolation-regions-of-parameter-and-result-values/70830)) ([revision](https://forums.swift.org/t/returned-for-revision-se-0430-transferring-isolation-regions-of-parameter-and-result-values/71297))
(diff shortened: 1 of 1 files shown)


proposals/0430-transferring-parameters-and-results.md
```diff
@@ -31,20 +32,22 @@ class NonSendable {}
...
 ```
 
-Actor initializers have a special rule that allows transferring its parameter
-values into the actor-isolated region. Actor initializers are `nonisolated`, so
-a call to an actor initializer does not cross an isolation boundary, meaning
-the argument values would be usable in the caller after the initializer returns
-under the standard region isolation rules. SE-0414 consider actor initializer
-parameters as being transferred into the actor's region to allow initializing
-actor-isolated state with those values:
+Actor initializers have a special rule that requires its parameter values to be
```
rjmccall, line 44:
```suggestion
Actor initializers have a special rule that requires their parameter values to be
```

proposals/0430-transferring-parameters-and-results.md
```diff
@@ -109,181 +114,227 @@ nonisolated func test() async {
...
-A `transferring` parameter requires the argument value to be in a disconnected
-region. At the point of the call, the disconnected region is transferred away
-and cannot be used in the caller's isolation domain after the transfer,
-allowing the callee to transfer the parameter value to a region that is opaque
-to the caller:
+### Sendable Values and Sendable Types
+
+A `sendable` value is defined as a value that at its initialization point
+possesses the capability of being safely sent across an isolation boundary. For
+values that do not conform to the `Sendable` protocol this is achieved by
```
rjmccall, line 162:
```suggestion
values that do not conform to the `Sendable` protocol, this is achieved by
```

proposals/0430-transferring-parameters-and-results.md
```diff
@@ -109,181 +114,227 @@ nonisolated func test() async {
...
+specifying that such a value must be in a disconnected region at its
+initialization point. Thus a newly defined value that does not conform to the
+`Sendable` protocol is always a `sendable` value:
```
rjmccall, line 165: Should this paragraph be a little more user-focused?  Maybe:

```
A type that conforms to the `Sendable` protocol is a thread-safe type:
its values can be shared with and used from multiple concurrent contexts
at once without causing data races.  If a value's type is not `Sendable`,
Swift must ensure that it is never used concurrently.  The value can still
be sent between concurrent contexts, but the send must be a complete
transfer of the value's region: all uses of the value (or anything
non-`Sendable` that can be reached through it) must end on the original
context before they can begin on the new context.  For Swift to enforce
this, the value must be in a disconnected region.  Such a value is said to
be a `sendable` value.

Any newly-created value with no connections to existing regions is
always a `sendable` value:
```

beccadax (approved):

beccadax:
Needs a rebase, as the updates are based on a version of the proposal with an outdated `Review:` line.

PR #2447, opened 2024-05-10
SE-0306 Actors: Make Types consistent in re-entrancy sample code

Diggory:
I have been reading up on Actors, and think that there are some Type name errors in the sample code.  Forgive me if I have misunderstood and got this wrong....

In the re-entrancy section there is some sample code showing how a value can change across `await`s.

The types don't match - I suspect that at some point `Person` was changed to `DecisionMaker` but not all occurrences of the type name were changed in the sample code.

The Actor type should either be a `Person` or a `DecisionMaker` Similarly the `opinion` property's type should either be `Judgment` or 'Decision'


ktoso (approved):
Yeah seems the type was renamed huh, thanks for catching it

PR #2481, opened 2024-06-11
Homogenize KeyPath and update diagnostic.

amritpan:


rjmccall (changes requested):
The concept is "key path", and documentation like TSPL is consistent about spelling it out as two words.  `KeyPath` is the type name in Swift because you can't have spaces in identifiers.  Basically all of these places should use the former.  I can also see multiple places in the document that are still using `keypath`, just looking at the diff; please do a word search.

PR #2489, opened 2024-06-14
[SE-0437] Amend with small corrections

lorentey:
- Fix accidental omission of `UnboundedRange`-taking overloads of `extracting`.
- Reflect SE-0427’s outcome and add explicit `where Foo: Copyable` clauses for conditional conformances to Copyable.
- Fix typo where `Result` forgot to declare itself `~Copyable`
- Fix typo where an `withExtendedLifetime(_:_:)` overload was shown with a signature requiring copyable Ts.


rjmccall (approved):
Approved by the LSG; no need for a review.

PR #2496, opened 2024-06-19
Mark a bunch of proposals as implemented in Swift 6.0.

hborla:


proposals/0414-region-based-isolation.md
```diff
@@ -3,7 +3,7 @@
 * Proposal: [SE-0414](0414-region-based-isolation.md)
 * Authors: [Michael Gottesman](https://github.com/gottesmm) [Joshua Turcotti](https://github.com/jturcotti)
 * Review Manager: [Holly Borla](https://github.com/hborla)
-* Status: **Accepted**
+* Status: **Implemented (Swift 6.0)**
 * Implementation: On `main` gated behind `-enable-experimental-feature RegionBasedIsolation`
```
slavapestov, line 7: Do you want to remove the mention of this flag too?

hborla: I'll remove the Implementation field, but the flag does need to be listed as an upcoming feature instead of an experimental feature. The [acceptance revision](https://forums.swift.org/t/accepted-with-modifications-se-0414-region-based-isolation/70051) means this proposal does have source compatibility impact, so that section needs to be updated too. cc @gottesmm

PR #2500, opened 2024-06-24
feature flag for 0192-non-exhaustive-enums.md

mattmassicotte:
Refereced here:

https://github.com/swiftlang/swift/blob/54898ce105c71855d01b270282757a2f09d8edf9/include/swift/Basic/Features.def#L218


mattmassicotte:
Actually has already been proposed here: https://github.com/swiftlang/swift-evolution/pull/2450

PR #2504, opened 2024-06-26
[SE-0417] Revise Task(executorPreference:) ownership semantics

ktoso:
This is just a cleanup of a semantics section which previously showed what was unsafe, but actually isn't unsafe anymore since we fixed it.

This should not require a re-review, as it does not change API surface in any way, it's more of a bugfix really, and making the proposal match reality now.

1. https://github.com/swiftlang/swift-evolution/commit/5a8ec8cde7a172815a32ab78098d2e2e38ded48e revises the ownership semantics of Task initializers. They were previously documented to NOT own the task executors and pushed the responsibility onto developers to keep executors alive (similar to serial executors). 
    - We can do better and have implemented initializers taking ownership of task executors passed to unstructured task initializers: https://github.com/swiftlang/swift/pull/74174

<details>
<summary>Removed some changes since not necessary anymore</summary>

2. ~~https://github.com/swiftlang/swift-evolution/commit/3ad8d51ecc6e21527d7d1902699d3f7a555e91f3 is a necessary revision in order to uphold default actor isolation safety when a task executor is used. Without this change task executors could violate actor isolation and even run concurrent code on a single actor. We must address this by forwarding the serial executor to the task execution such that it can do the usual actor locking.~~
    - ~~The implementation for this change is here: https://github.com/swiftlang/swift/pull/74658/  and we need to decide how aggressively we can phase out the incorrect APIs.~~
    - This was fixed without changes to the API 👍 

</details>


proposals/0417-task-executor-preference.md
```diff
@@ -422,37 +452,23 @@ and wrapping the code that is required to run on a specific executor in an `with
...
-Another use case may be carrying the same task executor into an un-structured Task -- although this should only be done with **extreme caution**,
-because it breaks structured concurrency lifetime expectations of executors. For example, the following code is correct under structured concurrency's
-default and automatic behavior surrounding task executors:
+### TaskExecutor ownership
 
-```swift
-func computeThings() async {
-  let eventLoop = MyCoolEventLoop()
-  defer { eventLoop.shutdown() }
+Task executors, unlike serial executors, are explicitly owned by tasks as long as they are running on given task executor.
```
FranzBusch, line 457:
```suggestion
Task executors, unlike serial executors, are explicitly owned by tasks as long as they are running on the given task executor.
```

PR #2510, opened 2024-07-02
Add an ABI Stability vision document.

al45tair:
This talks about ABI stability and where we'd like to go with it for Swift.

### DIFF
--- a/visions/abi-stability.md
+++ b/visions/abi-stability.md
@@ -1,0 +1,248 @@
+# A Vision for ABI Stability on non-Apple platforms
+
+## Introduction
+
+ABI or _Application Binary Interface_ refers to the details of how an
+application will interact with some other piece of code, including:
+
+* Calling conventions
+* Layout of data structures
+* Representations of scalar types
+* Availability of APIs
+
+If a piece of code has a _stable ABI_, that means that application
+programs can rely on there being no incompatible changes between the
+state of the ABI when they were being compiled and the ABI provided at
+runtime by the component they were linked with.
+
+End users of consumer operating systems, including both Apple and
+Microsoft platforms, tend to take ABI stability for granted; that is,
+they expect that they can install software onto their machine and that
+operating system updates will not arbitrarily break that software.  It
+would be easy to fool yourself into thinking, therefore, that ABI
+stability was the default, or that it was easy, whereas in practice
+considerable time and effort goes into making sure that things stay
+this way.
+
+A related concept is that of the _ABI boundary_, which is the level at
+which a given component provides a stable ABI.  For instance, on Apple
+and Microsoft operating systems, the stable ABI is provided by system
+libraries; while it is possible to dip below this level, for instance
+by doing direct system calls, no guarantee is made that the system
+call interface is ABI stable on those platforms---it is not part of
+the ABI boundary.  Contrast that with Linux, where the system call
+interface _is_ an ABI boundary, while some widely used libraries may
+not themselves be ABI stable _at all_.
+
+## What Does ABI Stability Buy Us?
+
+Put simply, it allows for shared binary components that get updated
+separately.  This is typically a requirement for consumer operating
+systems and software developed for them, so that software purchased by
+consumers can be guaranteed to operate across a wide range of
+operating system versions.
+
+These shared components might be system software or language runtimes
+on the one hand, but ABI stability also enables other patterns like
+binary plug-ins, which need a stable ABI from their host application.
+
+## Swift and ABI Stability
+
+At time of writing, Swift is officially ABI stable only on Apple
+platforms, where it forms part of the operating system, and indeed
+many operating system components are written in Swift.
+
+Apple's basic approach to ABI stability on its platforms relies on
+being able to tell the compiler the minimum system version on which
+the code it is building should run, coupled with the ability to
+annotate API with `@availability` annotations describing which system
+versions support which APIs.  The availability mechanism also allows
+the compiler to decide automatically whether to weak link the
+symbol(s) in question.
+
+That does not completely solve the problem, however, because the
+platform needs to be able to evolve the data types it exposes to
+applications, while allowing applications to subclass Swift or
+Objective-C types declared in the platform API.
+
+You may be familiar with the C++ "fragile base class" problem, which
+is essentially that a subclass of a C++ class needs to know the layout
+(or at least the size) of the base class, which implies that the
+latter cannot change.  In C++ the traditional solution to that is the
+_`pimpl` idiom_ (essentially, you make the base class have a single
+pointer member that points to its actual data).
+
+Both Swift and its predecessor language Objective-C solve this problem
+in a different way, namely for ABI stable structs and classes (all
+classes in Objective-C, but in Swift only `public` non-`frozen` types
+compiled with _library evolution_ enabled), the _runtime_ is in charge
+of data layout, and the compiler, when asked to access a specific data
+member, will make runtime calls as necessary to establish where it
+should read or write data.  This avoids the extra indirection of the
+C++ `pimpl` idiom and in practice it's possible to generate reasonably
+efficient code.
+
+## Platforms and ABI Stability
+
+### Darwin
+
+In addition to the `@availability`/minimum system version and
+non-fragile base class support, the Darwin linker has a number of
+features intended to allow code to be moved from one framework or
+dynamic library to another without an ABI break, including support for
+re-exporting symbols, as well as a variety of meta symbols that allow
+for manipulation of symbol visibility and indeed apparent location
+based on system versions.
+
+### Windows
+
+Like Apple platforms, Windows is ABI stable at the API layer; SPIs are
+not officially ABI stable (though in practice they may be), and
+neither is the system call interface itself.
+
+Windows has a platform-wide exception handling mechanism (Structured
+Exception Handling or SEH), which means that the operating system
+specifies in detail how stack unwinding must take place.  Unlike other
+platforms we care about, Windows does not use DWARF for exception
+unwinding, but instead relies on knowledge of the standard function
+prologue and epilogue.  This creates problems for tail call
+optimization, which is also needed for Swift Concurrency.  While we
+have *a* solution in place, we need to be convinced that it's the
+*right* solution before declaring ABI stability on Windows.
+
+Windows DLLs also have some interesting design features that need to
+be borne in mind here.  In particular, symbols exported from DLLs can
+be exported by 16-bit _ordinal_ as well as by name; name exports are
+actually done by mapping the name to an ordinal, and then looking up
+the ordinal in the export table.  If symbols are exported by ordinal
+alone, the ordinals need to be stable in order to achieve ABI
+stability.  Further, because of the way linking works, people have in
+the past generated import libraries directly from DLL files, which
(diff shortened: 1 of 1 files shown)


visions/abi-stability.md
```diff
@@ -0,0 +1,248 @@
...
+* Representations of scalar types
+* Availability of APIs
+
+If a piece of code has a _stable ABI_, that means that application
+programs can rely on there being no incompatible changes between the
+state of the ABI when they were being compiled and the ABI provided at
+runtime by the component they were linked with.
+
+End users of consumer operating systems, including both Apple and
+Microsoft platforms, tend to take ABI stability for granted; that is,
```
etcwilde, line 19: Is it as true for Microsoft/Windows? I know they're moving toward a "universal C runtime" so this may not be as true in the future (though certainly still true for the C++ runtime), but you still end up bundling the language runtime DLLs in the application bundle next to the executable so it's less ABI stability and more, "you pack the runtime you built against so there aren't ABI conflicts"? Sort of like snaps or how we used to ship the Swift standard library on Apple platforms before we had ABI stability.

visions/abi-stability.md
```diff
@@ -0,0 +1,248 @@
...
+    necessary.
+
+  - `@availability` as a system-wide concept doesn't make as much
+    sense on non-Apple UNIX/UNIX-like platforms where it's possible to
+    install different versions of many standard packages.  This is
+    especially true on Linux where there are multiple "distributions",
+    so the major versions of packages in major versions of those
+    distributions are fixed, but one distribution's set of packages
+    may bear little relation to another's.  In this case,
+    `@availability` annotations could perhaps be used for the system C
```
etcwilde, line 212: It would be great to have an actual Swift-version availability annotation that doesn't map back to an OS version.

visions/abi-stability.md
al45tair: :-) It's complicated.

The C runtime on Windows isn't tightly coupled with the OS in the way that the C library on Darwin or Linux is; system components don't use the same C runtime library that applications use, and it's totally possible to write programs that just use the Win32 API and don't use the C runtime at all (or very little, at any rate).

So yes, programs will often install a C runtime library somewhere (hopefully, these days, not in `C:\Windows\System32`) and not share that with other programs or the rest of the system, but the underlying system API is ABI stable.

visions/abi-stability.md
al45tair: Agreed.

visions/abi-stability.md
al45tair: (This does have some interesting implications for Swift C++ interop on Windows, in that it means that user code would need to be using the same C/C++ runtime as the Swift runtime. Similarly if anything uses Swift runtime functions that return `malloc`-ed memory and so on.)

visions/abi-stability.md
```diff
@@ -0,0 +1,248 @@
...
+Model (COM), which relies on the way the Microsoft compiler lays out
+C++ vtables.
+
+Windows doesn't have anything quite like the `availability` mechanism
+Apple platforms use, though there are C preprocessor macros like
+`WINVER` or `_WIN32_WINNT` that control exactly which versions of data
+structures and which functions are exported by the Windows SDK headers.
+
+Windows programs that need to cope with the possibility that a function
+might not be present at runtime need to use `GetProcAddress()` (the
```
compnerd, line 144:
```suggestion
might not be present at runtime need to use `GetProcAddress` (the
```

The parenthesis makes it feel like there are no arguments to the invocation.

visions/abi-stability.md
```diff
@@ -0,0 +1,248 @@
...
+Win32 equivalent of `dlsym()`).
```
compnerd, line 145:
```suggestion
Win32 equivalent of `dlsym`).
```

The parenthesis makes it feel like there are no arguments to the invocation.

visions/abi-stability.md
al45tair: The parentheses just indicate that we're talking about a function. I don't think it's an unusual style, marking function names with parentheses like that (pretty sure I've seen documentation written like that all over the place).

visions/abi-stability.md
al45tair: See previous remark.

visions/abi-stability.md
```diff
@@ -191,6 +198,42 @@ package builds take this approach, but have to restrict themselves as
...
 
+## Declaring ABI Stability
+
+Supported platforms for Swift must have a Platform Owner.  Because a
+declaration of ABI stability for a platform has implications for
+compiler and runtime engineers who do not develop on that platform, we
+intend to set a high bar for declaring ABI stability.  As such, to
+become officially ABI stable:
+
+1. The Platform Owner will raise a Swift Evolution proposal for
```
etcwilde, line 209: What is a "Platform"? Is "Linux" a platform, or are "Debian", "Fedora", and "Arch Linux" separate platforms? At least between Debian and Arch, there is a considerable difference in their release strategies that impact the ABI stories, ranging from, "libraries should be held stable for one or two releases" to "rolling-release. ABI is whatever you happened to have downloaded at the moment. Partial-upgrades are not supportable".

al45tair: I think Linux probably _is_ a platform, but I appreciate it's a complicated case and will require some special handling.

etcwilde: I'm 100% on-board with nailing down the calling conventions, but still have questions about the library ABI/binary compatibility story.

Will we be requiring full bi-directional binary compatibility like on Darwin where you can build against a new runtime and run it on older runtimes (assuming minimum deployment targets work out) as well as building against older runtimes and running it on newer runtimes?
Or are we okay with something more like glibc/libstdc++ where something built against the older runtime will continue working on newer runtimes, but not the other way around?

al45tair: Personally I think we should aim high here; if we can make Swift itself work the way it does on Darwin, that's actually quite a compelling feature for Linux development — Swift programs won't have the deployment problems that C/C++ programs do.

I realise that's not easy and will require `@availability` support and some kind of Swift minimum target version support from the compiler too.

visions/abi-stability.md
```diff
@@ -0,0 +1,304 @@
...
+pointer member that points to its actual data).
+
+Both Swift and its predecessor language Objective-C solve this problem
+in a different way, namely for ABI stable structs and classes (all
+classes in Objective-C, but in Swift only `public` non-`frozen` types
+compiled with _library evolution_ enabled), the _runtime_ is in charge
+of data layout, and the compiler, when asked to access a specific data
+member, will make runtime calls as necessary to establish where it
+should read or write data.  This avoids the extra indirection of the
+C++ `pimpl` idiom and in practice it's possible to generate reasonably
```
rokhinip, line 82: It seems more like it's a different type of indirection since you're making a function call instead of doing a load off another load instead of avoiding it altogether.

visions/abi-stability.md
```diff
@@ -0,0 +1,304 @@
...
+## Swift and ABI Stability
+
+At time of writing, Swift is officially ABI stable only on Apple
+platforms, where it forms part of the operating system, and indeed
+many operating system components are written in Swift.
+
+Apple's basic approach to ABI stability on its platforms relies on
+being able to tell the compiler the minimum system version on which
+the code it is building should run, coupled with the ability to
+annotate API with `@availability` annotations describing which system
```
tshortli, line 58:
```suggestion
annotate API with `@available` annotations describing which system
```

visions/abi-stability.md
```diff
@@ -0,0 +1,304 @@
...
+member, will make runtime calls as necessary to establish where it
+should read or write data.  This avoids the extra indirection of the
+C++ `pimpl` idiom and in practice it's possible to generate reasonably
+efficient code.
+
+## Platforms and ABI Stability
+
+### Darwin
+
+In addition to the `@availability`/minimum system version and
```
tshortli, line 89:
```suggestion
In addition to the `@available`/minimum system version and
```

visions/abi-stability.md
```diff
@@ -0,0 +1,304 @@
...
+   been met, the platform will be declared ABI Stable and all future
+   changes to the compiler, runtime or standard library will be
+   expected to maintain ABI stability for that platform.
+
+## Goals
+
+* Swift should support existing platform ABI stability mechanisms,
+  so that the language is a good citizen when writing native software.
+
+  - `@availability` should be well-defined for Windows; it should be
```
tshortli, line 244:
```suggestion
  - `@available` should be well-defined for Windows; it should be
```

visions/abi-stability.md
```diff
@@ -0,0 +1,304 @@
...
+    possible to annotate Windows APIs to say e.g. that they were
+    introduced in Windows 10 version 1709.  Swift should automatically
+    generate the code to weakly reference functions it imports where
+    necessary.
+
+  - `@availability` as a system-wide concept doesn't make as much
```
tshortli, line 250:
```suggestion
  - `@available` as a system-wide concept doesn't make as much
```

visions/abi-stability.md
```diff
@@ -0,0 +1,304 @@
...
+    sense on non-Apple UNIX/UNIX-like platforms where it's possible to
+    install different versions of many standard packages.  This is
+    especially true on Linux where there are multiple "distributions",
+    so the major versions of packages in major versions of those
+    distributions are fixed, but one distribution's set of packages
+    may bear little relation to another's.  In this case,
+    `@availability` annotations could perhaps be used for the system C
```
tshortli, line 257:
```suggestion
    `@available` annotations could perhaps be used for the system C
```

visions/abi-stability.md
```diff
@@ -0,0 +1,304 @@
...
+    that just use the standard library would work as long as a
+    suitable copy of the Swift runtime was present on the system.
+
+* It should be possible to write ABI stable APIs _in_ Swift.
+
+## Non-Goals
+
+At this point, the following are explicitly not goals:
+
+* Providing `@availability`-like support for third party library code.
```
tshortli, line 298:
```suggestion
* Providing `@available`-like support for third party library code.
```

visions/abi-stability.md
```diff
@@ -0,0 +1,304 @@
...
+  This is an interesting idea, but would require a method of
+  specifying the minimum version, which probably means being able to
+  specify version numbers when importing modules, e.g.
+  e.g. `@version(12.3) import Libfoo` or `import Libfoo@12.3` or
+  similar.  It would also complicate the compiler's availability
```
tshortli, line 303: FWIW, this is pitched in https://forums.swift.org/t/pitch-extensible-availability-checking/79308 (and significant progress on implementing the fundamentals has already been made)

visions/abi-stability.md
```diff
@@ -0,0 +1,304 @@
...
+
+## Non-Goals
+
+At this point, the following are explicitly not goals:
+
+* Providing `@availability`-like support for third party library code.
+  This is an interesting idea, but would require a method of
+  specifying the minimum version, which probably means being able to
+  specify version numbers when importing modules, e.g.
+  e.g. `@version(12.3) import Libfoo` or `import Libfoo@12.3` or
```
tshortli, line 302: This is a possible direction, but I think it would be better to specify the "deployment target" for a dependency on the command line, since repeating it with potentially conflicting values on multiple import statements for the same module could be messy to handle

PR #2511, opened 2024-07-02
[SE-0257]: Add link to pitch and review threads, unify metadata

ole:
This adds the links for the pitch and review forum threads to SE-0257. It also formats the `Status` metadata field according to the current format.

I chose to retain the label "previous pitch" for the links to the earlier pitch threads (instead of, say, "pitch 1", "pitch 2", "pitch 3") because (a) the first previous pitch was years earlier and (b) the second previous pitch was for a different feature (allow trailing commas), which then turned into the real pitch for eliding commas a few months later (I think).


rjmccall:
I'm happy enough with how you've done the labels here.

PR #2515, opened 2024-07-12
[SE-0438] Metatype Keypaths proposal is accepted

benrimmington:
Update status field. Add review and acceptance links.


benrimmington:
@jckarter (review manager)

PR #2529, opened 2024-07-24
Add proposal for useful String.Index descriptions

lorentey:
This retroactively conforms String.Index to CustomStringConvertible.

rdar://22794467

### DIFF
--- a/proposals/NNNN-string-index-printing.md
+++ b/proposals/NNNN-string-index-printing.md
@@ -43,9 +43,13 @@ print(string.firstRange(of: "ell")!) // ⟹ 10[utf8]..<13[utf8]
 
 The sample description strings shown above are illustrative, not normative. This proposal does not specify the exact format and information content of the string returned by the `description` implementation on `String.Index`. As is the case with most conformances to `CustomStringConvertible`, the purpose of these descriptions is to expose internal implementation details for debugging purposes. As those implementation details evolve, the descriptions may need to be changed to match them. Such changes are not generally expected to be part of the Swift Evolution process; so we need to keep the content of these descriptions unspecified.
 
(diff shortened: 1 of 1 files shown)


benrimmington:
You could mention that LLDB uses the same format (since Swift 5.8):

* [SwiftFormatters.cpp](https://github.com/swiftlang/llvm-project/blob/swift-5.8-RELEASE/lldb/source/Plugins/Language/Swift/SwiftFormatters.cpp#L463-L481)

* [SwiftStringIndex.h](https://github.com/swiftlang/llvm-project/blob/swift-5.8-RELEASE/lldb/source/Plugins/Language/Swift/SwiftStringIndex.h)

---

Was there any discussion about using `debugDescription` and `CustomDebugStringConvertible` instead? Or do you want to reserve those for a more verbose format?

lorentey:
> Was there any discussion about using debugDescription and CustomDebugStringConvertible instead? Or do you want to reserve those for a more verbose format?

Yes! After pitching this stuff, I investigated the role of the `CustomDebugStringConvertible` protocol, and I've had a long-standing reminder to start a forum discussion clarifying it. Surprisingly, it turns out it `debugDescription` isn't actually there to produce a more verbose display, or to give more information about the item. (For all I know, it may have been designed for that purpose, but that is certainly not how Swift is using it.)

`CustomDebugStringConvertible` is somewhat of a misnomer -- it actual de facto purpose is to produce a display that is functionally equivalent to `description`, but in a form that works well in structured, syntactic output: i.e., the string it returns needs to be free of unpaired parenthesis/brackets/quotes/etc, naked top-level commas/spaces, etc.

Collection types such as `Array` or `Dictionary` use `debugDescription` to print their contents, to avoid unintelligible output. Therefore, `debugDescription` must not be overly verbose -- it just needs to produce a syntactically well-formed format. (It will appear in array/dictionary/etc descriptions, where brevity is important.)

PR #2536, opened 2024-08-02
[SE-0427] Noncopyable Generics is accepted

benrimmington:
Update status field. Add review links.

Cc: @hborla @airspeedswift (review managers)


hborla (approved):
Thank you!

PR #2538, opened 2024-08-05
Avoid confusing acronym in SE-0371

ktoso:
GAIT is not a term used by the Swift language, and we should just spell it out that we mean global actor annotated types explicitly.

### DIFF
--- a/proposals/0371-isolated-synchronous-deinit.md
+++ b/proposals/0371-isolated-synchronous-deinit.md
@@ -13,7 +13,7 @@ This feature allows `deinit`'s of actors and global-actor isolated classes to ac
 
 ## Motivation
 
-Restrictions imposed by [SE-0327](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0327-actor-initializers.md) reduce the usefulness of explicit `deinit`s in actors and GAITs. Workarounds for these limitations may involve creation of `close()`-like methods, or even manual reference counting if the API should be able to serve several clients.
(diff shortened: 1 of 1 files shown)


xwu:
I believe we've been calling them "global actor isolated types" in other proposals—probably best to be consistent here?

Jumhyn (approved):

proposals/0371-isolated-synchronous-deinit.md
```diff
@@ -13,7 +13,7 @@ This feature allows `deinit`'s of actors and global-actor isolated classes to ac
 
 ## Motivation
 
-Restrictions imposed by [SE-0327](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0327-actor-initializers.md) reduce the usefulness of explicit `deinit`s in actors and GAITs. Workarounds for these limitations may involve creation of `close()`-like methods, or even manual reference counting if the API should be able to serve several clients.
+Restrictions imposed by [SE-0327](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0327-actor-initializers.md) reduce the usefulness of explicit `deinit`s in actors and global actor annotated types. Workarounds for these limitations may involve creation of `close()`-like methods, or even manual reference counting if the API should be able to serve several clients.
```
Jumhyn, line 16:
```suggestion
Restrictions imposed by [SE-0327](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0327-actor-initializers.md) reduce the usefulness of explicit `deinit`s in actors and global actor isolated types. Workarounds for these limitations may involve creation of `close()`-like methods, or even manual reference counting if the API should be able to serve several clients.
```

PR #2554, opened 2024-08-26
[SE-0443] Exclude -Wsuppress and -suppress-warnings from the proposal

DmT021:


DmT021:
@rjmccall We've decided to make an amendment to the proposal that removes support for `-Wsuppress` and `-suppress-warnings`.

PR #2558, opened 2024-09-05
[SE-0371] Isolated synchronous deinit is accepted

benrimmington:
Update status and review fields.

Cc: @jumhyn, @nickolas-pohilets


ktoso:
LGTM though I'm not sure we need to do it via the "amended" wording in the text -- we should just change the text to reflect the accepted revision I think?

And add the change to the changelog at the bottom.

cc @rjmccall @hborla

benrimmington:
I've used ~~strikethrough~~ on the sections that need to be rewritten, preferably by the proposal's author.

benrimmington:
I've changed this pull request so that it only updates the status and review fields.

Naveen-C-Ramachandrappa (approved):

PR #2559, opened 2024-09-07
Introduce SuppressedAssociatedTypes as a standalone feature.

kavon:


kavon:
This effort has moved to a new proposal: https://github.com/swiftlang/swift-evolution/pull/3047

PR #2572, opened 2024-09-22
Add a proposal to change the execution semantics of nonisolated async functions.

hborla:


proposals/NNNN-async-function-isolation.md
```diff
@@ -0,0 +1,897 @@
...
+
+```swift
+class NotSendable { ... }
+
+func useAsValue(_ ns: NotSendable) async { ... }
+
+@MainActor let global: NotSendable = .init()
+
+@MainActor
+func callSendableClosure(closure: @Sendable (NotSendable) async -> Void) {
```
mattmassicotte, line 247: I *think* this function signature should be changed to:

`func callSendableClosure(_ closure: @Sendable (NotSendable) async -> Void) async`

proposals/NNNN-async-function-isolation.md
```diff
@@ -0,0 +1,897 @@
...
+It's tempting to not introduce a new attribute to control where an async
+function executes, and instead control this behavior with an explicit
+`nonisolated` annotation. However, this approach falls short for the following
+reasons:
+
+1. It does not accomplish the goal of having consistent semantics for
+   `nonisolated` regardless of whether it's applied to synchronous or
+   async functions.
+2. It's important to have an explicit, easy-to-write spelling for async
+   functions that run on the caller's actor. For example, this is useful to
```
jamieQ, line 883: should this say 'run off' or 'do not run on'?

proposals/NNNN-async-function-isolation.md
hborla: Indeed, fixed!

hborla:
@xwu (cc @DougGregor) as review manager: I revised the proposal based on feedback from the pitch and vision document discussion in https://github.com/swiftlang/swift-evolution/pull/2572/commits/2480b835a9793f8d687595d930077655f7173fff. I made a ton of editorial changes, but the substantive revisions that I made are explicitly listed in the new [Revisions section](https://github.com/swiftlang/swift-evolution/blob/2480b835a9793f8d687595d930077655f7173fff/proposals/NNNN-async-function-isolation.md#revisions):

> ## Revisions
> 
> The proposal was revised with the following changes after the pitch discussion:
> 
> * Gate the behavior change behind an `AsyncCallerExecution` upcoming feature
>   flag.
> * Change the spelling of `@concurrent` to `@execution(concurrent)`, and add an
>   `@execution(caller)` attribute to allow expressing the new behavior this
>   proposal introduces when the upcoming feature flag is not enabled.
> * Apply `@execution(caller)` to nonisolated async function types by default to
>   make the execution semantics consistent between async function declarations
>   and values.
> * Change the terminology in the proposal to not use the "inherits isolation"
>   phrase.

There is also now an implementation of this proposal on `main` under an experimental feature.

proposals/NNNN-async-function-isolation.md
hborla: No, but I agree that the way I worded this is confusing

hborla:
@xwu I could continue with editorial passes on this proposal all day, but I think I've addressed the feedback from the LSG

PR #2576, opened 2024-09-25
[SE-0448] Fix link to lookbehind assertions pitch

benrimmington:


benrimmington:
Cc: @milseman, @stephentyrone

PR #2581, opened 2024-10-01
Prospective vision: Optional Strict Memory Safety for Swift

DougGregor:


visions/memory-safety.md
```diff
@@ -0,0 +1,165 @@
...
+
+* `@unchecked Sendable`
+* `unowned(unsafe)`
+* `nonisolated(unsafe)`
+* `unsafeAddressor`, `unsafeMutableAddressor`
+
+In addition, an `@unsafe` attribute would be added to the language and would be used to mark any declaration that is unsafe to use. In the standard library, the following functions and types would be marked `@unsafe` :
+
+* `Unsafe(Mutable)(Raw)(Buffer)Pointer`
+* `(Closed)Range.init(uncheckedBounds:)`
```
apple-fcloutier, line 53: Does this lead to memory unsafety?

visions/memory-safety.md
```diff
@@ -0,0 +1,165 @@
...
+
+Providing memory safety does not imply the absence of run-time failures. Good language design often means defining away runtime failures in the type system. However, memory safely requires only that an error in the program cannot be escalated into a violation of one of the safety properties. For example, having reference types by non-nullable by default defines away most problems with NULL pointers. With explicit optional types, the force-unwrap operator (postfix `!` ) meets the definition of memory safety by trapping at runtime if the unwrapped optional is `nil` . The standard library also provides the [`unsafelyUnwrapped` property](https://developer.apple.com/documentation/swift/optional/unsafelyunwrapped) that does not check for `nil` in release builds: this does not meet the definition of memory safety because it admits violations of initialization and lifetime safety that could be exploited.
```
apple-fcloutier, line 21: having reference types ~~by~~be non-nullable

visions/memory-safety.md
```diff
@@ -0,0 +1,165 @@
...
+func memcpy(
+  _: UnsafeMutableRawPointer?,
+  _: UnsafeRawPointer?,
+  _: Int
+) -> UnsafeMutableRawPointer?
+```
+
+The rules described above make it possible to detect and report the use of unsafe constructs in Swift.
+
+An `@unsafe` function should be allowed to use other unsafe constructs without emitting any diagnostics. However, there are also library functions that encapsulate unsafe behavior in a safe API, such as the standard library’s `Array` and [`Span`](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0447-span-access-shared-contiguous-storage.md)that are necessarily built from unsafe primitives. Such functions need some way to acknowledge the unsafe behavior while still being considered safe from the outside, such as an `unsafe { ... }` code block or a `@safe(unchecked)` attribute.
```
apple-fcloutier, line 77: Missing space between span link and "that"

visions/memory-safety.md
```diff
@@ -0,0 +1,165 @@
...
+
+* **Lifetime safety** : all accesses to a value are guaranteed to occur during its lifetime. Violations of this property, such as accessing a value after its lifetime has ended, are often called use-after-free errors.
+* **Bounds safety**: all accesses to memory are within the intended bounds of the memory allocation, such as accessing elements in an array. Violations of this property are called out-of-bounds accesses.
+* **Type safety** : all accesses to a value use the type to which it was initialized, or a type that is compatible with that type. For example, one cannot access a `String` value as if it were an `Array`. Violations of this property are called type confusions.
+* **Initialization safety** : all values are initialized property to being used, so they cannot contain unexpected data. Violations of this property often lead to information disclosures (where data that should be invisible becomes available) or even other memory-safety issues like use-after-frees or type confusions.
```
Joannis, line 16: "Initialized property" is this meant to be "properly"?

visions/memory-safety.md
DougGregor: Yes, because bounds checking for various collection operations taking an range as input assume that the invariants hold for the range.

visions/memory-safety.md
```diff
@@ -0,0 +1,253 @@
...
```
rolson, line 16:
```suggestion
* **Initialization safety** : all values are initialized prior to being used, so they cannot contain unexpected data. Violations of this property often lead to information disclosures (where data that should be invisible becomes available) or even other memory-safety issues like use-after-frees or type confusions.
```

visions/memory-safety.md
```diff
@@ -0,0 +1,253 @@
...
+
+The C family of languages do not provide memory safety along any of the dimensions described in this document. As such, a Swift program that makes use of C APIs is never fully “memory safe” in the strict sense, because any C code called from Swift could undermine the memory safety guarantees Swift is trying to provide. Requiring that all such C code be rewritten in Swift would go against Swift’s general philosophy of incremental adoption into existing ecosystems. Therefore, this document proposes a different strategy: code written in Swift will be auditably memory-safe so long as the C APIs it uses follow reasonable conventions with respect to memory safety. As such, writing new code (or incrementally rewriting code from the C family) will not introduce new memory safety bugs, so that adopting Swift in an existing code base will incrementally improve on memory safety. This approach is complementary to any improvements made to memory safety within the C family of languages, such as [bounds-safety checks for C](https://clang.llvm.org/docs/BoundsSafety.html) or [C++ standard library hardening](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p3471r0.html).
```
geoffreygaren, line 146: > so long as the C APIs it uses follow reasonable conventions with respect to memory safety

Instead of "reasonable conventions", it might be more precise to say "declarable conventions" and also to specify "that are consistent with Swift's approach to memory safety". There exist some C conventions that are arguably reasonable, but also
* impractical to declare (e.g. printf); and/or
* inconsistent with Swift's view of the universe (e.g. the Boehm Garbage Collector)
Such conventions will remain @unsafe in this proposal, not because they are intrinsically unsafe, but rather because we are unable to represent their notion of safety in a way that Swift can verify.

> a Swift program that makes use of C APIs is never fully “memory safe” in the strict sense, because any C code called from Swift could undermine the memory safety guarantees Swift is trying to provide.

I believe the same is true of a Swift program that makes use of Swift APIs, since Swift APIs may internally use unsafe features that undermine memory safety, whether intentionally or accidentally. The same goes for a Swift program that makes use of an OS or a piece of hardware. 

I would suggest that the important distinction is something like this: "Though Swift is memory safe by default, a Swift program that makes use of C APIs is not. C APIs do not describe their memory safety invariants, so Swift today must suspend memory safety enforcement when invoking C APIs."

visions/memory-safety.md
```diff
@@ -0,0 +1,253 @@
...
+
+This vision proposes that the strict safety checking described be an opt-in feature with no path toward becoming the default behavior in some future language mode. There are several reasons why this checking should remain an opt-in feature for the foreseeable future:
+
+* The various `Unsafe` pointer types are the only way to work with contiguous memory in Swift today, and  the safe replacements (e.g., `Span`) are new constructs that will take a long time to propagate through the ecosystem. Some APIs depending on these `Unsafe` pointer types cannot be replaced because it would break existing clients (either source, binary, or both).
+* Interoperability with the C family of languages is an important feature for Swift. Most C(++) APIs are unlikely to ever adopt the safety-related attributes described above, which means that enabling strict safety checking by default would undermine the usability of C(++) interoperability.
+* Swift's current (non-strict) memory safety by default is likely to be good enough for the vast majority of users of Swift, so the benefit of enabling stricter checking by default is unlikely to be worth the disruption it would cause.
```
geoffreygaren, line 252: Another way to make this point, without predicting that safety by default will prove "good enough" in the end, is: "Many Swift projects contain very little unsafe code, but not zero. A new default could break the build in many projects, causing significant disruption, in exchange for very little net reduction in unsafe code (or none at all, if haggard engineers simply turn of the new default or hastily add @safe(checked) blocks)."

rjmccall:
[Accepted by the LSG](https://forums.swift.org/t/accepted-vision-optional-strict-memory-safety-for-swift/77223).

PR #2590, opened 2024-10-16
A Vision for WebAssembly Support in Swift

MaxDesiatov:


visions/webassembly.md
```diff
@@ -0,0 +1,196 @@
...
+
+2. Improve support for cross-compilation in Swift and SwiftPM. We can simplify versioning, installation, and overall
+management of Swift SDKs for cross-compilation in general, which is beneficial not only for WebAssembly, but for all
+platforms.
+
+3. Continue work on Wasm Component Model support in Swift as the Component Model proposal is stabilized. Ensure
+that future versions of WASI are available to Swift developers targeting Wasm.
+
+4. Make interoperability with Wasm components as smooth as C and C++ interop already is for Swift. With a formal
+specification for Canonical ABI progressing, this will become more achievable with time.
```
endocrimes, line 103:
```suggestion
4. Make interoperability with Wasm components as smooth as C and C++ interop already is for Swift. With a formal
specification for Canonical ABI progressing, this will become more achievable with time. This includes consuming components from, and building components with Swift.
```

Some languages early implementations are good at one but not the other (Golang's current WASI Support for example, supports _importing_ from components, but not yet exporting functions that can be used from other components or hosts - due to complications in the runtime) - we should be explicit that both are things we would _like_ to see in Swift as we advance component support.

visions/webassembly.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Goals
+
+As of March 2024 all patches necessary for basic Wasm and WASI support have been merged to the Swift toolchain and
+core libraries. Based on this, we propose a high-level roadmap for WebAssembly support and adoption in the Swift
+ecosystem:
+
+1. Make it easier to evaluate and adopt Wasm with increased API coverage for this platform in the Swift core libraries.
+As a virtualized embeddable platform, not all system APIs are always available or easy to port to WASI. For example,
+multi-threading, file system access, and localization need special support in Wasm runtimes and a certain amount of
```
endocrimes, line 92:
```suggestion
multi-threading, file system access, networking and localization need special support in Wasm runtimes and a certain amount of
```

visions/webassembly.md
```diff
@@ -0,0 +1,196 @@
...
+environment to provide multi-threading support. This means that multi-threading in Wasm is dependent on the Wasm runtime
+that executes a module. There are two proposals to standardize ways to create threads in Wasm:
+
+(1) [wasi-threads](https://github.com/WebAssembly/wasi-threads), which is already supported by some toolchains,
+runtimes, and libraries but has been superseded;
+
+(2) The new [shared-everything-threads](https://github.com/WebAssembly/shared-everything-threads) proposal is still
+in the early stages, but is expected to be the future of multi-threading in Wasm.
+
+Swift currently supports two threading models in Wasm: single-threaded (`wasm32-unknown-wasi`) and multi-threaded
```
endocrimes, line 163: aside: should we consider changing the wasm triple to match other languages? (specifically rust's changes here would also make sense for swift https://blog.rust-lang.org/2024/04/09/updates-to-rusts-wasi-targets.html)

MaxDesiatov: This was partially done by @kateinoigakukun, at least for the threads triple.

kateinoigakukun: Agree. "Officialization" would be a good opportunity to rename the target name.

visions/webassembly.md
```diff
@@ -0,0 +1,196 @@
...
+WebAssembly currently uses a 32-bit address space, but [64-bit address space](https://github.com/WebAssembly/memory64/)
+proposal is already in the implementation phase.
+
+Swift supports 64-bit pointers on other platforms where available, however WebAssembly is the first platform where
+relative reference from data to code is not allowed. Alternative solutions like image-base relative addressing or
+"small code model" for fitting 64-bit pointer in 32-bit are unavailable, at least for now. This means that we need
+cooperation from the WebAssembly toolchain side or different memory layout in Swift metadata to support 64-bit linear
+memory support in WebAssembly.
+
+### Shared libraries
```
tschneidereit, line 180: A few of us recently hashed out [a design](https://github.com/WebAssembly/component-model/issues/401) for supporting a more dynamic approach to shared libraries in the component model. Based on that sketch, I think it's reasonable to assume that we'll have something similarly expressive as traditional `dlopen` for shared libraries in the component model.

PR #2598, opened 2024-10-23
Mark SE-0449 as implemented in Swift 6.1.

simanerush:
Marks SE-0449 as implemented in Swift 6.1.

### DIFF
--- a/proposals/0449-nonisolated-for-global-actor-cutoff.md
+++ b/proposals/0449-nonisolated-for-global-actor-cutoff.md
@@ -3,8 +3,7 @@
 * Proposal: [SE-0449](0449-nonisolated-for-global-actor-cutoff.md)
 * Authors: [Sima Nerush](https://github.com/simanerush), [Holly Borla](https://github.com/hborla)
 * Review Manager: [Tony Allevato](https://github.com/allevato)
-* Status: **Accepted**
-* Implementation: On `main` gated behind `-enable-experimental-feature GlobalActorInferenceCutoff`
+* Status: **Implemented (Swift 6.1)**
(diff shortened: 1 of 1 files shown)


ktoso:
It’s not picked to 6.0 right? so I think this may need to wait for upcoming release from main eventually once that happens.

or don’t say which version it’s on?

benrimmington:
[Swift 6.1](https://www.swift.org/swift-evolution/#?version=6.1) has been [announced](https://forums.swift.org/t/swift-6-1-release-process/75442).

simanerush:
Fixed, thanks!

benrimmington:
Should the [upcoming feature](https://github.com/swiftlang/swift/blob/549262a57dcc430cb56a76fc60474580e7765b2e/include/swift/Basic/Features.def#L222) be moved to the next language mode?

```diff
-UPCOMING_FEATURE(GlobalActorInferenceCutoff, 0449, 6)

 // Swift 7
 UPCOMING_FEATURE(ExistentialAny, 335, 7)
 UPCOMING_FEATURE(InternalImportsByDefault, 409, 7)
 UPCOMING_FEATURE(MemberImportVisibility, 444, 7)
+UPCOMING_FEATURE(GlobalActorInferenceCutoff, 449, 7)
```

(Also, the `SENumber` of a few upcoming features is in *octal*, but it probably doesn't matter.)

xwu:
If there are no source-breaking changes, there should be no upcoming feature flag at all.

allevato:
It looks like the implementation was updated to make it an [upcoming feature flag](https://github.com/swiftlang/swift/pull/77175) enabled by default in Swift 6 language mode. Should it be updated to remove the flag? If so, we can merge this as-is. Otherwise, we should update this to mention the flag, since evolution proposals that add upcoming features should keep those references for documentation purposes.

If nothing in the proposal is source-breaking (I don't recall any of it being such), then the flag should probably just be removed entirely as @xwu suggested, and then it will be on by default in 6.1.

allevato:
Now that https://github.com/swiftlang/swift/pull/77227 has been merged, we can merge this to match. Thanks!

PR #2602, opened 2024-10-29
SE0450: Mention about `swift test`

kateinoigakukun:
`swift test` was missed in the list of commands accepting `--traits`


FranzBusch (approved):
Thanks. Good catch

PR #2613, opened 2024-11-10
[SE-0430] Fix a small example so that the example matches the narrative within.

gottesmm:
Specifically, the narrative is talking about sending the variable 't' not the variable 'ns'.


gottesmm:
@DougGregor @hborla

PR #2616, opened 2024-11-14
A Vision for Accessors in Swift

tbkka:
Explains how the upcoming read/modify and the
eventually-to-be-proposed borrow/mutate accessors complement the existing get/set capability.

### DIFF
--- a/visions/accessors.md
+++ b/visions/accessors.md
@@ -1,0 +1,348 @@
+# A Vision for Accessors in Swift
+
+Swift properties are implemented by providing one or more “accessors” which are able to provide or update the value of a property.  Most Swift developers should be familiar with the `get` and `set` accessors that are used to implement computed properties:
+
+```swift
+struct Foo {
+  var value: Int {
+    get { ... provide a value ... }
+    set { ... accept a value ... }
+  }
+}
+```
+
+In this case, the `get` accessor gets compiled into a special kind of method that returns a value, the `set` accessor gets compiled into a method that accepts a new value as an argument.  Reading or writing such a property is compiled into a call to the corresponding method.  Swift provides resilience for stored properties by automatically generating `get` and `set` accessor methods so that clients can be compiled the same way regardless of whether the property is actually written as a stored or computed property.
+
+> The description of “methods” here explains what happens when you access resilient stored properties that are defined in a separate module.  When accessing stored properties from within the *same* module, Swift typically uses efficient direct property access.
+
+However, `get` and `set` accessors have significant limitations.  For example, because the `get` and `set` accessors transfer the value as a method argument or return value, they generally require the value to be copyable, which is no longer universally true in Swift.  This copying also adds significant overhead for large or complex values.  As an additional complication, more complex APIs, such as the subscript access for `Dictionary`, need to construct a temporary value and then clean up that value at the end of the access.
+
+These limitations have led Swift to add a variety of other accessors, including several experimental forms that have never been officially standardized through the Swift Evolution process.  Recent work on noncopyable support and an expanded interest in new kinds of collections has made it important to provide a complete standard set of accessors to support future work.  This document will explain the various accessors that currently exist or might be proposed in the near future and how they will provide the full set of capabilities needed for different applications.
+
+> Note:  Accessors are also used to implement subscripts; the only real difference is that a subscript operation can have additional arguments.  This document will discuss property accesses and subscript operations interchangeably.
+
+## Two Dimensions of Access
+
+There are a wide variety of theoretically possible accessors, varying in what kind of access they provide and what mechanism they use to provide it.  Different programming languages support different combinations.  For Swift, there are three fundamental kinds of access that we are interested in:
+
+* Read access.
+* Write or “set” access.
+* Update access.
+
+Write and update access differ in that write access can create a value that did not exist before, while update access requires there be a pre-existing value.  For example, consider how the dictionary write operation `myDictionary["newKey"] = value` can create a value that did not exist before. In contrast, `myDictionary["existingKey"] += 1` modifies a value that must necessarily already exist.
+
+For each kind of access, there are multiple ways that the access might be provided.  The Swift implementation today supports three different possible mechanisms, although not all of these are currently supported by accessors:
+
+* Copying.  The value might be copied between the caller and the callee.
+* Borrowing.  The callee might provide access to the value without actually copying it.  This can be implemented by having the callee provide a pointer or reference to the value *in situ.*
+* Yielding.  Each of the above is typically provided by calling a function.  The value or pointer is passed as an argument or return value from that function.  It’s also possible for the callee to start a *coroutine* whose execution is interleaved with that of the caller.  This allows the callee to resume after the caller is finished, which is not possible for a function- or method-based accessor.
+
+## Six of Nine Fundamental Accessors
+
+Combining the possibilities above gives us a 3x3 matrix of possible accessors:
+
+|           |Read   |Write  |Update |
+|---        |---    |---    |---    |
+|Copying    |       |       |✗      |
+|Borrowing  |       |✗      |       |
+|Yielding   |       |✗      |       |
+
+Three of these combinations turn out to be either impossible or not useful:
+
+* “Copying Update” is unnecessary.  A copying update would by definition copy the value into the caller and then copy the updated value back, which is easily achieved by using a “Copying Read” followed by a “Copying Write”
+* “Borrowing Write” and “Yielding Write” are simply impossible.  Both borrowing and yielding require an existing value.  As such, they provide “update” access.
+
+## Existing and Proposed Standard Accessors
+
+Based on the discussion above, we are left with six combinations that are both possible and useful.  The following table gives each one a name:
+
+|          |Read   |Write |Update |
+|---       |---    |---   |---    |
+|Copying   |get    |set   |✗      |
+|Borrowing |borrow |✗     |mutate |
+|Yielding  |read   |✗     |modify |
+
+Some of the above exist in the current Swift language, others we expect to be proposed in the near future.  Here is a slightly more detailed explanation of each one:
+
+* `get` and `set` - These are the original standard accessors, which compile into methods that return or accept a value of the indicated type.
+* `borrow` and `mutate` - These borrowing accessors are not yet implemented, but we expect to propose them for Swift Evolution at some point in the future.  These provide access to the value without copying by relying on Swift’s “borrow” implementation which avoids pointer overhead for simple types.  In particular, these can be used for non-copyable and non-escapable values.  (In various discussions, the proposed `mutate` accessor has been called `inout`.)
+* `read` and `modify` -  These exist today as experimental `_read` and `_modify` implementations and we expect to propose a revised final form without the underscored names in the near future. They compile into coroutines that “yield” the value.  In effect, each such accessor becomes two functions:  The “top half of the coroutine” creates a value if necessary and then provides the caller with a pointer to this value.  The “bottom half of the coroutine” cleans up the value.  In effect, callers invoke the top half to obtain a pointer to the value, use the pointer to read and/or modify the value, then call the bottom half to give the property a chance to clean itself up.
+
+In the process of developing the above, a number of other approaches have been explored.  Some of the following are currently supported in the compiler, but none are expected to go through Swift Evolution
+
+* `unsafeAddress` and `unsafeMutableAddress` - These are implemented but are not expected to ever go through Swift Evolution.  They compile into methods that return a pointer to the value in question.  This pointer value is explicit in the property implementation but is not visible in the Swift source code of the client.  This does not require copying the value, but does require that the value already be present somewhere in memory.  This was an early form of the idea behind `borrow` and `mutate` but based on unsafe constructs.
+* `unsafeRawAddress` and `unsafeRawMutableAddress` - These are not yet implemented and may not ever go through Swift Evolution.  These are similar to the above but allow the property implementation to work directly with an untyped “raw” pointer while the caller sees an access to typed data in memory.  This document will refer to these and the previous two collectively as `unsafe*Address`.
+* `_read` and `_modify` - These were the early experimental forms of `read` and `modify`.  They will likely continue to be supported in order to avoid breaking existing code, but users should migrate to `read` and `modify` once they are finalized.
+
+## Distinguishing Features
+
+These accessors vary in how they treat both the property value and the containing value.  In this section, we’ll explore the forms that variation takes:
+
+### Copying vs. Borrowing
+
+A key distinction is whether a particular accessor copies the property value or whether it provides access to the value without copying it.  The `get` and `set` accessors copy the value by returning it as a method result or accepting it as a method argument.  Other accessors provide access to a value without copying:
+
+*  `unsafe*Address` have the accessor provide an explicit pointer
+* `read`/`modify` *yield* access to a value from a coroutine.  This can be a directly stored value, or a constructed value stored temporarily in a coroutine execution frame.
+* `borrow` and `mutate` allow the client to *borrow* the value.  Following the existing Swift ABI rules, this may involve passing a pointer or “bitwise borrowing” the value to avoid pointer overhead for small values.  (Note that bitwise borrowing can be used even for noncopyable values since it is not formally a copy.)
+
+The `unsafe*Address` and `borrow`/`mutate` can avoid copying because the property value is already in memory within the containing value.  The compiler needs to ensure that the containing value is not destroyed or modified until after the pointer or borrow reference is no longer in use.
+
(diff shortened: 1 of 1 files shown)


visions/accessors.md
```diff
@@ -0,0 +1,348 @@
...
+
+Another example:  Providing a `Span` over the contents of a String will require that we somehow handle the case where a short String is stored inline.
+
+### Access Scope
+
+The `get`/`set` accessors conceptually represent “instantaneous” access of the value.  After the `get` accessor returns, there is no relation between the returned copy of the property value and the containing value.  Each of the other accessors implies that the containing value must continue to exist for some period.  This period is generally fairly short, but optimization can extend this interval in order to simplify other operations:
+
+* `unsafe*Address` accessors return pointers into the containing value.  This is safe for the caller of these accessors because the compiler knows about this relationship and can extend the lifetime of the containing value as needed.  (These accessors are nominally “unsafe” because their implementation requires constructing an unsafe pointer and there are no checks to ensure that the pointer so constructed is in fact valid.)
```
apple-fcloutier, line 103: It sounds like the compiler makes an assumption that the object that has the accessor owns the storage that is being returned. That might not be the case and you could have memory safety problems even if the pointer was constructed correctly. From this description, it sounds like if you have a weak/unowned reference and you return a pointer to one of its fields, we might be able to find a UAF bug.

visions/accessors.md
```diff
@@ -0,0 +1,348 @@
...
+* `read`/`modify` expose a value for the lifetime of a coroutine. Coroutines enforce that the containing value remains alive for the duration of the coroutine.  Note that in current Swift, the coroutines are generally quite short-lived and the compiler copies the value into or out of the coroutine fairly aggressively.  In the future, the compiler will likely become more adept at expanding the coroutine lifetime to reduce such copying.
+* `borrow`/`mutate` return a borrow of the property value.  This borrow has an implied dependence on the containing value, and the compiler must guarantee that the containing value outlives the property access.
```
apple-fcloutier, line 105: I imagine that borrow/mutate would not work for the example I gave, given that the accessor would need to take a strong reference and there wouldn't be a "bottom half coroutine" to drop it?

visions/accessors.md
```diff
@@ -0,0 +1,348 @@
...
+
+* `unsafeAddress` and `unsafeMutableAddress` - These are implemented but are not expected to ever go through Swift Evolution.  They compile into methods that return a pointer to the value in question.  This pointer value is explicit in the property implementation but is not visible in the Swift source code of the client.  This does not require copying the value, but does require that the value already be present somewhere in memory.  This was an early form of the idea behind `borrow` and `mutate` but based on unsafe constructs.
+* `unsafeRawAddress` and `unsafeRawMutableAddress` - These are not yet implemented and may not ever go through Swift Evolution.  These are similar to the above but allow the property implementation to work directly with an untyped “raw” pointer while the caller sees an access to typed data in memory.  This document will refer to these and the previous two collectively as `unsafe*Address`.
+* `_read` and `_modify` - These were the early experimental forms of `read` and `modify`.  They will likely continue to be supported in order to avoid breaking existing code, but users should migrate to `read` and `modify` once they are finalized.
```
apple-fcloutier, line 75: Will it be an ABI break to move from the underscored versions to the finalized versions?

visions/accessors.md
```diff
@@ -0,0 +1,348 @@
...
+* `borrow`/`mutate` return a borrow of the property value.  This borrow has an implied dependence on the containing value, and the compiler must guarantee that the containing value outlives the property access.
+
+### Ownership of the containing value
+
+All of the above accessors include a reading variant (`get`, `read`, `borrow`) and a writing variant (`set`, `modify`, `mutate`).  By default, the reading variant is not considered to be a mutation of the enclosing value.  The writing variant conversely is considered to be a write access of the enclosing value and the usual Swift exclusivity rules apply to limit the scope of such access.
+
+However, a reading accessor can be explicitly marked as `mutating`.  This relaxes checks on the accessor implementation so that it can mutate the containing value, and also causes the caller to treat this as a write access requiring exclusivity enforcement.  This is commonly used for caching or bookkeeping:
```
apple-fcloutier, line 111: Does this imply the existence of a `mutating read`? What is the difference between `mutating read` and `modify`? (I guess it gives you a read-only object while requiring the enclosing object to be mutable?)

visions/accessors.md
```diff
@@ -0,0 +1,348 @@
...
+}
+
+let v = Values()
+let v2 = v.dropOldest
+// `v` is no longer alive here
+```
+
+Note that `consuming` makes no sense for setters — there is no point to changing a property on a value and then immediately ending the lifetime of that value.  Similarly, `consuming` makes no sense for `borrow` or `unsafeAddress` accessors — those require that the containing value survive for the duration of the returned borrow access or address, so it does not make sense to explicitly terminate the value lifetime immediately.
+
+This only leaves `consuming get` and `consuming read` as meaningful combinations.  A `consuming get` can be used for operations such as the `dropOldest` example above that model a transformation by creating a new value and terminating the old one.  The `consuming read` variant can be used similarly.  Unlike `borrow`, the `read` operation coroutine structure forces the containing value to live for a certain period of time, the value is consumed only at the end of that coroutine.
```
apple-fcloutier, line 151: In a `consuming get`, I imagine that the object is `deinit`ed at some point before the accessor returns. In a `consuming read`, is it `deinit`ed at the return of the top half or the return of the bottom half?

visions/accessors.md
```diff
@@ -0,0 +1,348 @@
...
+
+#### Some combinations are nonsensical
+
+* Providing both a `mutate` and a `modify` for the same property doesn’t make sense.  If `mutate` is possible, then you can provide the value without making a copy, which means `modify` is unnecessary and redundant.
+
+#### We need to support existing APIs
+
+The current Swift API seems to require the following:
+
+* `set` is needed to be able to initialize new values for subscript operations.  `modify` is capable of doing this for types that can be trivially initialized, but there can be considerable overhead to construct an empty placeholder, expose it for mutation, and then record the final result.
```
apple-fcloutier, line 173: What combinations are legal/intended to be legal? It sounds like you can have a property that has both `set` and `modify`? If that's the case, are there interesting guarantees about the accessor that Swift will pick for a given operation?

visions/accessors.md
```diff
@@ -0,0 +1,348 @@
...
+* Either `borrow` or `read` is needed to provide safe reading of noncopyable values
+* Either `mutate` or `modify` is needed to provide safe mutation of noncopyable values
+* We have `Dictionary` APIs (among others) that expose values as a different type than they are stored.  These can be supported using `read`/`modify` and storing the transformed value in the coroutine frame for the duration of the access.  They could also be supported using `get` to return a transformed value, but that would require a more expensive read-modify-write cycle for modifications.
+* Because `read`/`modify` have constraints on the lifetime of the access, they cannot fully replace `borrow`/`mutate`.
```
apple-fcloutier, line 177: Possibly related to the other comment: what is the difference between the lifetime constraints?

visions/accessors.md
tbkka: We do plan to provide a way to make this migration while preserving ABI (for people who need to do so).

visions/accessors.md
tbkka: Exactly:  A `modify` accessor provides an API for the client to mutate the property; a `mutating read` allows the read operation to have side-effects (such as updating a cache).  Mutating reads need to be declared as such so the compiler can accurately diagnose cases where those side-effects could cause problems.

visions/accessors.md
tbkka: I'm not entirely certain whether technically the owning value is destroyed before or after the accessor returns.  (@jckarter would certainly know.)  In the coroutine case, we could not invoke the bottom half if the value were already destroyed, so it would have to be destroyed at the end of the bottom half.

visions/accessors.md
tbkka: You can certainly have situations where there is more than one accessor that could be used to satisfy a particular access.  I should track down someone who knows the current rules and see if I can lay them out here.

visions/accessors.md
```diff
@@ -0,0 +1,348 @@
...
+|              |Read   |Write  |Update |
+|---           |---    |---    |---    |
+|**Copying**   |       |       |✗      |
+|**Borrowing** |       |✗      |       |
+|**Yielding**  |       |✗      |       |
+
+Three of these combinations turn out to be either impossible or not useful:
+
+* “Copying Update” is unnecessary.  A copying update would by definition copy the value into the caller and then copy the updated value back, which is easily achieved by using a “Copying Read” followed by a “Copying Write”
+* “Borrowing Write” and “Yielding Write” are simply impossible.  Both borrowing and yielding require an existing value.  As such, they provide “update” access.
```
xwu, line 53: This seems to be a think-o. If you construe "copying write" (as you do below) to mean copying the source into the destination, then it is conceivable—even if not broadly useful—to borrow that source value or even to manipulate a yielded value as part of a write operation (as long as you put something back).

Here you are arguing that these are "impossible" because there is no existing value at the destination—but only in the same sense that "copying write" would be useless because you're not copying the value at the destination back to the destination.

visions/accessors.md
```diff
@@ -0,0 +1,348 @@
...
+
+|              |Read   |Write  |Update |
+|---           |---    |---    |---    |
+|**Copying**   |       |       |✗      |
+|**Borrowing** |       |✗      |       |
+|**Yielding**  |       |✗      |       |
+
+Three of these combinations turn out to be either impossible or not useful:
+
+* “Copying Update” is unnecessary.  A copying update would by definition copy the value into the caller and then copy the updated value back, which is easily achieved by using a “Copying Read” followed by a “Copying Write”
```
xwu, line 52: Is that so categorically or only given present expressivity limits we place in the language? For example, if access can be async or throwing, is there some expressivity made possible with a unique "copying update" with respect to guaranteeing that there is/isn't an `async` operation between reading and writing, and/or that an update does/doesn't attempt to write if reading throws, etc.?

visions/accessors.md
```diff
@@ -0,0 +1,348 @@
...
+
+## Six of Nine Fundamental Accessors
+
+Combining the possibilities above gives us a 3x3 matrix of possible accessors:
+
+|              |Read   |Write  |Update |
+|---           |---    |---    |---    |
+|**Copying**   |       |       |✗      |
+|**Borrowing** |       |✗      |       |
+|**Yielding**  |       |✗      |       |
```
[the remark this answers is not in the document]
tbkka: I moved that sentence to before the table to make it clearer.  ("N/A" is clearer by itself, but the next table builds on this and `✗` is a better contrast with the keywords there.)

visions/accessors.md
```diff
@@ -0,0 +1,348 @@
...
+
+These limitations have led Swift to add a variety of other accessors, including several experimental forms that have never been officially standardized through the Swift Evolution process.  Recent work on noncopyable support and an expanded interest in new kinds of collections has made it important to provide a complete standard set of accessors to support future work.  This document will explain the various accessors that currently exist or might be proposed in the near future and how they will provide the full set of capabilities needed for different applications.
+
+> Note:  Accessors are also used to implement subscripts; the only real difference is that a subscript operation can have additional arguments.  This document will discuss property accesses and subscript operations interchangeably.
+
+## Two Dimensions of Access
```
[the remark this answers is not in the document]
tbkka: Two.

visions/accessors.md
```diff
@@ -0,0 +1,348 @@
...
+
+There are a wide variety of theoretically possible accessors, varying in what kind of access they provide and what mechanism they use to provide it.  Different programming languages support different combinations.  For Swift, there are three fundamental kinds of access that we are interested in:
+
+* Read access.
+* Write or “set” access.
+* Update access.
+
+Write and update access differ in that write access can create a value that did not exist before, while update access requires there be a pre-existing value.  For example, consider how the dictionary write operation `myDictionary["newKey"] = value` can create a value that did not exist before. In contrast, `myDictionary["existingKey"] += 1` modifies a value that must necessarily already exist.
```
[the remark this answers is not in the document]
tbkka: Swift's actual `Dictionary` indeed does not quite support this, but a simple property misses the key point:  A pure write operation can create/initialize a stored value where no stored value existed before.  A simple property always has storage associated and by Swift's definite-initialization rules that storage must always be initialized.

I've changed the text here to refer more vaguely to "write operation on an expandable collection" to try to avoid this problem.

visions/accessors.md
```diff
@@ -0,0 +1,348 @@
...
+
+The `unsafe*Address` and `borrow`/`mutate` can avoid copying because the property value is already in memory within the containing value.  The compiler needs to ensure that the containing value is not destroyed or modified until after the pointer or borrow reference is no longer in use.
+
+A `read` or `modify` accessor can usually avoid a copy, but not always:  In practice, the implementation cannot delay the bottom half of the coroutine indefinitely.  This can lead to situations where the value is still needed after the coroutine ends, which requires copying the value from the coroutine’s execution frame into the calling context.  When this happens, a `_read` or `_modify` accessor is generally slower than a regular `get`, since it incurs copy overhead similar to a `get` in addition to the coroutine execution overhead.  This also limits the use of `read` or `modify` with property values that are not copyable.
```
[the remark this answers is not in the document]
tbkka: Today, the optimizer is pretty quick to insert a copy and end the coroutine.  I expect that will change with ongoing optimizer improvements to dynamically determine the best point to end the coroutine.  This is one of the motivations for the `borrow` accessor:  It will be much easier to reason about copies with that accessor.

visions/accessors.md
tbkka: Yes, one way in which the pointer could be "not valid" is that the pointee's lifetime is not assured.

visions/accessors.md
tbkka: Unless `unsafe*Address`, the `borrow`/`mutate` accessors should be amenable to various checks at the time the accessor is compiled.  For example, we could warn or error if you tried to return an interior pointer to a class object where there was not an associated strong reference to keep it alive.  So `borrow`/`mutate` would be safe in that we should be able to issue an error if you try to do this -- but for your example as written, it would likely require some variation of coroutine-based accessor.

visions/accessors.md
tbkka: Remember that accessors are also used for subscript operations, and consider a collection that allocates new storage to hold a new value.  A "copying write" adds a new value to the container by copying it from the client into the container.  This works:  the write accessor accepts the value as a function argument, that function allocates the storage and puts the value into that storage.  A "borrowing write" has the container return a reference to the _existing_ value storage to the client so that the client can then update it in place without further involvement from the container.

In this scenario, such a "borrowing write" is truly impossible:  the collection doesn't even have the storage before the access begins; if it does have such storage, it may not be initialized.  This is clearest with dictionary-like containers (though Swift's standard library `Dictionary` type has a wrinkle that makes this a little tricky to demonstrate).  Such a container literally cannot yield or lend the existing value because there is no existing value to be lent or yielded.  Even a coroutine accessor (with a coroutine frame where temporary values can be stored) may not work because it may not be possible to initialize the temporary.

Does that clarify things?  (You might be thinking that by "borrowing write" I'm referring to a `set` accessor that accepts its argument as a borrow.  That is _not_ what I am attempting to describe here.  I'll play with the explanation to see if I can clarify it.)

visions/accessors.md
tbkka: As I hinted earlier, there are many other theoretically possible kinds of accessors that I am not describing here. My goal with this document was to find a (hopefully) clear way to explain how a certain set of six accessors seems to address Swift's current need to efficiently express the next generation of standard library APIs.  I would be interested in hearing more details about any use cases that you feel these do not cover.

For what it's worth:  The kind of guarantee you seem to be describing could be provided by a borrowing accessor plus a "swap" operator, or by a closure-taking update method.

visions/accessors.md
tbkka: The more detailed description further down for `read` and `borrow` illustrates this difference.

visions/accessors.md
xwu: Yes, I understand: My feedback is that the terminology is used inconsistently and the 3x3 matrix is hence not a systematic grid—as you say, you use the term "copying write" to mean a copy from the client to the container, but you _don't_ use "borrowing write" to mean a borrow from the client to the container.

If the directionality were consistently aligned with what you mean by "borrowing write," then a "copying write" (and hence the entire "write" column) would be impossible and/or useless by construction, since there's likewise no way to copy a non-existent value from the container.

visions/accessors.md
tbkka: If you really think this grid is not helpful, I can delete this section.  For myself, this entire document started as an unstructured list of many different kinds of accessors and I struggled to try to explain how that large menagerie could be distilled down to a small set.  This grid was a key step in crystallizing my own thinking about this area:  Among other things, it helps explain why "set" is necessary regardless of what other accessors you have (it's the only option in the "write" column).    But different people approach these things differently, and a tool that was useful for me may not be useful for others.  I'll think about removing this section.

tbkka:
Withdrawing this PR, since @rjmccall now has a completely rewritten version of this that's being discussed on the forums.

PR #2622, opened 2024-11-24
Normalize SE-0451 author and previous proposal fields

### LINKED ISSUE #856: SE-0451 has an additional author by mistake
The SE-0451 proposal has three authors on the list:

> Authors: [Tony Allevato](https://github.com/allevato), [SE-0275](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0275-allow-more-characters-like-whitespaces-and-punctuations-for-escaped-identifiers.md), [Alfredo Delli Bovi](https://github.com/adellibovi)

I believe this informations have been extracted from the proposal page, which contains the following:

> Author: [Tony Allevato](https://github.com/allevato); based on [SE-0275](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0275-allow-more-characters-like-whitespaces-and-punctuations-for-escaped-identifiers.md) by [Alfredo Delli Bovi](https://github.com/adellibovi)

I believe the extraction algorithm is taking all links from that line, which works correctly most of the time, but not in this case. I'm not sure what the ideal solution is here, if changing the proposal or the algorithm. The approach I took on my app that makes use of the metadata was to skip over the names that look too much like a proposal ID.

dempseyatgithub:
The author field of SE-0451 includes a reference to the previous proposal that SE-0451 was based on.

The `Author` field should only include people. The reference to a previous proposal should be in the `Previous Proposal` field.

Note also that the author of the previous proposal (SE-0275) is properly credited in that proposal and does not need to be included as part of the current proposal (SE-0451) unless the author of SE-0451 wishes to include them as a co-author.

This PR normalizes the `Author` and `Previous Proposal` fields and closes https://github.com/swiftlang/swift-org-website/issues/856.

### DIFF
--- a/proposals/0451-escaped-identifiers.md
+++ b/proposals/0451-escaped-identifiers.md
@@ -1,10 +1,11 @@
 # Raw identifiers
 
 * Proposal: [SE-0451](0451-escaped-identifiers.md)
-* Author: [Tony Allevato](https://github.com/allevato); based on [SE-0275](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0275-allow-more-characters-like-whitespaces-and-punctuations-for-escaped-identifiers.md) by [Alfredo Delli Bovi](https://github.com/adellibovi)
+* Author: [Tony Allevato](https://github.com/allevato)
 * Review Manager: [Joe Groff](https://github.com/jckarter)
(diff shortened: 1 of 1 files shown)


rjmccall:
@allevato, would you like to include Alfredo as a co-author?  You should probably only do so if he's interested in still participating.

PR #2624, opened 2024-12-04
[SE-0314] Fix typo

ole:
The usage of "`AsyncSeries`" looks like a typo. The term doesn't appear anywhere else in the document. The document does use the word "series" a few times as an alternative to "stream", but using it as an identifier seems wrong.

I replaced "`AsyncSeries`" with "`AsyncThrowingStream`" because it appears in the section on `AsyncThrowingStream`. Replacing it with "`AsyncStream`" would also work, but since this section is about the throwing variant, `AsyncThrowingStream` seems better.


rjmccall (approved):

rjmccall:
Thanks!

PR #2630, opened 2024-12-06
[SE-0453] Update `Vector` example and doc comment

benrimmington:
Cc: @Azoy, @Jumhyn


Azoy (approved):
Thanks! This looks good to me, @Jumhyn?

PR #2631, opened 2024-12-11
Update SE-0452 in response to LSG review

jckarter:


proposals/0452-integer-generic-parameters.md
```diff
@@ -54,11 +54,29 @@ Or it can be instantiated using integer generic parameters from the surrounding
...
 ```
 
+Integer generic parameters become static constant members of the type, with
+the same visibility as the type itself:
+
+```swift 
+public struct Matrix<let columns: Int, let rows: Int> {
+    // implicitly has these members:
+    //  public static let columns: Int
+    //  public static let row: Int
```
benrimmington, line 69: Typo: row(s). Should these be *computed* properties? Static stored properties aren't supported in generic types.

```suggestion
    //  public static var columns: Int { get }
    //  public static var rows: Int { get }
```

PR #2635, opened 2024-12-17
Opt-in Strict Memory Safety Checking

DougGregor:


proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,387 @@
...
+
+* **Lifetime safety** : all accesses to a value are guaranteed to occur during its lifetime. Violations of this property, such as accessing a value after its lifetime has ended, are often called use-after-free errors.
+* **Bounds safety**: all accesses to memory are within the intended bounds of the memory allocation, such as accessing elements in an array. Violations of this property are called out-of-bounds accesses.
+* **Type safety** : all accesses to a value use the type to which it was initialized, or a type that is compatible with that type. For example, one cannot access a `String` value as if it were an `Array`. Violations of this property are called type confusions.
+* **Initialization safety** : all values are initialized property to being used, so they cannot contain unexpected data. Violations of this property often lead to information disclosures (where data that should be invisible becomes available) or even other memory-safety issues like use-after-frees or type confusions.
```
finagolfin, line 30: initialized property -> initialized properly prior

finagolfin: Still missing the word "prior"

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,437 @@
...
+import B
+
+extension MyType {
+  @unsafe public func checksum() -> Int32 {}
+    return wrapper.checksum()
+  }
+}
+```
+
+However, these two APIs diff in the severity of the problem they post for memory safety: `wrapper` has an unsafe type in its interface, so any use of `wrapper` is fundamentally unsafe. However, `MyType` itself can encapsulate the memory-unsafe behavior within a safe API., The `MyType.checksum` operation is actually safe for clients to use. Marking it as `@unsafe` is therefore undesirable, because it (incorrectly) forces clients to treat this as an unsafe API, causing unnecessary extra work and deluting the value of correctly-identified unsafe APIs.
```
Xazax-hun, line 174:
```suggestion
However, these two APIs differ in the severity of the problem they post for memory safety: `wrapper` has an unsafe type in its interface, so any use of `wrapper` is fundamentally unsafe. However, `MyType` itself can encapsulate the memory-unsafe behavior within a safe API., The `MyType.checksum` operation is actually safe for clients to use. Marking it as `@unsafe` is therefore undesirable, because it (incorrectly) forces clients to treat this as an unsafe API, causing unnecessary extra work and deluting the value of correctly-identified unsafe APIs.
```

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,475 @@
...
+  @unsafe func f() { ... } // warning: override of safe instance method with unsafe instance method
+}
+```
+
+to suppress this warning, the `Sub` class itself can be marked as `@unsafe`, e.g.,
+
+```swift
+@unsafe
+class Sub: Super {
+  func f() { ... } // warning: override of safe instance method with unsafe instance method
```
stevapple, line 266: I think there should be no more warning…?
```suggestion
  func f() { ... }
```

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,475 @@
...
+    return wrapper.checksum()
+  }
+}
+```
+
+Module `A` defines a type, `DataWrapper`, that is `@unsafe`. It can be compiled with or without strict safety checking enabled, and is fine either way.
+
+Module `B` uses the `DataWrapper` type. If compiled without strict safety checking, there will be no diagnostics about memory safety. If compiled with strict safety checking, there will be a diagnostic about `wrapper` using an `@unsafe` type (`DataWrapper`) in its interface. This diagnostic can be ignored.
+
+In module `C` enables strict memory safety, the use of `MyType` is considered safe (since it was not marked `@unsafe` and doesn't involve unsafe types in its interface). However, the access to `wrapper` will result in a diagnostic, because the type of `wrapper` involves an `@unsafe` type.
```
geoffreygaren, line 149: "In" => "If"

proposals/nnnn-strict-memory-safety.md
geoffreygaren: post => pose

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,464 @@
...
+### Unsafe compiler flags
+
+There are a number of compiler flags that intentionally disable some safety-related checking. For each of these flags, the compiler will produce a diagnostic if they are used with strict memory safety:
+
+* `-Ounchecked`, which disables some checking in the standard library, including (for example) bounds checking on array accesses.
+* `-enforce-exclusivity=unchecked` and `-enforce-exclusivity=none`, which disables exclusivity checking that is needed for memory safety.
+* `-strict-concurrency=` for anything other than "complete", because the memory safety model requires strict concurrency to eliminate thread safety issues.
+* `-disable-access-control`, which allows one to break invariants of a type that can lead to memory-safety issues, such as breaking the invariant of `Range` that the lower bound not exceed the upper bound.
+
+### Unsafe overrides
```
apple-fcloutier, line 233: For lack of a better place to ask this question: what happens with `enum`s that have cases with unsafe payloads?

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,464 @@
...
+### `unsafe` expression
+
+When a declaration is marked `@unsafe`, it is free to use any other unsafe types as part of its interface. Any time there is executable code that makes use of unsafe constructs, that code must be within an `unsafe` expression or it will receive a diagnostic about uses of unsafe code. In the example from the previous section, `wrapper` can be marked as `@unsafe` to suppress diagnostics by explicitly propagating unsafety to their clients:
+
+```swift
+// Module B
+import A
+
+public struct MyType {
+  @unsafe public var wrapper: DataWrapper
```
apple-fcloutier, line 174: Can you clarify whether `@unsafe` is necessary on local variables of unsafe types as well? Similarly, what about local functions?

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,464 @@
...
+## Proposed solution
+
+This proposal introduces an opt-in strict memory safety checking mode that identifies all uses of unsafe behavior within the given module. There are several parts to this change:
+
+* A compiler flag `-strict-memory-safety` that enables warnings for all uses of unsafe constructs within a given module. All warnings will be in the diagnostic group `Unsafe`, enabling precise control over memory-safety-related warnings per [SE-0443](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0443-warning-control-flags.md).
+* An attribute `@unsafe` that indicates that a declaration is unsafe to use. Such declarations may use unsafe constructs within their signatures.
+* An `unsafe` expression that marks any use of unsafe constructs in an expression, much like `try` and `await`.
+* Standard library annotations to identify unsafe declarations.
+
+### Example of `unsafe` usage
```
apple-fcloutier, line 48: Can you confirm that I understand the rules correctly:

* If you have a symbol that uses an unsafe type in its signature, it needs to be `@unsafe` itself to avoid a diagnostic
* When you use a symbol that is `@unsafe` in an expression, it needs to be inside an `unsafe` expression
* Using `unsafe` in a function has no bearing on whether that function needs to be `@unsafe` or not

I am especially interested in the last point. I originally thought that your example showed `unsafe` markers are not necessary in `@unsafe` functions, but the "alternatives considered" section on the unsafe keyword in Rust makes me believe I initially got it wrong, and it took me a second read to realize you did not really say anything one way or the other.

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,464 @@
...
+
+The `@unsafe` attribute can be applied to any declaration to indicate that use of that declaration can undermine memory safety. Any use of a declaration marked `@unsafe` will result in a warning. The closest analogue in the language today is `@available(*, deprecated)`, which has effectively no impact on the type system, yet any use of a deprecated declaration results in a warning.
+
+When a type is marked `@unsafe`, a declaration that uses that type in its interface is implicitly `@unsafe`. For example, consider a program containing three separate modules:
+
+```swift
+// Module A
+@unsafe
+public struct DataWrapper {
+  var buffer: UnsafeBufferPointer<UInt8>
```
apple-fcloutier, line 136: For lack of a much better place to ask: if a function has an argument of an unsafe type but it has a default value, does the function need to be `@unsafe`? I would say yes, but this has implications for synthesized memberwise initializers.

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,464 @@
...
+* Interoperability with the C family of languages is an important feature for Swift. Most C(++) APIs are unlikely to ever adopt the safety-related attributes described above, which means that enabling strict safety checking by default would undermine the usability of C(++) interoperability.
+* Swift's current (non-strict) memory safety by default is likely to be good enough for the vast majority of users of Swift, so the benefit of enabling stricter checking by default is unlikely to be worth the disruption it would cause.
+
+### Optional `message` for the `@unsafe` attribute
+
+We could introduce an optional `message` argument to the `@unsafe` attribute, which would allow programmers to indicate *why* the use of a particular declaration is unsafe and, more importantly, how to safely write code that uses it. However, this argument isn't strictly necessary: a comment could provide the same information, and there is established tooling to expose comments to programmers that wouldn't be present for this attribute's message, so we have omitted this feature.
```
apple-fcloutier, line 460: This makes sense to me, but I wanted to see if you have any tooling in mind. DocC?

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,464 @@
...
+    }
+  }
+}
+```
+
+The `unsafe` keyword here indicates the presence of unsafe code within that expression. As with `try` and `await`, it can cover multiple sources of unsafety within that expression: the call to `c_library_sum_function` is unsafe, as is the use of `buffer` and `buffer.baseAddress`, yet they are all covered by one `unsafe`.
+
+Note that we do *not* require that the `sum` function be marked `@unsafe` just because it has unsafe code in it. The programmer may choose to indicate that `sum` is unsafe, but the assumption is that unsafe behavior is properly encapsulated when using `unsafe`.
+
+### Incremental adoption
```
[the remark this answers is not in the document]
DougGregor: Sure

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,464 @@
...
+* `withUnsafeCurrentTask` and `UnsafeCurrentTask`: The `UnsafeCurrentTask` type does not provide lifetime safety, and must only be used within the closure passed to `withUnsafeCurrentTask`.
+* `UnownedSerialExecutor`: This type is intentionally not lifetime safe. It's primary use is the `unownedExecutor` property of the `Actor` protocol, which documents the lifetime assumptions of the `UnownedSerialExecutor` instance it produces.
+
+All of these APIs will be marked `@unsafe`. For all of the types that are `@unsafe`, any API that uses that type in its signature will also be marked `@unsafe`, such as `Array.withUnsafeBufferPointer`. Unless mentioned above, standard library APIs that do not have an unsafe type in their signature, but use unsafe constructs in their implementation, will be considered to be safe.
+
+### Unsafe compiler flags
+
+There are a number of compiler flags that intentionally disable some safety-related checking. For each of these flags, the compiler will produce a diagnostic if they are used with strict memory safety:
```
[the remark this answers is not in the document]
DougGregor: As with all of the other strict safety-related diagnostics, it's a warning in the `Unsafe` group.

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,464 @@
...
+
+```swift
+protocol P {
+  func f()
+}
+
+struct ConformsToP { }
+
+extension ConformsToP: P {
+  @unsafe func f() { } // warning: unsafe instance method 'f()' cannot satisfy safe requirement
```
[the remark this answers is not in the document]
DougGregor: Happy to take suggestions on rewording this. We use the "cannot satisfy" terminology in many places for this kind of issue, and everything else I try seems wordier and not better.

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,464 @@
...
+```swift
+@unsafe
+class Sub: Super {
+  func f() { ... } // no more warning
+}
+```
+
+The `@unsafe` annotation is at the class level because any use of the `Sub` type can now introduce unsafe behavior, and any indication of that unsafe behavior will be lost once that `Sub` is converted to a `Super` instance.
+
+### Unsafe conformances
```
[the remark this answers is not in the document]
DougGregor: Yes, protocols can be `@unsafe`. Conformances to `@unsafe` protocols need not be annotated as `@unsafe`, because every use of the protocol already has to acknowledge the unsafety.

proposals/nnnn-strict-memory-safety.md
DougGregor: The cases themselves need to be marked `@unsafe`.

proposals/nnnn-strict-memory-safety.md
DougGregor: Yes, it is.

proposals/nnnn-strict-memory-safety.md
DougGregor: No, the function does not need to be `@unsafe`: the actual value of the default argument is effectively part of the implementation of the function.

proposals/nnnn-strict-memory-safety.md
DougGregor: You are correct, and I agree that the last point deserves more discussion. I've added an Alternatives Considered about `@unsafe` possibly implying `unsafe` throughout the body.

proposals/nnnn-strict-memory-safety.md
apple-fcloutier: Like this?

```swift
func illegal(pointer: UnsafePointer<Int>?) // warning: needs @unsafe
func legal(pointer: UnsafePointer<Int>? = nil) // no @unsafe, no warning
```

Is `@unsafe` implied on `legal(pointer: <something here>)`?

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,536 @@
...
+
+Swift provides memory safety with a combination of language affordances and runtime checking. However, Swift also deliberately includes some unsafe constructs, such as the `Unsafe` pointer types in the standard library, language features like `nonisolated(unsafe)`, and interoperability with unsafe languages like C. For most Swift developers, this is a pragmatic solution that appropriate level of memory safety while not getting in the way.
+
+However, there are certain projects, organizations, and code bases that require stronger memory-safety guarantees, such as in security-critical subsystems handling untrusted data or that are executing with elevated privileges in an OS. This proposal introduces opt-in strict memory safety checking to identify those places in Swift code that make use of unsafe language constructs and APIs. Any code written within this strictly-safe subset also works as “normal” Swift and can interoperate with existing Swift code.
```
rjmccall, line 17: "there are certain projects" reads as euphemistic to me, like you're alluding to top secret work. "some projects" would be a lot less leading.

I think you can set up the logic a little more clearly here. Some projects want stronger safety guarantees. Those projects want to pay closer attention to their unsafe constructs and discourage casual use. This proposal adds a mode to allow projects to opt in to that. Code in this mode is still valid as normal Swift and perfectly compatible.

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,536 @@
...
+
+Providing memory safety does not imply the absence of run-time failures. Good language design often means defining away runtime failures in the type system. However, memory safety requires only that an error in the program cannot be escalated into a violation of one of the safety properties. For example, having reference types be non-nullable by default defines away most problems with NULL pointers. With explicit optional types, the force-unwrap operator (postfix `!` ) meets the definition of memory safety by trapping at runtime if the unwrapped optional is `nil` . The standard library also provides the [`unsafelyUnwrapped` property](https://developer.apple.com/documentation/swift/optional/unsafelyunwrapped) that does not check for `nil` in release builds: this does not meet the definition of memory safety because it admits violations of initialization and lifetime safety that could be exploited.
```
rjmccall, line 37: Pick one of "run-time failures" or "runtime failures".

The argument here is a little muddy. Maybe: Swift achieves safety with a mixture of static and dynamic checks. Static checks are better when possible. Dynamic checks are sometimes necessary and are still safe as long as failure can't escalate. Swift offers unsafe features to allow problems to be solved when neither static nor dynamic checks are good enough. Unsafe features can still be used correctly; they just have requirements that Swift can't automatically check.

For example, Swift solves null references with optional types. Statically, Swift prevents you from using an optional reference without checking it first. If you're sure it's non-null, you can use the `!` operator, which is safe because Swift will dynamically check.  If you really can't afford that dynamic check, you can use `unsafelyUnwrapped`. This can still be correct if you can prove that the reference is definitely non-null for some reason that Swift doesn't know. But it is an unsafe feature because it admits violations if you're wrong.

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,536 @@
...
+```swift
+extension Array {
+  @unsafe
+  func withUnsafeBufferPointerSimplified<T>(_ body: (UnsafeBufferPointer<Element>) -> T) -> T {
+    // ...
+  }
+}
+```
+
+Users of this function that also enable strict safety checking will see warnings when using it. For example:
```
rjmccall, line 79: Would it be worth spelling out that adding `@unsafe` to a type or function has no impact on users that aren't using the mode?

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,536 @@
...
+
+* Strict concurrency checking, the focus of the Swift 6 language mode, required major changes to the type system, including the propagation of `Sendable` and the understanding of what code must be run on the main actor. These are global properties that don't permit local reasoning, or even local fixes, making the interoperability problem particularly hard. In contrast, strict safety checking has little or no effect on the type system, and unsafety can be encapsulated with `unsafe` expressions or ignored by a module that doesn't enable the checking.
+* [Embedded Swift](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0337-support-incremental-migration-to-concurrency-checking.md) is a subset of Swift that works without a runtime. Like the proposed strictly-safe subset, code written in Embedded Swift will also work as regular Swift. Embedded Swift and the strict safety checking proposed here are orthogonal and can be composed to (for example) ensure that firmware written in Swift has no runtime and provides the best memory-safety guarantees.
```
rjmccall, line 116: We might want to avoid saying that this mode itself provides better memory-safety guarantees. Maybe "more carefully considers its use of unsafe features" or something like that?

In fact, that might be worth a whole paragraph somewhere around here: "Note that this mode doesn't by itself ensure that code is any more memory safe. What it does do is..."

proposals/nnnn-strict-memory-safety.md
rjmccall: Hmm. What you've specified is that, if the parameter has an unsafe type, the function does need to be unsafe. Are you saying that there's an exception to that rule for parameters with default arguments, and then unsafely checking would just be enforced on call sites that provided an explicit argument?

I think you might be answering a different question about when there's a use of an unsafe feature in a default argument expression (but it's not reflected in the argument type).

In any case, the behavior should be specified more clearly here.

proposals/nnnn-strict-memory-safety.md
rjmccall: That seems a little extreme. Aren't all uses of it already forced to be marked? Or is that limited to just uses of `@unsafe` APIs that use it, which would exclude things like just assigning it to some other variable?

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,536 @@
...
+```
+
+The `unsafe` expression is much like `try` and `await`, in that it acknowledges that unsafe constructs (`wrapper`) are used within the subexpression but otherwise does not change the type. Unlike `try` and `await`, which require the enclosing context to handle throwing or be asynchronous, respectively, the `unsafe` expression does not imply any requirements about the enclosing block: it is purely a marker to indicate the presence of unsafe code, silencing a diagnostic.
+
+### Unsafe language constructs
+
+The following language constructs are always considered to be unsafe:
+
+* `unowned(unsafe)`: Used to store a reference without maintaining its reference count. The safe counterpart, `unowned`, uses dynamic checking to ensure that the reference isn't accessed after the corresponding object has been released. The `unsafe` variant disables that dynamic checking. Uses of `unowned(unsafe)` entities are not memory-safe.
+* `unsafeAddressor`, `unsafeMutableAddressor`: These accessors vend an unsafe pointer, and are therefore unsafe to declare. Other accessors (e.g., `get` and `set`) can provide safe alternatives.
```
rjmccall, line 199: What does this mean for clients? Using the storage is itself `unsafe`?

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,536 @@
...
+## Source compatibility
+
+The `unsafe` keyword in this proposal will be introduced as a contextual keyword following the precedent set by `await`'s' introduction in [SE-0296](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0296-async-await.md) and `consume`'s introduction in [SE-0366](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0366-move-function.md). This allows `unsafe` to continue to be used as an identifier, albeit with a small potential to break existing source that uses `unsafe` as a function that is then called with a trailing closure, like this:
+
+```swift
+func unsafe(_ body: () -> Void) { }
+
+unsafe {
+  // currently calls 'unsafe(_:)', but will become an unsafe expression
+}
```
rjmccall, line 321: Interesting. This is still necessary with the removal of `unsafe` blocks? Because this expression seems to be a no-op. I guess having a different interpretation rule for `unsafe { ... }` would be a little weird, but maybe we can at least warn that it doesn't cover any code that could possibly be unsafe.

proposals/nnnn-strict-memory-safety.md
DougGregor: For constructing the case, it means the construction would need to be in an `unsafe` expression. Pattern matching is harder: we would need `@unsafe` or `unsafe` in the pattern somewhere.

proposals/nnnn-strict-memory-safety.md
DougGregor: Yes, all uses of it are already forced to be marked. We could exempt local variables from the `@unsafe` requirement without losing safety.

proposals/nnnn-strict-memory-safety.md
DougGregor: Ah, I meant that this function doesn't need to be `@unsafe`:

```swift
func hasDefault(value: Int = /*unsafe*/ getIntegerUnsafely()) { ... }
```

I've pushed a clarification here, thanks to both of you.

proposals/nnnn-strict-memory-safety.md
DougGregor: I'm stealing this, thanks

proposals/nnnn-strict-memory-safety.md
DougGregor: Sure

proposals/nnnn-strict-memory-safety.md
DougGregor: Yes, that's a good idea. We're not changing the semantics of anything here.

proposals/nnnn-strict-memory-safety.md
DougGregor: Hmm, that wasn't the intent.

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,557 @@
...
+      let base = /*unsafe*/ buffer.baseAddress
+      // ...
+    }
+  }
+  ```
+
+* Default arguments of functions are part of the implementation of a function, not its signature. For example, the following function does not have any unsafe types in its signature, so it does not require `@unsafe`, even though the default argument for `value` involves unsafe code. That unsafe code is effectively part of the body of the function, so it follows the rules for `unsafe` expressions.
+
+  ```swift
+  func hasDefault(value: Int = /*unsafe*/ getIntegerUnsafely()) { ... }
```
apple-fcloutier, line 183: As I understand it, you're saying that you don't need to use `unsafe hasDefault()`. However, I would expect that you still need to use `unsafe getIntegerUnsafely()` in the default argument expression, no?

proposals/nnnn-strict-memory-safety.md
apple-fcloutier: Thank you for addressing this case. I have another question given this code:

```swift
struct Foo {
	var a: Int
	var b: UnsafePointer<Int>? = nil
}
```

Is the synthesized memberwise initializer of `Foo` `@unsafe`?

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,557 @@
...
+* `unsafeAddressor`, `unsafeMutableAddressor`: These accessors vend an unsafe pointer, and are therefore unsafe to declare. Other accessors (e.g., `get` and `set`) can provide safe alternatives.
+* `@exclusivity(unchecked)`: Used to remove dynamic exclusivity checks from a particular variable, which can mean that dynamic exclusivity violations go undetected at run time, causing a memory safety violation. Uses of `@exclusivity(unchecked)` entities are not memory-safe.
+
+The following language constructs are considered to be unsafe when strict concurrency checking is enabled (i.e., in the Swift 6 language mode):
+
+* `nonisolated(unsafe)`: Allows a property to be accessed from concurrent code without ensuring that such accesses are done so safely. Uses of `nonisolated(unsafe)` entities are not memory-safe.
+* `@preconcurrency` imports: Suppresses diagnostics related to data race safety when they relate to specific imported modules, which can introduce thread safety issues. The `@preconcurrency` import will need to be annotated with `@unsafe` in the strict dsafety mode.
```
apple-fcloutier, line 226: sp: "strict dsafety mode"

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,557 @@
...
+
+### Unsafe standard library APIs
+
+In the standard library, the following functions and types would be marked `@unsafe` :
+
+* `Unsafe(Mutable)(Raw)(Buffer)Pointer`, `OpaquePointer`, `CVaListPointer`: These types provide neither lifetime nor bounds safety. Over time, Swift code is likely to move toward their safe replacements, such as `(Raw)Span`.
+* `(Closed)Range.init(uncheckedBounds:)`: This operation makes it possible to create a range that doesn't satisfy invariants on which other bounds safety checking (e.g., in `Array.subscript`)
```
apple-fcloutier, line 233: sp: "on which other bounds safety checking" (rely?)

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,557 @@
...
+* `UnsafeContinuation`, `withUnsafe(Throwing)Continuation`: An unsafe form of `withChecked(Throwing)Continuation` that does not verify that the continuation is called exactly once, which can cause various safety problems.
+* `withUnsafeCurrentTask` and `UnsafeCurrentTask`: The `UnsafeCurrentTask` type does not provide lifetime safety, and must only be used within the closure passed to `withUnsafeCurrentTask`.
+* `UnownedSerialExecutor`: This type is intentionally not lifetime safe. It's primary use is the `unownedExecutor` property of the `Actor` protocol, which documents the lifetime assumptions of the `UnownedSerialExecutor` instance it produces.
+
+All of these APIs will be marked `@unsafe`. For all of the types that are `@unsafe`, any API that uses that type in its signature will also be marked `@unsafe`, such as `Array.withUnsafeBufferPointer`. Unless mentioned above, standard library APIs that do not have an unsafe type in their signature, but use unsafe constructs in their implementation, will be considered to be safe.
```
apple-fcloutier, line 243: This also includes typealiases, right? Like this:

```swift
@unsafe typealias PointerToInt = UnsafePointer<Int>
@unsafe typealias FunnyClosure = (UnsafePointer<Int>) -> Void
```

proposals/nnnn-strict-memory-safety.md
apple-fcloutier: It would be consistent that we don't need it in binding patterns given that we're exempting locals from `@unsafe` even when they have an unsafe type:

```swift
if case .pointer(let pointer) = foo() { // OK, locals are exempted from `@unsafe`
```

If you're matching against an expression, that expression hopefully needs `unsafe` if it contains unsafe code:

```swift
if case .some(unsafe getIntegerUnsafely()) = maybeWrappedInt() {
```

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,557 @@
...
+### Unsafe compiler flags
+
+There are a number of compiler flags that intentionally disable some safety-related checking. For each of these flags, the compiler will produce a diagnostic if they are used with strict memory safety:
+
+* `-Ounchecked`, which disables some checking in the standard library, including (for example) bounds checking on array accesses.
+* `-enforce-exclusivity=unchecked` and `-enforce-exclusivity=none`, which disables exclusivity checking that is needed for memory safety.
+* `-strict-concurrency=` for anything other than "complete", because the memory safety model requires strict concurrency to eliminate thread safety issues.
+* `-disable-access-control`, which allows one to break invariants of a type that can lead to memory-safety issues, such as breaking the invariant of `Range` that the lower bound not exceed the upper bound.
+
+### Unsafe overrides
```
apple-fcloutier, line 254: If you use an `if` or a `switch` as an expression, can you put `unsafe` ahead of it? And if so, what does it cover exactly?

```swift
let x = if foo() { unsafe bar() } else { baz() }
let y = unsafe if foo() { bar() } else { baz() } // allowed?
```

proposals/nnnn-strict-memory-safety.md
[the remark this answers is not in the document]
apple-fcloutier: It's fairly dense, but how do you feel about "memory safety preconditions for 'f()' may be unmet when used through safe conformance to 'P'"?

apple-fcloutier:
If you import a module that doesn't enable strict memory safety and it has a function that uses an UnsafePointer, will that function be implicitly `@unsafe` in a module that does enable strict memory safety?

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,557 @@
...
+}
+```
+
+The `unsafe` expression is much like `try` and `await`, in that it acknowledges that unsafe constructs (`wrapper`) are used within the subexpression but otherwise does not change the type. Unlike `try` and `await`, which require the enclosing context to handle throwing or be asynchronous, respectively, the `unsafe` expression does not imply any requirements about the enclosing block: it is purely a marker to indicate the presence of unsafe code, silencing a diagnostic.
+
+### Unsafe language constructs
+
+The following language constructs are always considered to be unsafe:
+
+* `unowned(unsafe)`: Used to store a reference without maintaining its reference count. The safe counterpart, `unowned`, uses dynamic checking to ensure that the reference isn't accessed after the corresponding object has been released. The `unsafe` variant disables that dynamic checking. Uses of `unowned(unsafe)` entities are not memory-safe.
```
Xazax-hun, line 219: Not sure what is the policy on mentioning constructs not officially part of the language yet. 

With the upcoming lifetimes work we have quite a few including:
* `@_unsafeNonescapableResult`
* `_overrideLifetime`

.. and probably more. @atrick probably has a more exhaustive list.

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,557 @@
...
+
+Swift provides memory safety with a combination of language affordances and runtime checking. However, Swift also deliberately includes some unsafe constructs, such as the `Unsafe` pointer types in the standard library, language features like `nonisolated(unsafe)`, and interoperability with unsafe languages like C. For most Swift developers, this is a pragmatic solution that provides an appropriate level of memory safety while not getting in the way.
+
+However, some projects want require stronger memory-safety guarantees than are provided Swift by default. These projects want to pay closer attention to uses of unsafe constructs in their code, and discourage casual use of unsafe constructs when a safe alternative exists. This proposal introduces opt-in strict memory safety checking to identify those places in Swift code that make use of unsafe language constructs and APIs. Any code written within this strictly-safe subset also works as “normal” Swift and can interoperate with existing Swift code.
```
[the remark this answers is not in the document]
ole: sp: "than are provided _by_ Swift by default" (or better: "than Swift provides by default" ?)

proposals/nnnn-strict-memory-safety.md
geoffreygaren: Are we sure that exempting local variables from unsafe annotation would not lose safety? Different locals can have different lifetimes, so each can carry a separate need to verify an outside-the-language precondition.

Example:

```
let lifetime1 = array.withUnsafeBufferPointer {
    let lifetime0 = $0
    ...
    return $0
}
```

lifetime0 and lifetime1 are very different here!

In my experience, tracking where an unsafe type travels in a program is one of the most difficult parts of auditing memory safety today in Swift (the other hardest part being knowing whether an interface is safe or unsafe). If travel can happen silently (especially through temporaries, magic variables like $0, capturing, and so on), auditing remains difficult.

geoffreygaren: Here's an example without closures, in case closures are off-topic:

```
if var x = unsafe unsafeGetX() {
    if let y = unsafeGetY() {
        ...
        x = y // Undocumented lifetime transfer, very hard to audit!
        ...
    else {
        ...
        x = condition() ? unsafe unsafeGetZ() : nil // Semi-undocumented lifetime transfer, maybe OK since I still said 'unsafe' nearby
        ...
    }
}
```

proposals/nnnn-strict-memory-safety.md
DougGregor: I think the only open hole here in safety is if we have an `@unsafe` case whose types aren't actually unsafe. Inventing an example here, but...

```swift
enum WeirdAddress {
  @unsafe case rawOffsetIntoGlobalArray(Int)
}
```

As the proposal stands, there is no place to put the `unsafe` in a pattern match like:

```swift
if case .rawOffsetIntoGlobalArray(let offset) = weirdAddress
```

We could create an unsafe pattern, e.g.,

```swift
if case unsafe .rawOffsetIntoGlobalArray(let offset) = weirdAddress
```

proposals/nnnn-strict-memory-safety.md
DougGregor: If we're assuming that `x` and `y` have unsafe types here, then the references to `x` and `y` (whether reading or assigning) will result in warnings about unsafe uses.

proposals/nnnn-strict-memory-safety.md
DougGregor: I've added a Future Directions section to capture this, but I want to think about it a little more before committing to a path.

proposals/nnnn-strict-memory-safety.md
DougGregor: It's effectively inferred to be `@unsafe` because its signature contains an unsafe type.

proposals/nnnn-strict-memory-safety.md
DougGregor: We do warn if it doesn't cover any code that's unsafe. If the closure had some unsafe types in its signature, then that warning wouldn't trigger.

proposals/nnnn-strict-memory-safety.md
DougGregor: Yes

proposals/nnnn-strict-memory-safety.md
DougGregor: You need to have `unsafe` on the inside, like we require with `try`

proposals/nnnn-strict-memory-safety.md
DougGregor: For underscored attributes, let's make the implementation Do The Right Thing. When those attributes get formalized, we should note that they are unsafe.

proposals/nnnn-strict-memory-safety.md
apple-fcloutier: I would say that it's safe to construct and unpack this case and the unsafety is when you use the index, for instance if you did `unsafe globalArray[unchecked: rawOffsetIntoGlobalArray]`.

apple-fcloutier: A logical conclusion is that the `unsafe` programming model we have, where we prefer to annotate the unsafety as close as possible to the load/store, is not practical to create indices that are unsafe even before we apply them to a collection. I'm comfortable with that even though we do have one instance of it today (the unsafe Range initializer that lets you have an upper bound below the lower bound) and I think that we will have a stronger story if we avoid creating more.

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,646 @@
...
+
+* A compiler flag `-strict-memory-safety` that enables warnings for all uses of unsafe constructs within a given module. All warnings will be in the diagnostic group `Unsafe`, enabling precise control over memory-safety-related warnings per [SE-0443](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0443-warning-control-flags.md). When strict memory safety is enabled, the `StrictMemorySafety` feature will be set: `#if hasFeature(StrictMemorySafety)` can be used to detect when Swift code is being compiled in this mode.
+* An attribute `@unsafe` that indicates that a declaration is unsafe to use. Such declarations may use unsafe constructs within their signatures. 
+* A corresponding attribute `@safe` that indicates that a declaration whose signature contains unsafe constructs is actually safe to use. For example, the `count` property on `Unsafe(Mutable)BufferPointer` has an unsafe type in its signature (`self`), but is actually safe to use because it is just producing an `Int` value. Marking the property with `@safe` means that its use will be treated as safe.
```
apple-fcloutier, line 48: We established that [referencing the unsafe pointer still requires you to use `unsafe`](https://github.com/swiftlang/swift-evolution/pull/2635#discussion_r1919021214), so I expect you would still need to write `unsafe buffer.count` because `buffer` is unsafe even if `count` isn't. I imagine that you could still use it to declare safe conformances to protocols?

apple-fcloutier: Speaking of that, we know that referring to instances of an `@unsafe` type needs `unsafe`, but do you need it when referencing the type? For instance:

```swift
@unsafe struct Foo: MyProto {
	static var bar: Int { get }
	
}

print(/*unsafe?*/ Foo.bar)
doStuff(/*unsafe?*/ MyProto.self)
```

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,646 @@
...
+    let pi = unsafe (baseAddress! + i)
+    let pj = unsafe (baseAddress! + j)
+    let tmp = unsafe pi.move()
+    unsafe pi.moveInitialize(from: pj, count: 1)
+    unsafe pj.initialize(to: tmp)
+  }
+}
+```
+
+Now, it is very likely that an `@unsafe` function is going to make use of other unsafe constructs, so we could choose to make `@unsafe` on a function acknowledge all uses of unsafe code within its definition. For example, this would mean that marking `swapAt` with `@unsafe` means that one need not have any `unsafe` expressions in its body:
```
apple-fcloutier, line 544: sp: "Now, it is very likely that an `@unsafe` function is going to make use of other unsafe constructs" repeats what the previous paragraph said

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,646 @@
...
+    unsafe withUnsafeBufferPointerSimplified { buffer in
+      unsafe c_library_sum_function(buffer.baseAddress, buffer.count, 0)
+    }
+  }
+}
+```
+
+This differs somewhat from the way in which throwing and asynchronous functions work. A function that has a `try` or `await` in the body needs to be `throws` or `async`, respectively. Essentially, the effect from the body has to also be reflected in the signature. With unsafe code, this could mean that having `unsafe` expressions in the function body requires you to either make the function `@unsafe` or use some other suppression mechanism to acknowledge that you are using unsafe constructs to provide a safe interface.
+
+There are several options for such a suppression mechanism. An attribute form, `@safe(unchecked)`, is described below as an alternative to the `unsafe` expression. Another approach would be to provide an `unsafe!` form the `unsafe` expression, which (like `try!`) acknowledges the effect but doesn't propagate that effect out to the function. For the `sum` function, it would be used as follows:
```
apple-fcloutier, line 582: FWIW, I think a split for `unsafe`/`unsafe!` is neat. In an `@unsafe` function, it lets you annotate clearly which operations whose unsafety you expect to fully encapsulate and which operations you rely on callers getting right for you. `@safe(unchecked)` blocks were unwieldy, but this is barely any additional ceremony.

apple-fcloutier: A counterpoint is that if you have multiple unsafe operations in the same expression, or if a single expression has multiple safety preconditions, it's easy to choose the wrong one:

```swift
extension UnsafeMutableBufferPointer {
  @unsafe public func swapAt(_ i: Element, _ j: Element) {
    guard i != j else { return }
    precondition(i >= 0 && j >= 0)
    precondition(unsafe i < endIndex && j < endIndex)

    // encapsulated: bounds checked
    let pi = unsafe! (baseAddress! + i) 

    // encapsulated: bounds checked
    let pj = unsafe! (baseAddress! + j)

    // propagate outwards: we cannot know whether the object is initialized
    let tmp = unsafe pi.move()

    // pi is guaranteed to be uninitialized at this location but pj is not guaranteed to be
    // initialized: this is subtle but it needs 'unsafe'
    unsafe pi.moveInitialize(from: pj, count: 1)

    // encapsulated: pj is guaranteed to be uninitialized after moveInitialize(from: pj)
    // and tmp is guaranteed to be initialized
    unsafe! pj.initialize(to: tmp)
  }
}
```

Since you can use `unsafe` and `unsafe!` interchangeably in a `@unsafe` function, there is no mechanism for the compiler to tell you that you had the wrong one and I expect that they would be frequently confused.

proposals/nnnn-strict-memory-safety.md
geoffreygaren: If you had a function like

```
func accumulateInto<T>(_: T.Type, /* some closure */) -> T {
    let result = T();
    /* invoke closure repeatedly and store into result */
    return result
}
```

and called it like

```
accumulateInto(UnsafeType.self) {
   ...
}
```

the call to `accumulateInto` would be your last / only place to say `unsafe` before instantiating and using UnsafeType. So I think it's probably needed?

proposals/nnnn-strict-memory-safety.md
geoffreygaren: I would add that there's a specific downside to `!` as a signifier in unsafe code. In `try!` and `optional!`, the `!` signifies "terminate execution if I was wrong". `unsafe` is the exact opposite -- it's there precisely because Swift cannot terminate execution.

proposals/nnnn-strict-memory-safety.md
```diff
@@ -0,0 +1,640 @@
...
+
+* Default arguments of functions are part of the implementation of a function, not its signature. For example, the following function does not have any unsafe types in its signature, so it does not require `@unsafe`, even though the default argument for `value` involves unsafe code. That unsafe code is effectively part of the body of the function, so it follows the rules for `unsafe` expressions.
+
+  ```swift
+  func hasDefault(value: Int = unsafe getIntegerUnsafely()) { ... }
+  ```
+
+### `@safe` attribute
+
+Like the `@unsafe` attribute, the `@safe` attribute ise used on declarations whose signatures involve unsafe types. However, the `@safe` attribute means that the declaration is consider safe to use even though its signature includes unsafe types. For example, marking `UnsafeBufferPointer` as `@unsafe` means that all operations involving an unsafe buffer pointer are implicitly considered `@unsafe`. The `@safe` attribute can be used to say that those particular operations are actually safe. For example, any operation involving buffer indices or count are safe, because they don't touch the memory itself. This can be indicated by marking these APIs `@safe`:
```
rjmccall:
```suggestion
Like the `@unsafe` attribute, the `@safe` attribute ise used on declarations whose signatures involve unsafe types. However, the `@safe` attribute means that the declaration is considered safe to use even though its signature includes unsafe types. For example, marking `UnsafeBufferPointer` as `@unsafe` means that all operations involving an unsafe buffer pointer are implicitly considered `@unsafe`. The `@safe` attribute can be used to say that particular operations are actually safe. For example, any operation involving buffer indices or count are safe, because they don't touch the memory itself. This can be indicated by marking these APIs `@safe`:
```

rjmccall (approved):

proposals/nnnn-strict-memory-safety.md
atrick: `_overrideLifetime` is annotated as `@unsafe`. That's the only one I would expect to formalize.

PR #2638, opened 2025-01-01
[DRAFT] Introduce @specialize attribute

airspeedswift:


proposals/NNNN-specialize.md
```diff
@@ -0,0 +1,193 @@
...
+
+Doing this restores the performance of the specialized version of `sum` (less a check and branch of the calling type) even when using the existential `any Summable` type.
+
+## Detailed design
+
+The `@specialize` attribute can be placed on any generic function. It takes an argument with the same syntax as a `where` clause for a generic signature. 
+
+The `where` clause must fully specialize all the generic placeholders of the types in the function signature. In the case of a protocol extension, as seen above, that includes specifying `Self`.
+
+It can also be used on extensions of generic types, on computed properties (note, it must be put on `get` explicitly, not on the shorthand where it is elided), and on free functions:
```
apple-fcloutier, line 104: This is clear from the example, but it could be worth calling out this also applies to protocol extensions.

proposals/NNNN-specialize.md
```diff
@@ -0,0 +1,193 @@
...
+func sum<T: BinaryInteger>(_ numbers: T...) -> Double {
+  numbers.reduce(0) { $0 + Double($1) }
+}
+```
+
+All placeholders must be fully specified in the `where` clause even if they do not appear to be used in the function body:
+
+```swift
+extension Dictionary where Value: BinaryInteger {
+  // error: Too few generic parameters are specified in '_specialize' attribute (got 1, but expected 2)
```
apple-fcloutier, line 129: Typo (old spelling): _specialize

proposals/NNNN-specialize.md
```diff
@@ -0,0 +1,193 @@
...
+  // note: Missing constraint for 'Key' in 'specialize' attribute
```
apple-fcloutier, line 130: Should the diagnostic say "missing equality constraint" to clarify what's expected?

proposals/NNNN-specialize.md
```diff
@@ -0,0 +1,193 @@
...
+
+### Making Specializations Publicly Available
+
+This proposal keeps expliicit specializations internal only. For ABI-stable frameworks, a useful future direction would be to make these symbols publicly available and listed in the `.swiftinterface` file, for callers to link to directly. 
+
+Doing this allows ABI-stable framework authors to expose specializations without exposing the full implementation details of a function as inlinable code. While adding a public specialized symbol to a framework is ABI, it is a much more limited ABI surface compared to providing an inlinable implementation, which requires any future changes to a type to consider the previous inlinable code's behavior to ensure it remains compatible forever in older binaries. A specialized entry point could be updated in a framework to fix a bug without recompiling the caller. Specializations in binary frameworks also have the benefit of avoiding duplication of code into the caller.
+
+### Tooling Directions
```
apple-fcloutier, line 183: There's been at least a handful of requests for some syntax to force the compiler to specialize generics or diagnose otherwise (rdar://127693760). Is this also considered for a future direction?

proposals/NNNN-specialize.md
```diff
@@ -0,0 +1,193 @@
...
+```
+
+If you call this function directly on an array of a specific integer type (e.g. `Array<Int>`):
+
+```
+let arrayOfInt: [Int] = // ...
+let result = arrayOfInt.sum()
+```
+
+in optimized builds, the Swift compiler it will generate a specialized version of `sum` for that type.
```
eeckstein, line 34: the Swift compiler will generate...

proposals/NNNN-specialize.md
```diff
@@ -0,0 +1,193 @@
...
+```
+let arrayOfInt: [Int] = // ...
+let result = arrayOfInt.sum()
+```
+
+in optimized builds, the Swift compiler it will generate a specialized version of `sum` for that type. 
+
+If you inspect the binary, you will see this specialized version under a symbol `_$sST3sumSz7ElementRpzrlEAASdyFSaySiG_Tg5`, which demangles to `generic specialization <[Swift.Int]> of (extension in example):Swift.Sequence< where A.Element: Swift.BinaryInteger>.sum() -> Swift.Double`, alongside the unspecialized version.
+
+This specialized version of `sum` will be optimized specifically for `Array<Int>`. It can move a pointer directly over the the array's buffer, loading the elements into registers directly from memory. On modern hardware, it can make use of dedicated instructions to convert the them to floating point. It can do this because it can see the implementation of `sum`, and can see the exact types on which it is being called. What exact assembly instructions are generated will differ significantly between e.g. `Array<Int8>.sum` and `Array<UInt64>.sum`.
```
eeckstein, line 38: ...to convert them...

proposals/NNNN-specialize.md
```diff
@@ -0,0 +1,193 @@
...
+
+Closely related to partial specialization is the ability to specialize once for all particular type layouts. In the `Dictionary.sum` example, it might only matter what the size of the key type is in order to efficiently iterate the values.
+
+Another example of this is `Array.append`. There is only one single implementation needed for appending an `AnyObject`-constrained to an array. The append operation needs to retain the object, but does not need any other properties. So two class instances could be appended using the same method irrespective of their actual type. Similar techniques could be used to provide shared specializations for `BitwiseCopyable` types, possibly with the addition of a supplied size argument to avoid having to use value witness calls.
+
+### Making Specializations Publicly Available
+
+This proposal keeps expliicit specializations internal only. For ABI-stable frameworks, a useful future direction would be to make these symbols publicly available and listed in the `.swiftinterface` file, for callers to link to directly.
```
aschwaighofer, line 179: Typo: expliicit

proposals/NNNN-specialize.md
```diff
@@ -0,0 +1,193 @@
...
+        ret
+```
+
+Now consider some code that places an optimization barrier between the call site and the concrete type:
+
+```swift
+protocol Summable: Sequence where Element: BinaryInteger { }
+extension Array: Summable where Element: BinaryInteger { }
+
+var summable: any Summable = []
```
eeckstein, line 65: -> `var summable: any Summable = Array<Int>()` to avoid the `error: type 'Any' cannot conform to 'BinaryInteger'`

proposals/NNNN-specialize.md
```diff
@@ -0,0 +1,193 @@
...
+}
+```
+
+## Source compatibility
+
+The addition or removal of explicit specializations has no impact on the caller, and is opt-in. As such it has no source compatability implications.
+
+## ABI compatibility
+
+This proposal covers only _internal_ specializations, dispatched to from within the implementation of an unspecialized generic function. As such it has no impact on ABI. It can be applied to existing ABI-stable functions in libraries built for distribution, and can be removed later without ABI impact. Generic functions can continue to be inlinable to be specialized by the caller, and the code to dispatch to specialized versions will not appear in the inlinable code emitted into the swift interface file.
```
eeckstein, line 157: Is there a use case for a function both being inlinable and having a specialize attribute?

eeckstein:
@al45tair recently brought up this feature request, which might be good to add in "future directions":

> `@_specialize` only appears to affect functions/methods. That makes sense from an implementation perspective, but it would be nice to be able to write `@_specialize` on a `class` or `struct` to get Swift to specialise *all* the methods that are affected by the generics on that `class` or `struct` without having to write `@_specialize` before every single one.

proposals/NNNN-specialize.md
airspeedswift:
```suggestion
This proposal keeps explicit specializations internal only. For ABI-stable frameworks, a useful future direction would be to make these symbols publicly available and listed in the `.swiftinterface` file, for callers to link to directly. 
```

proposals/NNNN-specialize.md
airspeedswift:
```suggestion
in optimized builds, the Swift compiler will generate a specialized version of `sum` for that type. 
```

proposals/NNNN-specialize.md
airspeedswift:
```suggestion
This specialized version of `sum` will be optimized specifically for `Array<Int>`. It can move a pointer directly over the the array's buffer, loading the elements into registers directly from memory. On modern hardware, it can make use of dedicated instructions to convert the integers to floating point. It can do this because it can see the implementation of `sum`, and can see the exact types on which it is being called. What exact assembly instructions are generated will differ significantly between e.g. `Array<Int8>.sum` and `Array<UInt64>.sum`.
```

proposals/NNNN-specialize.md
airspeedswift:
```suggestion
var summable: any Summable
```

proposals/NNNN-specialize.md
airspeedswift: Yes. It means the specialization is created for invocations via e.g. an existential, while still allowing _other_ types to be specialized from concrete call sites (especially types outside the module).

proposals/NNNN-specialize.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Implications on adoption
+
+Explicit specializations impose no new runtime requirements on either the caller or the called function, so use of them can be back-deployed to earlier Swift runtimes.
+
+## Future directions
+
+### Partial Specialization
+
+The need to fully specify every type when using `@specialize` is somewhat limiting. For example, in the `Dictionary` example above, no reference is made in the code to the `Key` type in summing the values. Having to explicitly state the concrete type for `Key` means you need separate specializations for `[String:Int]`, `[Int:Int]` and so on.
```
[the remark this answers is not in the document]
airspeedswift: Partial specialization of the element is probably more impactful, you'd still have to use the iterator but it's the use of the `BinaryInteger` protocol to go from `Int` to `Double` that's the most heartbreaking, performance-wise.

proposals/NNNN-specialize.md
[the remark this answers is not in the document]
airspeedswift:
```suggestion
This proposal keeps explicit specializations internal only. For ABI-stable frameworks, a useful future direction would be to make these symbols publicly available and listed in the `.swiftinterface` file, for callers to link to directly. 
```

airspeedswift:
> it would be nice to be able to write @_specialize on a class or struct
Hmm. There are lots of open questions here. For example, would that get applied to methods added outside the module? I guess it could. That's quite a subtle at-a-distance impact though.

proposals/NNNN-specialize.md
```diff
@@ -0,0 +1,193 @@
...
+
+Even in cases where dynamic dispatch through the protocol is still required for the unspecialized aspects (such as determining where to find the values in a dictionary's layout), this might not be in the hot part of the function and so partial specialization might be the better trade-off.
+
+Closely related to partial specialization is the ability to specialize once for all particular type layouts. In the `Dictionary.sum` example, it might only matter what the size of the key type is in order to efficiently iterate the values.
+
+Another example of this is `Array.append`. There is only one single implementation needed for appending an `AnyObject`-constrained to an array. The append operation needs to retain the object, but does not need any other properties. So two class instances could be appended using the same method irrespective of their actual type. Similar techniques could be used to provide shared specializations for `BitwiseCopyable` types, possibly with the addition of a supplied size argument to avoid having to use value witness calls.
+
+### Making Specializations Publicly Available
```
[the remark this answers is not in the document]
airspeedswift: Hmm. It really comes down to how "obvious" the current solution is. If there's more to debate here about how exactly it works, I'd prefer to just leave it as a future direction.

al45tair:
> Hmm. There are lots of open questions here. For example, would that get applied to methods added outside the module? I guess it could. That's quite a subtle at-a-distance impact though.

That's a great question. I'm not sure whether that's desirable; maybe it should only apply within the defining module, and if someone wants to extend the class and have things specialised they should have to write `@_specialize` on their extension?

stephentyrone (approved):

PR #2639, opened 2025-01-02
[DRAFT] Last value return proposal

airspeedswift:
Adds a rule that the last expression in a function is the return value, and introduces multi-statement `if`, `switch`, and `do` expressions.

### DIFF
--- a/proposals/NNNN-last-value-return.md
+++ b/proposals/NNNN-last-value-return.md
@@ -1,0 +1,227 @@
+# Last-value rule for return values and expressions
+
+* Proposal: [SE-NNNN](NNNN-last-value-expressions.md)
+* Authors: [Ben Cohen](https://github.com/airspeedswift), [Hamish Knight](https://github.com/hamishknight)
+* Review Manager: TBD
+* Status: **Awaiting Review**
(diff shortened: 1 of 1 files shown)


proposals/NNNN-last-value-return.md
```diff
@@ -0,0 +1,227 @@
...
+
+The introduction of a `then` keyword would set `if` expressions apart from functions. The presence of a `then` keyword in an `if` expression that was part of a single-expression function body would act _like_ a `return` but not exactly, causing cognitive load on the part of the reader.
+
+Overall, the composable consistency between `if`/`switch`/`do` expressions, and function bodies, along with the "`return` means exiting _early_" benefits, weighs in favor of a last expression rule and against a `then` (or similar) keyword.
+
+## Source compatibility
+
+As discussed in detailed design, there are rare edge cases where this new rule may break source, but none have been found in the compatibility test suite. Where they do occur, backticks can be applied, and this fix will back deploy to earlier compiler versions.
```
benrimmington, line 222:
```suggestion
As discussed in "Impact on existing code" section, there are edge cases where this new rule may break source.
```

PR #2644, opened 2025-01-04
Update 0414-region-based-isolation.md

barbaramartina:
# Clarification for one of the examples
Updating example to clarify func 'caller' and 'callee' (callee was used twice)


rjmccall:
Thanks!

PR #2645, opened 2025-01-10
SE-0454: Update status

al45tair:
Update the status to "In Review"


apple-fcloutier:
It made me jump to see a proposal called "memory allocator". I'm not concerned if it's just for the Windows toolchain, but I wish that had been reflected in the file name!

al45tair:
> It made me jump to see a proposal called "memory allocator". I'm not concerned if it's just for the Windows toolchain, but I wish that had been reflected in the file name!

@apple-fcloutier Yeah, it won't ever affect Darwin; there's no way I'd let that through any kind of review — the system allocator is the right choice there.

PR #2654, opened 2025-01-16
Custom main and global executors

al45tair:
Add write up of custom executor proposals. There are still some concerns surrounding integration with Dispatch that need working through.

### DIFF
--- a/proposals/nnnn-custom-main-and-global-executors.md
+++ b/proposals/nnnn-custom-main-and-global-executors.md
@@ -293,11 +293,10 @@ public protocol RunLoopExecutor: Executor {
 }
 ```
 
-We will also add a protocol for the main actor's executor (see later
-for details of `EventableExecutor` and why it exists):
+We will also add a protocol for the main actor's executor:
 
 ```swift
-protocol MainExecutor: RunLoopExecutor & SerialExecutor & EventableExecutor {
+protocol MainExecutor: RunLoopExecutor & SerialExecutor {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


proposals/nnnn-custom-main-and-global-executors.md
```diff
@@ -0,0 +1,873 @@
...
+extension Task {
+  /// Get the current executor; this is the executor that the currently
+  /// executing task is executing on.
+  ///
+  /// This will return, in order of preference:
+  ///
+  ///   1. The custom executor associated with an `Actor` on which we are
+  ///      currently running, or
+  ///   2. The preferred executor for the currently executing `Task`, or
+  ///   3. The task executor for the current thread
```
FranzBusch, line 358: I think this should be "for the current task" instead of "current thread"

proposals/nnnn-custom-main-and-global-executors.md
```diff
@@ -0,0 +1,873 @@
...
+```swift
+protocol Executor {
+  ...
+  /// Return this executable as a SchedulableExecutor, or nil if that is
+  /// unsupported.
+  ///
+  /// Executors can implement this method explicitly to avoid the use of
+  /// a potentially expensive runtime cast.
+  @available(SwiftStdlib 6.2, *)
+  var asSchedulable: AsSchedulable? { get }
```
FranzBusch, line 509:
```suggestion
  var asSchedulableExecutor: SchedulableExecutor? { get }
```

proposals/nnnn-custom-main-and-global-executors.md
al45tair: I'm not sure there's a practical difference there; the task executors are tracked using thread local data.

FranzBusch: Is that true? Aren't they tracked using task local data?

al45tair: It's this code here that does the tracking: https://github.com/swiftlang/swift/blob/5a6a14785075436b8bdb2c423cd47df4945d3ec8/stdlib/public/Concurrency/Actor.cpp#L114

PR #2668, opened 2025-01-27
[SE-0419] Tweak proposal slightly.

al45tair:
Added an `ImageMap` type to hold the list of loaded images.  This is neater and allows for some optimizations as well as make it easier to support `Codable` for `Backtrace`.  This also removes the `captureImages()` API in favour of `ImageMap.capture()`, which is more natural.

Also exposed the ability to construct a `Backtrace.Address` from a `FixedWidthInteger`, write-protected `Backtrace.architecture`, and switch `SymbolicatedBacktrace` back to an explicit `Array` of `Frame`s as we aren't trying to minimise space for the symbolicated version.


al45tair:
See also https://github.com/swiftlang/swift/pull/78516

PR #2675, opened 2025-01-30
[SE-NNNN] Method and Initializer Key Paths

amritpan:
This is the proposal text for method and initializer key paths, and is ready for review.

### DIFF
--- a/proposals/NNNN-method-and-initializer-keypaths.md
+++ b/proposals/NNNN-method-and-initializer-keypaths.md
@@ -1,0 +1,165 @@
+# Method and Initializer Key Paths
+
+* Proposal: [SE-NNNN](NNNN-filename.md)
+* Authors: [Amritpan Kaur](https://github.com/amritpan), [Pavel Yaskevich](https://github.com/xedin)
+* Review Manager: TBD
+* Status: **Awaiting review**
+* Implementation: [swiftlang/swift#78823](https://github.com/swiftlang/swift/pull/78823) [swiftsyntax/swiftsyntax#2950](https://github.com/swiftlang/swift-syntax/pull/2950) 
+* Upcoming Feature Flag: `KeyPathWithMethodMembers`
+* Review: ([pitch](https://forums.swift.org/t/pitch-method-key-paths/76678))
+
+## Introduction
+
+Swift key paths can be written to properties and subscripts. This proposal extends key path usage to include references to method members, such as instance and type methods, and initializers.
+
+## Motivation
+
+Key paths to method members and their advantages have been explored in several discussions on the Swift forum, specifically to [unapplied instance methods](https://forums.swift.org/t/allow-key-paths-to-reference-unapplied-instance-methods/35582) and to [partially and applied methods](https://forums.swift.org/t/pitch-allow-keypaths-to-represent-functions/67630). Extending key paths to include reference to methods and initializers and handling them similarly to properties and subscripts will unify instance and type member access for a more consistent API. Key path methods and initializers will also enjoy all of the benefits offered by existing key path component kinds, e.g. simplify code by abstracting away details of how these properties/subscripts/methods are modified/accessed/invoked, reusability via generic functions that accept key paths to methods as parameters, and supporting dynamic invocation while maintaining type safety.
+
+## Proposed solution
+
+We propose the following usage:
+
+```swift
+struct Calculator {
+  func square(of number: Int) -> Int {
+    return number * number * multiplier
+  }
+
+  func cube(of number: Int) -> Int {
+    return number * number * number * multiplier
+  }
+
+  init(multiplier: Int) {
+    self.multiplier = multiplier
+  }
+
+  let multiplier: Int
+}
+
+// Key paths to Calculator methods
+let squareKeyPath = \Calculator.square
+let cubeKeyPath = \Calculator.cube
+```
+
+These key paths can then be invoked dynamically with a generic function:
+
+```swift
+func invoke<T, U>(object: T, keyPath: KeyPath<T, (U) -> U>, param: U) -> U {
+  return object[keyPath: keyPath](param)
+}
+
+let calc = Calculator(multiplier: 2)
+
+let squareResult = invoke(object: calc, keyPath: squareKeyPath, param: 3)
+let cubeResult = invoke(object: calc, keyPath: cubeKeyPath, param: 3)
+```
+
+Or used to dynamically create a new instance of Calculator:
+
+```swift
+let initializerKeyPath = \Calculator.Type.init(multiplier: 5)
+```
+
+This proposed feature homogenizes the treatment of member declarations by extending the expressive power of key paths to method and initializer members.
+
+## Detailed design
+
+Key path expressions can refer to instance methods, type methods and initializers, and imitate the syntax of non-key path member references.
+
+### Argument Application
+
+Key paths can reference methods in two forms:
+
+1. Without argument application: The key path represents the unapplied method signature.
+2. With argument application: The key path references the method with arguments already applied.
+
+Continuing our `Calculator` example, we can write either:
+
+```swift
+let squareWithoutArgs: KeyPath<Calculator, (Int) -> Int> = \Calculator.square
+let squareWithArgs: KeyPath<Calculator, Int> = \Calculator.square(of: 3)
+```
+
+If the member is a metatype (e.g., a static method, class method, initializer, or when referring to the type of an instance), you must explicitly include `.Type` in the key path root type.
+
+```swift
+struct Calculator {
+  func add(_ a: Int, _ b: Int) -> Int {
+    return a + b
+  }
+}
+
+let calc = Calculator.self
+let addKeyPath: KeyPath<Calculator.Type, (Calculator) -> (Int, Int) -> Int> = \Calculator.Type.add
+```
+
+Here, `addKeyPath` is a key path that references the add method of `Calculator` as a metatype member. The key path’s root type is `Calculator.Type`, and it resolves to an unapplied instance method: `(Calculator) -> (Int, Int) -> Int`. This represents a curried function where the first step binds an instance of `Calculator`, and the second step applies the method arguments.
+
+```swift
+let addFunction = calc[keyPath: addKeyPath]
+let fullyApplied = addFunction(Calculator())(20, 30)`
+```
+
+`addFunction` applies an instance of Calculator to the key path method. `fullyApplied` further applies the arguments (20, 30) to produce the final result.
+
+### Overloads
+
+Keypaths to methods with the same base name and distinct argument labels can be disambiguated by explicitly including the argument labels:
+
+```swift
+struct Calculator {
+  var subtract: (Int, Int) -> Int { return { $0 + $1 } }
+  func subtract(this: Int) -> Int { this + this}
+  func subtract(that: Int) -> Int { that + that }
+}
+  
+let kp1 = \S.subtract // KeyPath<S, (Int, Int) -> Int
+let kp2 = \S.subtract(this:) // WritableKeyPath<S, (Int) -> Int>
+let kp3 = \S.subtract(that:) // WritableKeyPath<S, (Int) -> Int>
+let kp4 = \S.subtract(that: 1) // WritableKeyPath<S, Int>
+```
+
+### Dynamic member lookups
+
+`@dynamicMemberLookup` can resolve method references through key paths, allowing methods to be accessed dynamically without explicit function calls:
+
+```swift
+@dynamicMemberLookup
+struct DynamicKeyPathWrapper<Root> {
+    var root: Root
+
+    subscript<Member>(dynamicMember keyPath: KeyPath<Root, Member>) -> Member {
+        root[keyPath: keyPath]
+    }
+}
+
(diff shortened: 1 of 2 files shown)


proposals/NNNN-method-and-initializer-keypaths.md
```diff
@@ -0,0 +1,165 @@
...
+```swift
+struct Calculator {
+  func add(_ a: Int, _ b: Int) -> Int {
+    return a + b
+  }
+}
+
+let calc = Calculator.self
+let addKeyPath: KeyPath<Calculator.Type, (Calculator) -> (Int, Int) -> Int> = \Calculator.Type.add
+```
```
beccadax, line 95: I found this example confusing because you're using an unapplied instance method, which is a very niche feature (and one that member key paths are meant to supplant!). Instead, I might illustrate this point with an example involving a static method or initializer—which I bet will be more common use cases—and then mention that it also works with unapplied instance methods.

proposals/NNNN-method-and-initializer-keypaths.md
```diff
@@ -0,0 +1,165 @@
...
+Component chaining between methods or from method to other key path types is also supported with this feature and will continue to behave as `Hashable`/`Equatable` types.
+
+```swift
+let kp5 = \Calculator.subtract(this: 1).signum()  
+let kp6 = \Calculator.subtract(this: 2).description
+```
+
+## Source compatibility
+
+This feature has no effect on source compatibility.
```
beccadax, line 157: How certain are we that expanding `@dynamicMemberLookup` to methods won't cause source compatibility breaks in common libraries? Is this something we've tested? Should we discuss an alternative where, for instance, you need to say `@dynamicMemberLookup(all)` if you want to support method lookup?

(This wasn't really a concern with metatype keypaths because the `subscript(dynamicMember:)` would have to accept `KeyPath<Foo.Type, T>` instead of `KeyPath<Foo, T>` for the new members to affect it.)

proposals/NNNN-method-and-initializer-keypaths.md
```diff
@@ -0,0 +1,165 @@
...
+
+## ABI compatibility
+
+This feature does not affect ABI compatibility.
+
+## Implications on adoption
+
+This feature has no implications on adoption.
```
beccadax, line 165: This would be a good place to discuss whether you expect this feature to back-deploy and what constraints there might be on doing so.

In particular, in SE-0438, there were some subtleties around symbols that we needed to start emitting. Are there similar concerns here? How are they being resolved? (If things are shaking out in basically the same way they did in the metatype keypath proposal, you could quickly summarize and link back to it.)

proposals/NNNN-method-and-initializer-keypaths.md
```diff
@@ -0,0 +1,165 @@
...
+}
+
+let dynamicCalculator = DynamicKeyPathWrapper(root: Calculator())
+let subtract = dynamicCalculator.subtract
+print(subtract(10))
+```
+
+### Effectful value types
+
+Methods annotated with `nonisolated` and `consuming` are supported by this feature. `mutating`, `throwing` and `async` are not supported for any other component type and will similarly not be supported for methods. Keypaths cannot capture method arguments that are not `Hashable`/`Equatable`, so `escaping` is also not supported.
```
beccadax, line 144: A few questions:

* How does `consuming` work? (Maybe there's a statement that ought to be here about whether key path roots and/or arguments can be noncopying?)
* `throwing` should probably be spelled `throws`.
* Perhaps instead of singling out `escaping`, you should simply state that keypaths cannot capture closure arguments? (I assume this is the case?)

beccadax (changes requested):
Excited about this proposal, but there are a few suggestions I'd like to make.

beccadax:
Oh, one more question that isn't addressed in the proposal: Do these key path literals support the implicit closure coercion like other key path literals do? The ability to say `fooArray.map(\.foo(bar: baz))` might be worth highlighting!

proposals/NNNN-method-and-initializer-keypaths.md
amritpan: This makes sense - I will update this example

proposals/NNNN-method-and-initializer-keypaths.md
amritpan: > How certain are we that expanding @dynamicMemberLookup to methods won't cause source compatibility breaks in common libraries?

I hadn't considered this! Also to make sure I fully understand - are we talking about being able to differentiate between a static and an instance method via `@dynamicMemberLookup`? I have added [these passing tests to `Interpreter/keypath.swift`](https://github.com/swiftlang/swift/blob/a34bdb49557d6546dded5407e83cffb83268fc2e/test/Interpreter/keypath.swift#L352) for `@dynamicMemberLookup` but it doesn't test across libraries. How can I test for this?

proposals/NNNN-method-and-initializer-keypaths.md
amritpan: I initially assumed that this feature would have the same implications on back-deployment as SE-0438. But then in the pitch discussion, it was [suggested](https://forums.swift.org/t/pitch-method-key-paths/76678/7) that methods do not emit descriptors. I also did not run into any linking errors with my Interpreter tests which is what I used to detect missing descriptors for metatype keypaths. 

In the SE-0438 implementation I also added handling to account for nil descriptors to KeyPath in the Standard Library and so between that and not emitting any descriptors with this implementation, I made the assumption that this is a non issue and this feature would have no constraints on back-deployment.

proposals/NNNN-method-and-initializer-keypaths.md
amritpan: I have updated this section with your suggestions.

amritpan:
Yes, implicit closure coercion is supported with this feature and I have added a section to note this!

proposals/NNNN-method-and-initializer-keypaths.md
beccadax: My concern is that suddenly having a lot of extra methods on types that use `@dynamicMemberLookup` will cause problems for real-world codebases. The fact that dynamic members are disfavored by the type checker should reduce source compatibility concerns, but it might not be the end of the story.

I suppose there are two pieces to this question:

1. **Does this change cause projects to actually fail to build?** You can assess this by asking Swift CI to run a [source compatibility test](https://github.com/swiftlang/swift/blob/main/docs/ContinuousIntegration.md#source-compatibility-testing) on your PRs; Swift CI will try to build a bunch of open source projects using a compiler with your changes and will report any projects that fail.

2. **How well does this change work for types that have already adopted `@dynamicMemberLookup`?** That's something you'd have to assess manually, by looking at popular libraries with `@dynamicMemberLookup` types, figuring out how much new API surface this exposes, and thinking about whether that new surface is bad. (To be clear, this is a subjective call!)

proposals/NNNN-method-and-initializer-keypaths.md
beccadax: IIRC, these descriptors are used to support `Equatable` on key paths. If the descriptors are always null, does that mean the `Equatable` conformance will behave strangely?

proposals/NNNN-method-and-initializer-keypaths.md
```diff
@@ -0,0 +1,172 @@
...
+struct Calculator {
+  var subtract: (Int, Int) -> Int { return { $0 + $1 } }
+  func subtract(this: Int) -> Int { this + this}
+  func subtract(that: Int) -> Int { that + that }
+}
+  
+let kp1 = \S.subtract // KeyPath<S, (Int, Int) -> Int
+let kp2 = \S.subtract(this:) // WritableKeyPath<S, (Int) -> Int>
+let kp3 = \S.subtract(that:) // WritableKeyPath<S, (Int) -> Int>
+let kp4 = \S.subtract(that: 1) // WritableKeyPath<S, Int>
```
beccadax, line 112: Just noticed: Would these be `WritableKeyPath` or just `KeyPath`?

proposals/NNNN-method-and-initializer-keypaths.md
amritpan: [Based on this code](https://github.com/swiftlang/swift/blob/38c8cc8ff45f7d02711140db5956ce20fbdb5a3b/stdlib/public/core/KeyPath.swift#L260), `Keypath` no longer uses the descriptors to determine Equatable/Hashable and instead relies on the component kind and component type. To the best of my understanding, the descriptors are only used to reference computed/external properties.

proposals/NNNN-method-and-initializer-keypaths.md
amritpan: Yes, this should be just `KeyPath` - thank you for catching this!

proposals/NNNN-method-and-initializer-keypaths.md
amritpan: Could you show me an example of the problem that you're describing? I'm worried that I may not have fully grasped the issues here. 

Here is what I think I understand: The typechecker disfavors the dynamic members over the real instance members. However, with this feature we might still run into a situation where a dynamicMemberLookup might attempt to resolve an unknown and unexpected resolve to a method? For example, if we have this type using `@dynamicMemberLookup`: 

```
@dynamicMemberLookup
struct DynamicWrapper {
    private var storage: [String: Any]

    init(_ storage: [String: Any]) {
        self.storage = storage
    }

    subscript(dynamicMember member: String) -> Any? {
        return storage[member]
    }
}

let json = DynamicWrapper(["fruit": "Mandarin", "qty": 30])
```

And then we have this keypath to a method: 

```
struct Food {
    func getName() -> String {
        return "Grapefruit!"
    }
}

let kpName = \Food.getName
```

Now before this feature, we could write: 

```
let jsonLookup = DynamicWrapper(["fruit": "Clementine"])
print(json.getName)
```

and it would fall back to Clementine. But now with implicit keypath methods, like:

`let kpNameDynamic = \DynamicWrapper.getName`

we'll prioritize the real method over the dynamic method (Grapefruit, when applied), which is expected. However, this could break projects that rely on dynamic lookups for method like names.

amritpan: I have been trying to run the source compatibility tests on the PR but they are currently failing due to unrelated issues (latest [build](https://ci.swift.org/job/swift-PR-source-compat-suite-macos/2284/), [release](https://ci.swift.org/job/swift-PR-source-compat-suite-debug-macos/2264/) results) and it seems that they are broken in general. I've run them a few times last week and some of the unrelated broken tests have resolved but not all, so I'll keep tracking that and re-run them again. The tests are not showing any keypath or dynamicMemberLookup related test failures but I don't know if I should trust them in their current state.

> How well does this change work for types that have already adopted @dynamicMemberLookup? That's something you'd have to assess manually, by looking at popular libraries with @dynamicMemberLookup types, figuring out how much new API surface this exposes, and thinking about whether that new surface is bad. (To be clear, this is a subjective call!)

I wonder if this is a question that we can raise during the review to get input from library developers who might have more experience with this.

proposals/NNNN-method-and-initializer-keypaths.md
beccadax: I think the crucial thing here is [the `func ==` for `KeyPathComponent`](https://github.com/swiftlang/swift/blob/38c8cc8ff45f7d02711140db5956ce20fbdb5a3b/stdlib/public/core/KeyPath.swift#L889), which does check the `id` fields inside of `get` components (which I believe is what you're using for method keypaths?).

Looking at your branch, it looks like you're generating the ID for methods by [passing a `SILDeclRef` for `storage` to `getFunction()`](https://github.com/swiftlang/swift/blob/7d30aac7cc7ba29b8f0747fa657b6e3762b6acca/lib/SILGen/SILGenExpr.cpp#L4362-L4380). In other words, the computed property ID for a method is the pointer to the function itself. I *think* that will work okay since you're forming the `SILDeclRef` directly from the `AbstractFunctionDecl` without offering an opportunity for it to end up pointing to, say, a thunk, although I worry a little about edge cases involving resilience. (Might comment more on that in the implementation PR.)

In any case, it looks like you have a more or less reasonable computed property ID and so therefore you should be able to get good equality behavior.

proposals/NNNN-method-and-initializer-keypaths.md
beccadax: > I have been trying to run the source compatibility tests on the PR but they are currently failing due to unrelated issues (latest [build](https://ci.swift.org/job/swift-PR-source-compat-suite-macos/2284/), [release](https://ci.swift.org/job/swift-PR-source-compat-suite-debug-macos/2264/) results) and it seems that they are broken in general.

Looks like the debug one failed to build LLVM (obviously unrelated) and the release one had one unexpected pass (probably some recently-merged PR having fixed a bug). I think we can call that a pass on source compatibility.

> I wonder if this is a question that we can raise during the review to get input from library developers who might have more experience with this.

Okay.

proposals/NNNN-method-and-initializer-keypaths.md
amritpan: I will wait for your comments on the PR before merging! The swift-syntax and foundations PRs have been merged.

beccadax (approved):

beccadax (approved):

PR #2676, opened 2025-02-02
Replaceable Library Plugins

tayloraswift:


proposals/0461-replaceable-library-plugins.md
```diff
@@ -0,0 +1,156 @@
+# Replaceable Library Plugins
+
+* Proposal: [SE-0461](0641-replaceable-library-plugins.md)
```
allevato, line 3: Process note: Please leave the SE- number unassigned (NNNN). It will be assigned the next number in sequence by the review manager at the time the review is started.

PR #2682, opened 2025-02-05
[SE-0453] Rename Vector to InlineArray

Azoy:
😢

### DIFF
--- a/proposals/0453-inlinearray.md
+++ b/proposals/0453-inlinearray.md
@@ -1,16 +1,16 @@
-# Vector, a fixed-size array
+# InlineArray, a fixed-size array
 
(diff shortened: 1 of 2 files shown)


Jumhyn:
I think, somewhat unfortunately, we ought to keep the file name here as "-vector.md" since we would be breaking links otherwise (unless GH will auto-redirect those?). Relatedly, might be worth a clarification somewhere that this was originally proposed as `Vector`. Also seems like there's a few spots where links have been modified that should probably stay as-is?

apple-fcloutier:
(This could possibly be addressed with a soft link, which Git can track, although it might not be worth doing another iteration for)

Jumhyn (approved):

PR #2686, opened 2025-02-06
[SE-0460] Fix metadata of `@specialized` proposal

benrimmington:
The proposal isn't listed on the [dashboard](https://www.swift.org/swift-evolution/) because there's an error message: "Missing Review field."

```sh
curl https://download.swift.org/swift-evolution/v1/evolution.json
```

Cc: @stephentyrone @DougGregor


DougGregor (approved):
Thanks!

PR #2687, opened 2025-02-10
[SE-0430] Remove borrowing sending example and update ownership rules

GangWoon:
The combination of borrowing sending has been partially disallowed, and from a [Swift developer's perspective](https://forums.swift.org/t/accepted-with-modifications-se-0430-second-review-sendable-parameter-and-result-values/71850), it is no longer compilable.
To avoid confusion for readers (including myself! 😊), I believe it would be best to remove references to it from the document.

Please let me know if my suggestion seems off or if there are any additional changes that should be made!


rjmccall:
Thanks!

PR #2690, opened 2025-02-12
[SE-0453] `InlineArray` proposal is accepted

benrimmington:
* Update status and review fields.
* Fix example code and doc comment.

Cc: @Azoy @Jumhyn

PR #2692, opened 2025-02-14
Mark SE-0453 as accepted

ahoppen:
SE-0453 was accepted in https://forums.swift.org/t/accepted-with-modifications-se-0453-inlinearray-formerly-vector-a-fixed-size-array/77678 but the proposal wasn’t updated.


j-f1:
Dupe of https://github.com/swiftlang/swift-evolution/pull/2690?

PR #2693, opened 2025-02-15
Propose the `borrow` operator

rjmccall:


proposals/0000-borrow-operator.md
```diff
@@ -0,0 +1,599 @@
...
+justly wanted the modifier to have only local effects.
+
+## Proposed solution
+
+This proposal adds a new `borrow` operator that explicitly requires a
+value to be borrowed for a particular use.
+
+```swift
+// Force the value of c to be borrowed. If c is somehow modified during
+// this call, there may be an error at runtime.
```
tbkka, line 146: "error at runtime" here implies that borrows carry runtime costs that do not apply to similar copies.  Since most people's mental model seems to be that "borrow" is "pointer to" with no other overhead, that cost likely needs to be called out a little more explicitly.

proposals/0000-borrow-operator.md
```diff
@@ -0,0 +1,599 @@
...
+For the rest of this proposal, the word *value* by itself should be understood
+to mean this low-level concept in which copying a value produces a new value.
+
+Every use of a value in Swift has some interaction with ownership.
+Non-mutating uses of values are always either:
+
+- *borrows*, which just read the low-level information in the value, or
+- *copies*, which create a new, independent copy of that low-level
+  information according to the rules of the type.
+
```
tbkka, line 85: I wonder if there should be a footnote here that points out that "borrow" does not necessarily mean "pointer to."  For small items, it's more efficient to "bitwise borrow" by (temporarily) copying the contents.

proposals/0000-borrow-operator.md
tbkka: Ah.  I see you've explored that in the borrows-by-default section at the end.

PR #2695, opened 2025-02-18
[SE-0456] post review updates

glessard:
Updates the status of the proposal, calls the properties `span`, refers to `InlineArray` by its accepted name.


glessard:
Commit 58dbf76 in this PR may or may not be necessary or advisable. I'd appreciate feedback on that before this is merged.

rjmccall (approved):

PR #2696, opened 2025-02-18
Mark SE-0443 as implemented in Swift 6.1.

allevato:
It looks like the implementation PR is in the Swift 6.1 snapshots, assuming there's nothing else outstanding.


allevato:
@DougGregor @DmT021 Is this accurate?

PR #2699, opened 2025-02-19
Proposal for `#SwiftSettings`.

gottesmm:
Just an initial draft for feedback


apple-fcloutier:
Some compiler settings have a security impact and I don't expect the number of these to go down over time. As we get a new way to change settings, it's spreading the source of truth for what settings are or aren't enabled in any given project. Do we enumerate the ways compiler settings can be modified in an Xcode project or in a SwiftPM project anywhere?

hborla:
Please leave design feedback in the forthcoming pitch thread on the forums - it will be posted soon - and leave editorial feedback on this PR.

proposals/0000-compiler-settings.md
```diff
@@ -0,0 +1,279 @@
...
+}
+```
+
+In order to ensure that compiler settings are easy to find in the file, we
+require that `compilerSettings` be the first AST node in the file before any
+other statements including import statements.
+
+After parsing, the compiler in the type checker validates that each element is a
+`CompilerSetting` and then updates the relevant compiler state as appropriate
+for the command line option.
```
hborla, line 212: You don't need to mention parsing, type checking, compiler state, etc, here - those are compiler implementation details. I think all we need to say is that the argument to `compilerSetting` has type `[CompilerSetting]` and those settings are applied only to the file they're specified in.

proposals/0000-compiler-settings.md
```diff
@@ -0,0 +1,279 @@
...
+
+We do not allow for `compilerSettings` to be exposed in textual modules. Any
+impacts of `compilerSettings` on the textual module must occur by the compiler
+setting resulting in changing the output of the textual interface in a manner
+that would not create inconsistencies between files in the same module. As a
+result of this, we restrict the set of upcoming and experimental features that
+can be passed to `compilerSettings` to features whose affect can be constrained
+in such a manner. This necessarily implies an opt-in process that will be
+specified in the compiler's internal implementation. Some examples of command
```
hborla, line 221: Again, don't mention compiler implementation details. It's okay to just say that only flags that can actually be applied per file will be supported, which is a subset of the build settings that programmers can apply.

proposals/0000-compiler-settings.md
```diff
@@ -0,0 +1,279 @@
...
+
+We will update the swift grammer to allow for a new top level statement called
+`compilerSettings` that takes a list of expressions:
+
+```text
+top-level-statement ::= 'compilerSettings' '[' (expr ',')* expr ','? ']'
+```
+
+During typechecking, we will validate that each of the expressions are instances
+of the enum `CompilerSetting`. `CompilerSetting` will be a resilient enum
```
FranzBusch, line 152: How is this going to work in modules that support multiple compilers at once e.g. most packages in the ecosystem. Can we come up with a spelling that makes it easy to adopt new cases while not needing a compiler guard syntax around it? Does a compiler guard around it work even?

proposals/0000-compiler-settings.md
```diff
@@ -0,0 +1,279 @@
...
+defined in the standard library that provides the following API:
+
+```
+public enum CompilerSetting {
+case enableUpcomingFeature(UpcomingFeatureCompilerSetting)
+case enableExperimentalFeature(ExperimentalFeatureCompilerSetting)
+case warningAsError(WarningAsErrorCompilerSetting)
```
FranzBusch, line 159: This is interesting. So far everyone defines this on their CI. If you start defining this on a per file basis it might be that a compiler introduces a new warning and suddenly your dependency fails to build since they added this to their per-file compiler setting.

etcwilde:
For folks who landed here and are looking for the pitch to comment on: https://forums.swift.org/t/pitch-compilersettings-a-top-level-statement-for-enabling-compiler-flags-locally-in-a-specific-file/77994

proposals/0000-compiler-settings.md
```diff
@@ -0,0 +1,284 @@
...
+2. A programmer can specify options on a per file basis instead of only on a per
+   module basis allowing for warningsAsError to be used per file instead of per
+   module migration to use new features.
+
+As an additional benefit since enums are being used, code completion can make
+the compiler options discoverable to the user in the IDE.
+
+## Detailed design
+
+We will update the swift grammer to allow for a new top level statement called
```
slavapestov, line 148: grammar

proposals/0000-compiler-settings.md
```diff
@@ -0,0 +1,284 @@
...
+1. Enabling warnings on the entire module by enabling upcoming features from the
+   command line.
+
+2. Updating the module incrementally until all of the warnings have been
+   eliminated from the module.
+
+3. Updating the language version of the module so from that point on warnings
+   will become errors to prevent backsliding.
+
+This creates developer friction when applied to large modules since the size of
```
slavapestov, line 88: A general editorial comment: there are a number of long sentences in this proposal that would benefit from commas. For example, in the above, you probably want to insert a comma before "since":

> This creates developer friction when applied to large modules, since the size of the module makes it difficult to eliminate all of the warnings at once.

proposals/0000-compiler-settings.md
```diff
@@ -0,0 +1,284 @@
...
+
+We purposely make `CompilerSetting` resilient so additional kinds of compiler
+settings can be added over time without breaking ABI.
+
+`CompilerSetting` provides an enum case for each command line option and each
+case is able to provide associated values to specify arguments that would
+normally be passed on the command line to the option. These associated values
+can be one of:
+
+* an already existing type (e.x.: `Actor.Type`)
```
slavapestov, line 178: I think `e.g.` is a more common abbreviation than `e.x.`, but IMO in technical writing it's better to spell out "for example"

proposals/0000-compiler-settings.md
```diff
@@ -0,0 +1,284 @@
...
+
+  // Cases synthesized by the compiler.
+  public enum ExperimentalFeatureCompilerSetting { }
+
+  // Cases synthesized by the compiler.
+  public enum WarningsAsErrorsCompilerSetting { }
+
+  // We know ahead of time all of the cases, so this is specified explicitly.
+  public enum StrictConcurrency {
+  case minimal
```
slavapestov, line 204: Indentation

proposals/0000-compiler-settings.md
```diff
@@ -0,0 +1,284 @@
...
+]
+```
+
+By specifying compiler options in such a manner, we are able to solve all of the
+problems above:
+
+1. The programmer can avoid module splitting when they wish to use compiler
+   flags only on a specific file.
+
+2. A programmer can specify options on a per file basis instead of only on a per
```
slavapestov, line 139: why "a programmer" here and "the programmer" above?

proposals/0000-swift-settings.md
```diff
@@ -0,0 +1,446 @@
...
+
+Today Swift contains multiple ways of changing compiler and language behavior
+from the command line including specifying the default actor isolation via
+[Controlling Default Actor Isolation](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0466-control-default-actor-isolation.md). Since
+these compiler and language behaviors can only be manipulated via the command
+line and the compiler compiles one module at a time, users are forced to apply
+these flags one module at a time instead of one file at a time. When confronted
+with this, users are forced to split the subset of files into a separate module
+creating an unnecessary module split. We foresee that this will occur more often
+in the future if new features like [Controlling Default Actor Isolation](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0466-control-default-actor-isolation.md)
```
hborla, line 37: I think this specific argument isn't compelling, because we really don't want to introduce more features like default actor isolation in the future. I think you can just remove this sentence and say this applies for default actor isolation and also for the various diagnostic control flags.

proposals/0000-swift-settings.md
```diff
@@ -0,0 +1,446 @@
...
+options is added to `SwiftSetting` by defining a static method on `SwiftSetting`
+in an extension. By using static methods and extensions in this manner, we are
+able to make `SwiftSetting` extensible without requiring the standard library to
+be updated whenever a new options is added since the extension can be defined in
+other modules. As an example, we add support for `defaultIsolation` via an
+extension in the Concurrency module as follows:
+
+```swift
+extension SwiftSetting {
+  public static func defaultIsolation(_ isolation: Actor.Type?) -> SwiftSetting { SwiftSetting() }
```
hborla, line 114: This should match the package manifest API in SE-0466, which uses `MainActor.Type?`

proposals/0000-swift-settings.md
```diff
@@ -0,0 +1,446 @@
...
+If a command line option is set to different values at the module and file
+level, the setting at the file level will take precedence.
+
+## Source compatibility
+
+The addition of the `#SwiftSettings` and `SwiftSetting` cannot impact the
+parsing of other constructs since the name shadowing rules in the compiler will
+ensure that any local macros or types that shadow those names will take
+precedence. If the user wishes to still use `#SwiftSettings`, the user can spell
+the macro as `#Swift.SwiftSettings` as needed.
```
hborla, line 170: Does this work because it's parsed as a freestanding declaration macro? I'm pretty sure that in expression context, `#Swift.SwiftSettings` is parsed as a `#Swift` freestanding expression macro with `.SwiftSettings` as a postfix member. (doesn't really matter for this proposal, only for my curiosity!)

proposals/0000-swift-settings.md
```diff
@@ -0,0 +1,446 @@
...
+
+This proposal does not inherently break ABI. But it can break ABI if a command
+line option is enabled that causes the generated code's ABI to change. As part
+of enabling an option, one must consider such implications.
+
+## Implications on adoption
+
+Adopters of this proposal should be aware that while this feature does not
+inherently break ABI, enabling options using `#SwiftSettings` that break ABI can
+result in ABI breakage.
```
hborla, line 191: You can be more explicit here - say that changing default actor isolation on public API is not source compatible and not ABI compatible.

proposals/0000-swift-settings.md
```diff
@@ -0,0 +1,446 @@
...
+
+3. Commiting the updated file into main.
+
+Since one is only updating one file at a time and using warningsAsErrors, the
+updates can be done incrementally on a per file basis, backsliding cannot occur
+in the file, and most importantly main is never in an intermediate, partially
+updated state. We think that this approach will make it significantly easier for
+larger modules to be migrated to strict concurrency.
+
+#### `strictConcurrency`
```
hborla, line 237: I recommend omitting this section and the next one; since you've left `strictConcurrency` and `warningsAsErrors` as a future direction, it's not a good use of time for reviewers to comment on the details of how it would be accomplished in the review thread for this proposal.

proposals/0000-swift-settings.md
```diff
@@ -0,0 +1,446 @@
...
+
+## Proposed solution
+
+We propose introducing a new macro called `#SwiftSettings` that takes a list of
+structs of type `SwiftSetting`. `SwiftSetting` will possess a static method for
+each command line argument that it supports. Any parameters that would be passed
+on the command line are passed as parameters to the static method. The compiler
+upon parsing a `#SwiftSettings` macro updates its internal state to set the
+appropriate settings before compiling the rest of the file. An example of such a
+`#SwiftSettings` invocation is the following:
```
hborla, line 64: I didn't see a mention in the detailed design about the reason why you chose this spelling, which is to match the Swift Package Manifest API so there's a consistent spelling for settings in source code. I recommend putting that either here or in the detailed design.

proposals/0000-swift-settings.md
```diff
@@ -0,0 +1,446 @@
...
+1. It prevents the ability to perform lookup at point in the IDE and introduces
+   more bespoke code into the compiler to support this feature.
+2. It would prevent the introduction of new SwiftPM APIs in the future that take
+   a `Swift.SwiftSetting`.
+
+### "import feature" syntax
+
+We could also use an "import" syntax to enable features like Python and
+Scala. We decided to not follow this approach since `#` is a manner in swift of
+saying "compile time".
```
hborla, line 406: Further, we'd be introducing a third way of specifying build settings (command line spelling, package manifest spelling, and an import-like spelling). The approach in this proposal was taken primarily to keep a consistent way of spelling build settings in source code.

proposals/0000-swift-settings.md
```diff
@@ -0,0 +1,446 @@
...
+
+### Typealias syntax
+
+We could also use a typealiased based syntax. But this would only be able to be
+used to specify the current global actor and would not let us add additional
+features such as controlling strict concurrency or warnings as errors.
```
hborla, line 412: More importantly, a type alias does not work for default actor isolation, because we'd need some type to represent `nonisolated`. It's also easy to get wrong because misspelling a type alias is still valid Swift code, etc.

proposals/0000-swift-settings.md
```diff
@@ -0,0 +1,446 @@
...
+
+### Adding the ability to push/pop compiler settings
+
+We do not think that it makes sense to implement the ability to push/pop
+compiler settings since we are attempting to specifically support compiler
+settings that semantically can have file wide implications. If we wanted to
+support push/pop of compiler settings it would involve different trade-offs and
+restrictions on what settings would be able to be used since the restrictions
+would necessarily need to be greater. As an example, we could never allow for
```
hborla, line 421: There's an unfinished sentence here.

hborla (commented):
This is all editorial feedback, the proposal content itself looks good! I think you've addressed all of the feedback surfaced in the pitch thread.

rjmccall:
This has been superseded by SE-0478.

PR #2709, opened 2025-02-25
Add a proposal template for Swift Testing

stmontgomery:
This adds a template for proposals focused on the [Swift Testing](https://github.com/swiftlang/swift-testing) project.

### DIFF
--- a/proposal-templates/0000-swift-testing-template.md
+++ b/proposal-templates/0000-swift-testing-template.md
@@ -6,7 +6,7 @@
 * Bug: _if applicable_ [swiftlang/swift-testing#NNNNN](https://github.com/swiftlang/swift-testing/issues/NNNNN)
 * Implementation: [swiftlang/swift-testing#NNNNN](https://github.com/swiftlang/swift-testing/pull/NNNNN)
 * Previous Proposal: _if applicable_ [SWT-XXXX](XXXX-filename.md)
-* Previous Revision: _if applicable_ [1](https://github.com/swiftlang/swift-testing/blob/...commit-ID.../Documentation/Proposals/NNNN-filename.md)
+* Previous Revision: _if applicable_ [1](https://github.com/swiftlang/swift-evolution/blob/...commit-ID.../proposals/NNNN-filename.md)
 * Review: ([pitch](https://forums.swift.org/...))
 
 When filling out this template, you should delete or replace all of the text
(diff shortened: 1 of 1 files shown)


proposal-templates/0000-swift-testing-template.md
```diff
@@ -0,0 +1,184 @@
+# Swift Testing Feature name
+
+* Proposal: [SWT-NNNN](NNNN-filename.md)
+* Authors: [Author 1](https://github.com/author1), [Author 2](https://github.com/author2)
+* Status: **Awaiting implementation** or **Awaiting review**
+* Bug: _if applicable_ [swiftlang/swift-testing#NNNNN](https://github.com/swiftlang/swift-testing/issues/NNNNN)
+* Implementation: [swiftlang/swift-testing#NNNNN](https://github.com/swiftlang/swift-testing/pull/NNNNN)
+* Previous Proposal: _if applicable_ [SWT-XXXX](XXXX-filename.md)
+* Previous Revision: _if applicable_ [1](https://github.com/swiftlang/swift-evolution/blob/...commit-ID.../proposals/NNNN-filename.md)
```
stmontgomery, line 9: - [ ] Decide whether proposals which use this template should be placed in a distinct (sub)directory other than `proposals/`, and update this if so.

hborla:
The template itself looks great, the main thing I think we need to decide on before merging this is your question in https://github.com/swiftlang/swift-evolution/pull/2709#discussion_r1971998852

proposal-templates/0000-swift-testing-template.md
stmontgomery: We could decide to place them in a new subdirectory such as `proposals/testing/`, but the more I think about it, the more I think it might be better to keep them all in a single directory, the existing `proposals/` directory:

- The "**SWT**-NNNN" prefix will allow distinguishing these from "**SE**-NNNN" proposals.
- The [dashboard webpage](https://www.swift.org/swift-evolution/) should be able (with more effort) to distinguish them based on their prefix, too, and allow filtering.
- Having all the proposals in a single directory means that the list of places where the [swift-evolution-metadata-extractor](https://github.com/swiftlang/swift-evolution-metadata-extractor) tool looks for proposal files wouldn't need to be expanded. (It's possible it might need other modifications, I'm not sure.)
- If we _did_ place these in a new subdirectory, it would prompt a question about whether the existing proposals there should be organized into subdirectories of their own. (For example, should the language or SwiftPM proposals be in analogous subdirectories?) But I worry if we did that it would break existing links in various places. So then we'd be left in a situation where Testing proposals are in a special subdirectory, but for historical reasons the older proposals are in the parent directory, and that may seem odd.

So personally I'm coming to the conclusion that we should plan to place these proposals in the existing directory alongside all the others. Thoughts?

grynspan: I think putting them in the same directory is fine precisely because we're using a different prefix.

stmontgomery: We realized the _filenames_ of the proposals do not include the prefix, so we will plan to place new proposals which use the testing template into a distinct `proposals/testing/` subdirectory.

PR #2715, opened 2025-03-02
Fix invalid URLs in Improving the approachability of data-race safety

stzn:
A 404 not found page appeared, so I think we needed to update the URLs.


xwu:
Thank you!

PR #2720, opened 2025-03-03
Global-actor isolated conformances

DougGregor:


proposals/nnnn-isolated-conformances.md
```diff
@@ -0,0 +1,571 @@
...
+```
+
+The `Any` value could contain `MyModelType`, in which case the conformance to `Equatable` will be isolated. In such cases, the `as?` operation will check whether the code is running on the executor associated with the conformance's isolation. If so, the cast can succeed; otherwise, the case will fail (and produce `nil`).
+
+## Detailed design
+
+The proposed solution describes the basic shape of isolated conformances and how they interact with the type system. This section goes into more detail on the data-race safety issues that arise from the introduction of isolated conformances into the language. Then it details three rules that, together, ensure freedom from data race safety issues in the presence of isolated conformances:
+
+1. An isolated conformance can only be used within its isolation domain.
+2. When an isolated conformance is used to satisfy a generic constraint `T: P`, the generic signature must not include either of the following constraints: `T: Sendable` or `T.Type: Sendable`.
```
hborla, line 200: I think this is leftover from an earlier iteration that used `T.Type: Sendable` rather than `T: SendableMetatype`

```suggestion
2. When an isolated conformance is used to satisfy a generic constraint `T: P`, the generic signature must not include either of the following constraints: `T: Sendable` or `T: SendableMetatype`. 
```

DougGregor: Fixed, thank you!

proposals/nnnn-isolated-conformances.md
```diff
@@ -0,0 +1,610 @@
...
+nonisolated func callQGElsewhere_restricted<T: Q>(_: T.Type) where T: ~SendableMetatype {
+  Task.detached {
+    T.g() // error: capture of non-Sendable metatype T.Type in concurrently-executed closure
+  }
+}
+```
+
+With this notion, we could amend rule (2) to prohibit using an isolated protocol conformance for a requirement `T: Q` when the generic signature contains either the requirement `T: Sendable` or `T: SendableMetatype`. This closes the data race safety hole with isolated conformances being carried through metatypes.
+
+Unfortunately, this means that isolated conformances won't work with any existing generic code, because all generic code in existence today assumes that all metatypes are `Sendable`. Most of that code could be updated with requirements of the form `T.Type: ~Sendable` and without other changes, but it would require an ecosystem-wide change in support of a somewhat niche feature.
```
xwu, line 433:
```suggestion
Unfortunately, this means that isolated conformances won't work with any existing generic code, because all generic code in existence today assumes that all metatypes are `Sendable`. Most of that code could be updated with requirements of the form `T: ~SendableMetatype` and without other changes, but it would require an ecosystem-wide change in support of a somewhat niche feature.
```

proposals/nnnn-isolated-conformances.md
```diff
@@ -0,0 +1,610 @@
...
+
+Therefore, this proposal suggests that we change the meaning of existing generic code to *not* be able to assume that a given metatype depending on a generic parameter is `Sendable`. Essentially, it will be as-if `T: ~SendableMetatype` has been applied to every generic parameter `T`. This will have the effect of rejecting the implementation of `callQGElsewhere`. After such a change to the language, the signature of `callQGElsewhere` could be updated by adding the requirement `T: SendableMetatype`. 
+
+Note that, any time a value of type `T` crosses an isolation boundary, it's metatype is accessible via   [`type(of:)`](https://developer.apple.com/documentation/swift/type(of:)), so it also crosses the isolation boundary. This provides us with an inference rule that can help lessen the impact of this source compatibility break: if a generic signature contains a requirement `T: Sendable`, then we can infer the requirement `T: SendableMetatype`. Doing so involves specifying that `Sendable` refines a new marker protocol, `SendableMetatype`:
+
+```swift
+/*@marker*/ protocol SendableMetatype { }
+/*@marker*/ protocol Metatype: SendableMetatype { }
```
xwu, line 441:
```suggestion
/*@marker*/ protocol Sendable: SendableMetatype { }
```

Is this the intended statement?

DougGregor: Yes!

PR #2730, opened 2025-03-09
Mark SE-0436 as implemented in Swift 6.1

j-f1:
It seems that the feature was reverted out of Swift 6.0, but ended up landing in 6.1. I also added a link to the PR that enables the `ObjcImplementation` language feature.

cc: @beccadax


DougGregor (approved):
Thanks!

PR #2733, opened 2025-03-12
[SE-0467] Update the `MutableSpan` proposal

benrimmington:
* Add the implementation and review links.
* Fix a few missing or incorrect words.
* Fix two links in the "Future directions".

Cc: @glessard, @jckarter


benrimmington:
Ping @jckarter

PR #2772, opened 2025-04-04
Add a proposal for a string interpolation default parameter

natecook1000:
A new string interpolation syntax for providing a default string when interpolating an optional value.

```swift
let age: Int? = nil
print("Your age: \(age, default: "missing")")
// Prints "Your age: missing"
```


xwu:
@natecook1000 Happy to kick off review whenever—if you'd like a chance to tack on a paragraph about `Either` as an alternative/future direction, I can wait for that. Let me know!

PR #2801, opened 2025-04-22
A few updates to SE-0471 isIsolatingCurrentContext

ktoso:
SE review follow up, moving to a simpler model of detecting if we were able to determine the isolation without crashing.


proposals/0471-SerialExecutor-isIsolated.md
```diff
@@ -125,15 +123,63 @@ This proposal specifically adds the "if `isIsolatingCurrentContext` is available
...
+}
+```
+
+This allows developers to write "warn if wrong isolation" code, before moving on to enable preconditions in a future release of a library. This gives library developers, and their adopters, time to adjust their code usage before enabling more strict validation mode in the future, for example like this:
 
-The presence of a non-default implementation of the `isIsolatingCurrentContext` protocol witness is detected by the compiler and the runtime can detect this information in order to determine if the new function should be used for these checks. In other words, if there is an implementation of the requirement available _other than_ the default one provided in the concurrency library, the runtime will attempt to use this method _over_ the `checkIsolated` API. This allows for a smooth migration to the new API, and enables the use of this method in if the runtime would like issue a check that cannot cause a crash.
+```swift
+func something(operation: @escaping @isolated(any) () -> ()) {
+  operation.isolation.withSerialExecutor { se in 
+    if !se.isIsolatingCurrentContext() {
```
DougGregor, line 172:
```suggestion
    if let knownIsolating = se.isIsolatingCurrentContext(), !knownIsolating {
```

ktoso: Missed to apply this change? @DougGregor

PR #2809, opened 2025-04-24
[SE-0479] Amend Motivation text and add Future Directions.

amritpan:
Updated text to reflect limitations.


amritpan:
@beccadax I'd like to add this to our proposal under review.

beccadax:
I'll run this past the LSG. In the mean time, do you think you could reply to this effect in the review thread? It should be mentioned there either way.

amritpan:
Yes, I thought I should merge this and then reply, but I will update there first.

beccadax (approved):
Since this is just a clarification, not a design change, I'll merge it and announce it in the discussion thread.

PR #2821, opened 2025-05-02
[WIP][DNM] identical functions

vanvoorden:


vanvoorden:
@lorentey Thanks for all the help so far! I'm going to continue working on this. My plan for right now is to actually split this up into two proposals:

1. Add `isIdentical` functions where it makes sense in Standard Library and Foundation. This would include `Array`, `Dictionary`, `Set`, and other types where it makes sense.
2. Add `isKnownIdentical` to `Equatable` with a default implementation returning `nil`.

I think it makes more sense to break these apart and review these independently. Mixed feedback on one does not have to block shipping the other. How does that idea sound?

I can plan to spend time writing out these two proposals and following up with questions in the review cycle. I can also plan to spend time writing the implementation code in standard library and foundation… but I might possibly have another friend I can recruit to also help for that.

My goal is to land before the 6.3 branch cut. I'm guessing that will be end of Q3.

You added a lot of great ideas and I would like to list you as a coauthor on both of these if you are interested. Let me know what you think. No super urgent rush… I'm blocked on something else for right now and I'm expecting to put some more time on this in about two or three weeks.

vanvoorden:
Closing for now and moving over to https://github.com/swiftlang/swift-evolution/pull/2875.

PR #2829, opened 2025-05-02
SE-0443: Update proposal to reflect chosen diagnostic group name style

tshortli:
The text of SE-0443 uses snake case diagnostic group names, but in practice we've settled on using Pascal case for diagnostic group names. Additionally, the name `DeprecatedDeclaration` was chosen for the deprecation diagnostic group. I think it would be best for the proposal to reflect both of these stylistic choices so that it can serve as accurate documentation for the feature in practice.


DmT021 (approved):

rjmccall:
Seems reasonable to me.

PR #2836, opened 2025-05-07
Update inline-array with other languages with similar feature

airspeedswift:


hborla:
The other major concern from the pitch and review discussion that isn't addressed in the proposal is the concern that the inline storage is not emphasized enough in the syntax sugar. This was the primary consideration for choosing the name `InlineArray` over other alternatives, and I think this proposal needs to address that concern somewhere.

PR #2843, opened 2025-05-10
Add SE proposal for Codable error printing

ZevEisenberg:
SE proposal to accompany https://github.com/swiftlang/swift/pull/80941

### DIFF
--- a/proposals/NNNN-codable-error-printing.md
+++ b/proposals/NNNN-codable-error-printing.md
@@ -1,0 +1,160 @@
+# Improve `EncodingError` and `DecodingError`'s printed descriptions
+
+* Proposal: [SE-NNNN](NNNN-codable-error-printing.md)
+* Authors: [Zev Eisenberg](https://github.com/ZevEisenberg)
+* Review Manager: TBD
+* Status: **Awaiting review**
+* Implementation: https://github.com/swiftlang/swift/pull/80941
+* Review: TBD
+
+## Introduction
+
+`EncodingError` and `DecodingError` do not specify any custom debug description. The default descriptions bury the useful information in a format that is difficult to read. Less experienced developers may assume they are not human-readable at all, even though they contain useful information. The proposal is to conform `EncodingError` and `DecodingError` to `CustomDebugStringConvertible` and provide nicely formatted debug output.
+
+## Motivation
+
+Consider the following example model structs:
+
+```swift
+struct Person: Codable {
+  var name: String
+  var home: Home
+}
+
+struct Home: Codable {
+  var city: String
+  var country: Country
+}
+
+struct Country: Codable {
+  var name: String
+  var population: Int
+}
+```
+
+Now let us attempt to decode some invalid JSON. In this case, it is missing a field in a deeply nested struct.
+
+```swift
+// Note missing "population" field
+let jsonData = Data("""
+[
+  {
+    "name": "Ada Lovelace",
+    "home": {
+      "city": "London",
+      "country": {
+        "name": "England"
+      }
+    }
+  }
+]
+""".utf8)
+
+do {
+  _ = try JSONDecoder().decode([Person].self, from: jsonData)
+} catch {
+  print(error)
+}
+```
+
+This outputs the following:
+
+`keyNotFound(CodingKeys(stringValue: "population", intValue: nil), Swift.DecodingError.Context(codingPath: [_CodingKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "home", intValue: nil), CodingKeys(stringValue: "country", intValue: nil)], debugDescription: "No value associated with key CodingKeys(stringValue: \"population\", intValue: nil) (\"population\").", underlyingError: nil))`
+
+All the information you need is there:
+- The kind of error: a missing key
+- Which key was missing: `"population"`
+- The path of the value that had a missing key: index 0, then key `"home"`, then key `"country"`
+- The underlying error: none, in this case
+
+However, it is not easy or pleasant to read such an error, particularly when dealing with large structures or long type names. It is common for newer developers to assume the above output is some kind of log spam and not even realize it contains exactly the information they are looking for.
+
+## Proposed solution
+
+Conform `EncodingError` and `DecodingError` to `CustomDebugStringConvertible` and provide a clean, readable debug description for each. Here is an example of the proposed change for the same decoding error as above.
+
+```
+Key 'population' not found in keyed decoding container.
+Debug description: No value associated with key CodingKeys(stringValue: "population", intValue: nil) ("population").
+Path: [0]/home/country
+```
+
+Complete examples of the before/after diffs are available in the description of the [implementation pull request](https://github.com/swiftlang/swift/pull/80941) that accompanies this proposal.
+
+**Note 1:** this proposal is _not_ intended to specify an exact output format. The above is provided as an example, and is not a guarantee of current or future behavior. You are still free to inspect the contents of thrown errors directly if you need to detect specific problems.
+
+**Note 2:** the output could be further improved by modifying `JSONDecoder` to write a better debug description. See [Future Directions](#future-directions) for more.
+
+## Detailed design
+
+```swift
+@available(SwiftStdlib 6.2, *)
+extension EncodingError: CustomDebugStringConvertible {
+  public var debugDescription: String {...}
+}
+
+@available(SwiftStdlib 6.2, *)
+extension DecodingError: CustomDebugStringConvertible {
+  public var debugDescription: String {...}
+}
+```
+
+## Source compatibility
+
+The new conformance changes the result of converting an `EncodingError` or `DecodingError` value to a string. This changes observable behavior: code that attempts to parse the result of `String(describing:)` or `String(reflecting:)` can be misled by the change of format.
+
+However, the documentation of these interfaces explicitly state that when the input type conforms to none of the standard string conversion protocols, then the result of these operations is unspecified.
+
+Changing the value of an unspecified result is not considered to be a source incompatible change.
+
+## ABI compatibility
+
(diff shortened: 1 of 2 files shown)


stephentyrone:
Hi @ZevEisenberg, I'm really excited about this. Can you begin a [pitch thread](https://forums.swift.org/c/evolution/pitches/5) on the forums, and then we can work on getting it ready for formal evolution review in a few weeks?

proposals/NNNN-codable-error-printing.md
```diff
@@ -0,0 +1,175 @@
...
+
+## ABI compatibility
+
+The proposal retroactively conforms two previously existing standard types to a previously existing standard protocol. This is technically an ABI breaking change: on ABI-stable platforms, we may have preexisting Swift binaries that assume that `EncodingError is CustomDebugStringConvertible` or `DecodingError is CustomDebugStrinConvertible` returns `false`, or ones that are implementing this conformance on their own.
+
+We do not expect this to be an issue in practice.
+
+## Implications on adoption
+
+[Unsure what to add here. I see stuff in [SE-0445](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0445-string-index-printing.md#implications-on-adoption), but I'm not sure how much of that applies here. I don't know if I need to be doing the `@backDeployed` things that proposal mentions, and when I look at the code from the PR, I see `@_alwaysEmitIntoClient // FIXME: Use @backDeployed`.]
```
kperryua, line 130: I suspect that in SE-0445 the back-deploying was important because the non-API public symbol `_description` existed on `String.Index` in prior Swift versions, and was reimplemented using `debugDescription`? I'm not certain that this is quite as relevant here, which suggests that there's nothing really useful to add to this section.

@lorentey can you provide any background on SE-0445 and/or advice here?

xwu:
I'll do an editorial pass later this week (or weekend) and have suggestions to help you fill out all the parts of the proposal text. Once it's taken out of draft status, I'll get back to you after the language steering group is satisfied it's ready to run as a proposal and schedule the review period, which will run over two weeks.

Implementation work proceeds separately; after a feature lands in main (which would be no earlier than July given the timeline above), then it can be evaluated if cherrypicking to a release branch is appropriate, but expect it's well out of the window for 6.2.

proposals/NNNN-codable-error-printing.md
```diff
@@ -0,0 +1,173 @@
...
+
+```
+Key 'population' not found in keyed decoding container.
+Debug description: No value associated with key CodingKeys(stringValue: "population", intValue: nil) ("population").
+Path: [0]/home/country
+```
+
+(Note: the output could be further improved by modifying `JSONDecoder` to write a better debug description. See [Future Directions](#future-directions) for more.)
+
+### Structure
```
xwu: In general, we don't guarantee the format of a description or debug description; giving a specific rundown of what information will be presented, under what conditions, and in which order reads like a guarantee that doesn't seem necessary to achieve your motivation (and we don't want users to be parsing this programmatically).

I would suggest either removing this information or phrasing this more as an example of how the implementation might improve usability—as well as explicitly noting, as we did in SE-0445, that it is _not_ normative and _is_ subject to future change without review or advance notice.

proposals/NNNN-codable-error-printing.md
```diff
@@ -0,0 +1,173 @@
...
+
+## ABI compatibility
+
+The proposal retroactively conforms two previously existing standard types to a previously existing standard protocol. This is technically an ABI breaking change: on ABI-stable platforms, we may have preexisting Swift binaries that assume that `EncodingError is CustomDebugStringConvertible` or `DecodingError is CustomDebugStrinConvertible` returns `false`, or ones that are implementing this conformance on their own.
+
+We do not expect this to be an issue in practice.
+
+## Implications on adoption
+
+[Unsure what to add here. I see stuff in [SE-0445](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0445-string-index-printing.md#implications-on-adoption), but I'm not sure how much of that applies here. I don't know if I need to be doing the `@backDeployed` things that proposal mentions, and when I look at the code from the PR, I see `@_alwaysEmitIntoClient // FIXME: Use @backDeployed`.]
```
xwu: Here's what I might say: The conformance being proposed here is not backdeployable. As a result, code that runs on ABI-stable platforms with earlier versions of the standard library won't output these nicer debug descriptions.

It is up to you if you want to make `debugDescription` itself (the property, not the conformance) backdeployable—this would make it possible for newly compiled code to invoke `debugDescription` explicitly if they absolutely need the nicer description, but that is not something we would either recommend or expect users to reach for pervasively to work around the adoption issue outlined above.

proposals/NNNN-codable-error-printing.md
```diff
@@ -0,0 +1,173 @@
...
+  "feathers" : "some",
+  "name" : null
+}
+```
+
+## Alternatives considered
+
+The original version of this proposal suggested conforming `EncodingError` and `DecodingError` to `CustomStringConvertible`, not `CustomDebugStringConvertible`. The change to the debug-flavored protocol emphasizes that the new descriptions aren't intended to be used outside debugging contexts. This is in keeping with the precedent set by [SE-0445](0445-string-index-printing.md).
+
+The original version also proposed changing `CodingKey.description` to return the bare string or int value, but changing the exsting implementation of an existing public method was deemed too potentially dangerous.
```
xwu: Can you specify what you mean here by "dangerous"?

ZevEisenberg: It was in reference to https://github.com/swiftlang/swift/pull/80941#discussion_r2064277369. I'll update the proposal, plus I'll link to that thread if appropriate.

proposals/NNNN-codable-error-printing.md
ZevEisenberg: So, if we back-deployed `debugDescription`, and if someone knew about this change, they would be able to ship an update that would work even when running on OSes that don't have the change? But only if they explicitly called `debugDescription` instead of just logging the errors with default string interpolation or whatever method they are using now?

Since the information in question is primarily for use as a developer convenience while debugging, I'm inclined to go the simpler route. If you're digging around in error logs, and you think of looking into this and learn about the `debugDescription` hack, I figure you're probably savvy enough to just read the old-style error message yourself?

proposals/NNNN-codable-error-printing.md
xwu: So, if I understand the point correctly, "dangerous" is referring concretely to `Foundation.JSONDecoder` having a known dependency on non-guaranteed implementation details of `CodingKey.description`?

proposals/NNNN-codable-error-printing.md
xwu: Yup, that part is a design decision and not a proposal drafting decision, so I leave it to you :)

proposals/NNNN-codable-error-printing.md
ZevEisenberg: yep! I'm thinking "dangerous" is not the right word, and @kperryua's original message or your summary is closer to what we should put here in the proposal.

xwu:
@ZevEisenberg When you're satisfied with your edits in relation to the above, feel free to ping me and take this out of draft status.

ZevEisenberg:
Will do! Sorry, I've been sick. I'll get back to it ASAP.

proposals/NNNN-codable-error-printing.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Implications on adoption
+
+### Conformance to `CustomDebugStringConvertible`
+
+The conformance to `CustomDebugStringConvertible` is not backdeployable. As a result, code that runs on ABI-stable platforms with earlier versions of the standard library won't output the new debug descriptions.
+
+### `debugDescription` Property
+
+The change to the debug output is purely cosmetic: it improves readability, but does not change the information contained in the message. For this reason, we think it is not worthwhile to backport the `debugDescription` property itself. Since the conformance to `CustomDebugStringConvertible` is not back-deployable, a backport of property itself would be of limited utility anyway.
```
xwu: Whether something _can_ be backdeployed is an implication on adoption that is appropriately part of the proposal, but it's outside of the scope of Swift Evolution whether something _actually will_ be backdeployed on any particular platform since it's ultimately up to the platform owner. So you could make this even shorter and say something like, the implementation of `debugDescription` itself is possible to backdeploy but would be on its own of limited utility.

proposals/NNNN-codable-error-printing.md
```diff
@@ -0,0 +1,160 @@
...
+
+However, the documentation of these interfaces explicitly state that when the input type conforms to none of the standard string conversion protocols, then the result of these operations is unspecified.
+
+Changing the value of an unspecified result is not considered to be a source incompatible change.
+
+## ABI compatibility
+
+The proposal conforms two previously existing stdlib types to a previously existing stdlib protocol. This is technically an ABI breaking change: on ABI-stable platforms, we may have preexisting Swift binaries that implement a retroactive `CustomDebugStringConvertible` conformance, or binaries that assume that the existing error types do _not_ conform to the protocol.
+
+We do not expect this to be an issue in practice. The above checks represent edge cases, since checking an arbitrary error for conformance to `CustomDebugStringConvertible` at run-time seems unlikely. And if someone has added a conformance, when they re-build with a stdlib that includes this conformance, they will get a build-time warning.
```
xwu:
```
And if someone has added a conformance, when they re-build with a stdlib that includes this conformance, they will get a build-time warning.
```

What happens when you recompile doesn't make ABI breakage on ABI-stable platforms any more or less problematic, so I don't think it needs comment here... The preceding statement that you think a certain usage is unlikely is pretty much a restatement of the first sentence in the paragraph that you don't think it's an issue in practice, so I think that one sentence pretty much captures the whole paragraph.

ZevEisenberg: I felt weird about just stating that we don't expect it to be an issue but [not explaining why](https://xkcd.com/979/), but I've tightened up the wording. I figure, this section will be most relevant if someone in the future _is_ negatively affected by the change, and in that case, I don't want them finding a years-old document telling them that their problem is unlikely 😆

xwu:
Thanks—we don't think there's additional feedback on this on our part. If you feel it's ready, you can take the PR out of draft status.

PR #2846, opened 2025-05-14
[SE-0466] Remove "prospective" when referring to vision doc

ole:
The proposal refers to the vision for approachable data-race safety as "prospective" in one place, but the vision has been accepted. So remove the "prospective".

This also fixes 2 typos. I force-pushed because I found the typos after submitting the PR.

cc @hborla


hborla (approved):
Thank you again!

PR #2850, opened 2025-05-16
SE-0472: Add missing `()` to function type in `Task.immediate{Detached}` declarations

xedin:


xedin:
cc @allevato as review manager

allevato (approved):
Thanks!

PR #2851, opened 2025-05-19
Binary library artifacts amends

daniel-grumberg:
Updates based on evolution review

### DIFF
--- a/proposals/0482-swiftpm-static-library-binary-target-non-apple-platforms.md
+++ b/proposals/0482-swiftpm-static-library-binary-target-non-apple-platforms.md
@@ -10,11 +10,10 @@
 
 ## Introduction
 
-Swift continues to grow as a cross-platform language supporting a wide variety of use cases from [programming embedded device](https://www.swift.org/blog/embedded-swift-examples/) to [server-side development](https://www.swift.org/documentation/server/) across a multitude of [operating systems](https://www.swift.org/documentation/articles/static-linux-getting-started.html). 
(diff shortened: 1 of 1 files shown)


proposals/0482-swiftpm-static-library-binary-target-non-apple-platforms.md
```diff
@@ -10,11 +10,10 @@
...
+Swift continues to grow as a cross-platform language supporting a wide variety of use cases from [programming embedded device](https://www.swift.org/blog/embedded-swift-examples/) to [server-side development](https://www.swift.org/documentation/server/) across a multitude of [operating systems](https://www.swift.org/documentation/articles/static-linux-getting-started.html).
 However, currently SwiftPM supports linking against binary dependencies on Apple platforms only.
-This proposal aims to make it possible to provide static library dependencies exposing a C interface on non-Apple platforms.
-
-Swift-evolution thread: 
+This proposal aims to make it possible to provide static library dependencies exposing a C interface on non-Apple platforms that depend only the standard C library.
+Unfortunately, this proposal doesn't provide support for distributing Swift libraries built in this way due to the Swift ABI not being stable on all platforms.
```
FranzBusch, line 16: I think this should just go into Future directions and not be called out in the intro.

kubamracek: See <https://forums.swift.org/t/se-0482-binary-static-library-dependencies/79634/22> -- I understand this is a weird callout, but for readers of this proposals, let's please clarify the scope of it. I'd suggest something like "The scope of this proposal is C libraries only, distributing Swift libraries has additional challenges (see Future directions)."

al45tair (approved):

PR #2875, opened 2025-06-04
[WIP][DNM] add is identical methods

vanvoorden:


vanvoorden:
@lorentey Here is my new attempt at a pitch only for identical functions on concrete types. My plan is to propose the update on `Equatable` in a separate pitch. Want to take a look and let me know what you think? How do you feel about being named as a coauthor? I expect to have enough time to complete all the implementations and unit tests myself and I do not need you to commit to allocating any engineering time.

I was planning to keep working on this over the next week and have it live on the forums 2025-06-16 after WWDC.

lorentey:
This looks good! Thank you for pushing these. 👍🏼

We may need to subset out the Foundation APIs and run them through the evolution process specific to that project. Cc @parkera

vanvoorden:
> We may need to subset out the Foundation APIs and run them through the evolution process specific to that project.

Ahh… that makes sense. Keeping them together in one place would be nice… but I'm not opposed to breaking them apart.

vanvoorden:
https://github.com/swiftlang/swift-evolution/blob/main/proposals/0376-function-back-deployment.md

@lorentey What would your thoughts be about shipping these functions as `@backDeployed`? Would you know if a decision like that must be documented in the proposal review itself… or can the proposal review remain agnostic over backdeployment and we save that conversation for the implementation diff reviews?

PR #2893, opened 2025-06-30
[SE-0470] Infer `nonisolated` on conformances

DougGregor:
Introduce two inference rules that infer `nonisolated` when `InferIsolatedConformances` is enabled. These inference rules ensure that most conformances that should be nonisolated remain nonisolated, particular when we're in the main-actor-by-default mode. The end result is that much less code needs to change when enabling isolated conformance inference (including via main-actor-by-default mode). The two rules are:

* If the protocol inherits from `SendableMetatype` (including indirectly, e.g., from `Sendable`), then the isolated conformance could never be used, so it is inferred to be `nonisolated`.
* If all of the declarations used to satisfy protocol requirements are `nonisolated`, the conformance will be assumed to be `nonisolated`.


proposals/0470-isolated-conformances.md
```diff
@@ -665,6 +665,24 @@ This is a generalization of the proposed rules that makes more explicit when con
...
+nonisolated protocol Q {
+  static func create() -> Self
+}
+
+@MainActor struct MyType: /*infers nonisolated*/ Q {
+  nonisolated static func create() -> MyType { ... }
+}
+```
+
+This inference is important for providing source compatibility with and without `InferIsolatedConformances`, and is especially useful useful when combined with default main-actor isolation ([SE-0466](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0466-control-default-actor-isolation.md)), where many more types will become main-actor isolated. Experience with using these features together also identified a some macros (such as [`@Observable`](https://developer.apple.com/documentation/observation/observable())) that produced `nonisolated` members for a protocol conformances, but had not yet been updated to mark the conformance as `nonisolated`. Macro-generated code is much harder for users to update when a source-compatibility issue arises, which makes `nonisolated` conformance inference particularly important for source compatibility.
```
xwu:
```suggestion
This inference is important for providing source compatibility with and without `InferIsolatedConformances`, and is especially useful useful when combined with default main-actor isolation ([SE-0466](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0466-control-default-actor-isolation.md)), where many more types will become main-actor isolated. Experience with using these features together also identified some macros (such as [`@Observable`](https://developer.apple.com/documentation/observation/observable())) that produced `nonisolated` members for a protocol conformances, but had not yet been updated to mark the conformance as `nonisolated`. Macro-generated code is much harder for users to update when a source-compatibility issue arises, which makes `nonisolated` conformance inference particularly important for source compatibility.
```

proposals/0470-isolated-conformances.md
```diff
@@ -662,6 +665,24 @@ This is a generalization of the proposed rules that makes more explicit when con
...
+
+```swift
+nonisolated protocol Q {
+  static func create() -> Self
+}
+
+@MainActor struct MyType: /*infers nonisolated*/ Q {
+  nonisolated static func create() -> MyType { ... }
+}
+```
```
xwu, line 680: So to clarify, based on the rules proposed in this amendment (rather than the alternative), we're looking at the following source evolution behavior:

```swift
// Library A version 1.0 (nonisolated by default)
protocol Q {
  static func create() -> Self
}

// MyApp (main actor by default)
class MyType: /* infers nonisolated */ Q {
  nonisolated static func create() -> MyType { ... }
  func destroy() { ... }
}
```

```swift
// Library A version 2.0 (still nonisolated by default)
protocol Q {
  static func create() -> Self
  func destroy() /* default implementation provided */
}

// MyApp (main actor by default, unchanged source code)
class MyType: /* this now infers @MainActor? */ Q {
  nonisolated static func create() -> MyType { ... }
  func destroy() { ... }
}
```

DougGregor: Yes, that's right. I added some words to that effect

PR #2894, opened 2025-06-30
[SE-0466] Disable `@MainActor` inference when type conforms to a SendableMetatype protocol

DougGregor:
Libraries need a way to communicate that certain protocols can't reasonably be used with global-actor-isolated types. When the primary definition of a type conforms to such a protocol, it will disable `@MainActor` inference for that type. This has the effect of keeping more types `nonisolated` when there's signal that they should not be on the main actor. There are some protocols like this in the standard library (e.g., CodingKey, which was hardcoded to be nonisolated in the synthesis code), where the following code is ill-formed in with main-actor isolation by default:

    struct S: Codable {
      var a: Int

      // error: CodingKeys inferred to `@MainActor`, which makes the conformance
      // to CodingKey main-actor-isolated and then fails (because the requirements
      // need to be nonisolated).
      enum CodingKeys: CodingKey {
        case a
      }
    }

With this amendment, the conformance to `CodingKey` (which inherits from `Sendable`), prevents `CodingKeys` from being inferred to be`@MainActor`.

### DIFF
--- a/proposals/0466-control-default-actor-isolation.md
+++ b/proposals/0466-control-default-actor-isolation.md
@@ -114,8 +114,6 @@ nonisolated protocol Q: Sendable { }
 // nonisolated
 struct S2: Q { }
 
-nonisolated protocol Q: Sendable { }
-
 // @MainActor
 struct S3 { }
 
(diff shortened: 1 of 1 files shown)


proposals/0466-control-default-actor-isolation.md
```diff
@@ -65,6 +65,7 @@ When the default actor isolation is specified as `MainActor`, declarations are i
 * Declarations with inferred actor isolation from a superclass, overridden method, protocol conformance, or member propagation
 * All declarations inside an `actor` type, including static variables, methods, initializers, and deinitializers
 * Declarations that cannot have global actor isolation, including typealiases, import statements, enum cases, and individual accessors
+* Declarations whose primary definition directly conforms to a protocol that inherits `SendableMetatype`
```
xwu, line 68: Q (can be discussed offline, but I've got to jot it down or else I'll forget)—could this be regarded as a special case of bullet point 2 above ("inferred isolation from a protocol"), and could we extend SE-0449 such that `nonisolated protocol P: Q` both cuts off `P` from inheriting the isolation of `Q` and cuts off conforming types from inheriting isolation from other protocol conformances?

DougGregor: I'm hesitant about overloading the meaning of `nonisolated protocol P` further here, because right now it means that the protocol itself should not inherit isolation from its inherited protocols (and that's it). If we also use it to mean "conforming types should be `nonisolated`", then we don't have a way to write the equivalent of `protocol P` (in nonisolated-by-default mode) within main-actor-by-default mode.

xwu: Makes sense. I bring it up because, while there are various reasons why `nonisolated` isn't an isolation, I think there are many users whose mental model is `@MainActor`-versus-`nonisolated`. And the rules here are such that `@MainActor protocol P` confers an isolation to conforming types while `nonisolated protocol P` does not correspondingly remove it. But quite rightly we've already given a different significance to the latter which would not be expressible independently.

stephentyrone (approved):

PR #2908, opened 2025-07-09
Fixups for SE-0466

xwu:
* Edit a typo in the header field
* Add a succinct revision history


MahdiBM:
@xwu I'm also seeing this error but I don't think this PR right now addresses it?

```bash
$ curl https://download.swift.org/swift-evolution/v1/evolution.json | jq '.proposals[] | select(.id == "SE-0466") | .errors'
[
  {
    "code": 0,
    "kind": "error",
    "message": "Missing or invalid proposal status.",
    "suggestion": ""
  }
]
```

xwu:
If people are using the dashboard to track when proposals are up for comment, maybe best to just say it's in review for the next week or so? It will flip back to implemented after anyway.

DougGregor:
> If people are using the dashboard to track when proposals are up for comment, maybe best to just say it's in review for the next week or so? It will flip back to implemented after anyway.

Sure, that makes sense!

PR #2923, opened 2025-07-25
Clarify use of Swift SDKs in `platform-support.md`

MaxDesiatov:
Clarify that platforms would distribute Swift SDKs, which is a different and separate concept from plain "SDK" used in Clang or Xcode.

### DIFF
--- a/visions/platform-support.md
+++ b/visions/platform-support.md
@@ -121,7 +121,7 @@ Platforms that are in Tier 1 should:
 
 - [ ] Have testing in CI, including PR testing.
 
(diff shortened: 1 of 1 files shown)


al45tair:
To be clear, there is _no_ requirement that a platform ship _Swift_ SDKs (which means the artifact bundles that are consumed by the Swift Package Manager), so this change is wrong.

SDK here _includes_ Swift SDKs, but it also includes other means of distributing an SDK for a platform (for instance, Windows ships `Windows.sdk`, which is an Xcode-style SDK containing an `SDKSettings.plist`).

PR #2924, opened 2025-07-25
Fix dead link in SE-0451

calda:
This PR fixes a dead link in SE-0451. The referred section was renamed from `Permissible Characters` to `Permitted Characters`.


rlmiller2024 (commented):
So confused

PR #2937, opened 2025-08-07
Mark some stdlib proposals as implemented

glessard:
Marks SE-0459, SE-0477, SE-0485 and SE-0488 as implemented.


benrimmington:
swiftlang/swift#81300 isn't merged yet for SE-0473.

PR #2949, opened 2025-09-03
Section Placement Control proposal

kubamracek:
Sending the "Section Placement Control" proposal (pitched 3 times on the forums, see links in the doc) for review. Note that the wording and smaller details of the proposal were changed specifically so that proposal **should be reviewable independently** from the [Swift Compile-Time Values](https://github.com/artemcm/swift-evolution/blob/const-values/proposals/0nnn-const-values.md) proposal: The text no longer refers to the `@const` syntax being proposed in Swift Compile-Time Values, and drops the extension that introduced `@constInitialized`.

Even though there is still a clear connection between Section Placement and Compile-Time Values, I would like to request an independent review of this proposal in separation, without having a final design for the exact syntax and allowed data types to be eligible for being compile-time guaranteed. Those aspects are still important and need to be finalized, but do not impact the content of this proposal.

### DIFF
--- a/proposals/0nnn-section-control.md
+++ b/proposals/0nnn-section-control.md
@@ -1,0 +1,447 @@
+# Section Placement Control
+
+* Proposal: [SE-0NNN](0nnn-section-control.md)
+* Authors: [Kuba Mracek](https://github.com/kubamracek)
+* Status: Pitch #3
+* Discussion threads:
+  * Pitch #1: https://forums.swift.org/t/pitch-low-level-linkage-control-attributes-used-and-section/65877
+  * Pitch #2: https://forums.swift.org/t/pitch-2-low-level-linkage-control/69752
+  * Pitch #3: TBD
+
+## Introduction
+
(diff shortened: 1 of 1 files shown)


proposals/0nnn-section-control.md
```diff
@@ -0,0 +1,438 @@
...
+
+The intention is that these attributes are to be used rarely and only by specific use cases; high-level application code should not need to use them directly and instead should rely on libraries, macros and other abstractions over the low-level attributes.
+
+The scope of this proposal is limited to compile-time behavior and compile-time control. We expect that full user-facing solutions for features like linker sets, test discovery or plugins will also require runtime implementations to discover and iterate the contents of custom sections, possibly from multiple modules. This proposal makes sure to provide the right building blocks and artifacts in binaries for the runtime components, but doesn’t prescribe the shape of those. However, it is providing a significant step towards generalized and safe high-level mechanisms for those use cases. See the discussion of that in the sections [Runtime discovery of data in custom sections](#runtime-discovery-of-data-in-custom-sections) and [Linker sets, plugins as high-level APIs](#linker-sets-plugins-as-high-level-apis) in Future Directions.
```
beccadax: This section isn’t really doing what the template says it should:

> A short description of what the feature is. Try to keep it to a single-paragraph "elevator pitch" so the reader understands what problem this proposal is addressing.

Could it be condensed into something like:
```suggestion
This proposal adds `@section` and `@used` attributes which can be applied to global variables. These allow users to directly control which section of the resulting binary should globals variables be emitted into, and give users the ability to disable DCE (dead code elimination) on those. The goal is to enable systems and embedded programming use cases like runtime discovery of test metadata from multiple modules, and also to serve as a low-level building block for higher-level features (e.g. linker sets, plugins).
```

And then the other details discussed later in the proposal?

PR #2981, opened 2025-10-02
Note that 0283 is no longer accepted

awwright:


awwright:
This proposal seems to have fallen into an exceptional state where it's "accepted" but not being implemented. This has caused much confusion, e.g.

https://forums.swift.org/t/accepted-se-0283-tuples-conform-to-equatable-comparable-and-hashable/36658/10
https://forums.swift.org/t/status-of-se-0283-tuples-conform-to-equatable-comparable-and-hashable/46942
https://forums.swift.org/t/why-isnt-my-int-int-hashable/82400

I'm not able to find any process that describes what to do in this situation, but it probably shouldn't be sitting around as "Accepted"

xwu:
The status we assign to such proposals is "expired."

rjmccall:
The LSG decided to make it Returned for Revisions.

PR #3065, opened 2026-01-05
Safe API to load values from `RawSpan`

glessard:
This proposal introduces of a set of safe API to load values of numeric and other safe-to-load types from the memory represented by `RawSpan` instances, with safe conversions from `RawSpan` to `Span` for those same types. In support, it also introduces a set of safe API for safe-to-store types, with conversions from `Span` to `RawSpan` for those safe-to-store types.

### DIFF
--- a/proposals/nnnn-rawspan-safe-loading-api.md
+++ b/proposals/nnnn-rawspan-safe-loading-api.md
@@ -6,7 +6,7 @@
 * Status: **Awaiting implementation**
 * Previous Proposal: follows [SE-0447][SE-0447]
 * Previous Revision: [pitch 1](https://github.com/glessard/swift-evolution/blob/fdd9b855befea7071c43b774330a02b9cc173174/proposals/nnnn-rawspan-safe-loading-api.md)
-* Review: ([pitch 1](https://forums.swift.org/t/83966)), ([pitch 2](https://forums.swift.org/...))
+* Review: ([pitch 1](https://forums.swift.org/t/83966)), ([pitch 2](https://forums.swift.org/t/84144))
 
 [SE-0447]: https://github.com/swiftlang/swift-evolution/blob/main/proposals/0447-span-access-shared-contiguous-storage.md
 [swift-binary-parsing]: https://github.com/apple/swift-binary-parsing
--- a/proposals/nnnn-rawspan-safe-loading-api.md
+++ b/proposals/nnnn-rawspan-safe-loading-api.md
@@ -113,6 +113,7 @@ extension RawSpan {
   /// - Returns: A new value of type `UInt8`, read from `offset`.
   func load(fromByteOffset: Int = 0, as: UInt8.Type) -> UInt8
 
+  /// Returns a value constructed from the raw memory at the specified offset.
   func load(fromByteOffset: Int = 0, as: Int8.Type) -> Int8
 
   /// Returns a value constructed from the raw memory at the specified offset.
@@ -129,24 +130,49 @@ extension RawSpan {
   ///     or the native runtime endianness if `nil`.
   /// - Returns: A new value of type `UInt16`, read from `offset`.
   func load(fromByteOffset: Int = 0, as: UInt16.Type, endianness: Endianness? = nil) -> UInt16
+
+  /// Returns a value constructed from the raw memory at the specified offset.
   func load(fromByteOffset: Int = 0, as: Int16.Type, endianness: Endianness? = nil) -> Int16
+
+  /// Returns a value constructed from the raw memory at the specified offset.
   func load(fromByteOffset: Int = 0, as: UInt32.Type, endianness: Endianness? = nil) -> UInt32
+
+  /// Returns a value constructed from the raw memory at the specified offset.
   func load(fromByteOffset: Int = 0, as: Int32.Type, endianness: Endianness? = nil) -> Int32
+
+  /// Returns a value constructed from the raw memory at the specified offset.
   func load(fromByteOffset: Int = 0, as: UInt64.Type, endianness: Endianness? = nil) -> UInt64
+
+  /// Returns a value constructed from the raw memory at the specified offset.
   func load(fromByteOffset: Int = 0, as: Int64.Type, endianness: Endianness? = nil) -> Int64
+
+  /// Returns a value constructed from the raw memory at the specified offset.
   func load(fromByteOffset: Int = 0, as: UInt.Type, endianness: Endianness? = nil) -> UInt
+
+  /// Returns a value constructed from the raw memory at the specified offset.
   func load(fromByteOffset: Int = 0, as: Int.Type, endianness: Endianness? = nil) -> Int
 
+
+  /// Returns a value constructed from the raw memory at the specified offset.
   func load(fromByteOffset: Int = 0, as: Float32.Type, endianness: Endianness? = nil) -> Float32
+
+  /// Returns a value constructed from the raw memory at the specified offset.
   func load(fromByteOffset: Int = 0, as: Float64.Type, endianness: Endianness? = nil) -> Float64
 
   // available on platforms that support `Float16`
+  /// Returns a value constructed from the raw memory at the specified offset.
   func load(fromByteOffset: Int = 0, as: Float16.Type, endianness: Endianness? = nil) -> Float16
+
   // available on platforms that support `Float80`
+  /// Returns a value constructed from the raw memory at the specified offset.
   func load(fromByteOffset: Int = 0, as: Float80.Type, endianness: Endianness? = nil) -> Float80
+
   // available on platforms that support `UInt128`
+  /// Returns a value constructed from the raw memory at the specified offset.
   func load(fromByteOffset: Int = 0, as: UInt128.Type, endianness: Endianness? = nil) -> UInt128
+
   // available on platforms that support `Int128`
+  /// Returns a value constructed from the raw memory at the specified offset.
   func load(fromByteOffset: Int = 0, as: Int128.Type, endianness: Endianness? = nil) -> Int128
 }
 ```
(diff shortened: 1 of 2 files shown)


proposals/nnnn-rawspan-safe-loading-api.md
```diff
@@ -0,0 +1,663 @@
...
+
+In contrast, `Range<Int>` could not conform to `ConvertibleFromRawBytes`, even though on the surface it has the same composition as `Point`. There is a semantic constraint between two two stored properties of `Range`: `lowerBound` must be less than or equal to `upperBound`. This makes it unable to conform to `ConvertibleFromRawBytes`.
+
+Other examples of types that cannot conform to `ConvertibleFromRawBytes` are `UnicodeScalar` (some bit patterns are invalid,) a hypothetical UTF8-encoded `SmallString` (the sequencing of the constituent bytes matters for validity,) and `UnsafeRawPointer`. The case of pointers is illuminating: the semantic validity of a value is unknown until runtime, since the runtime environment determines the actual set of valid values.
+
+The compiler cannot enforce the semantic requirements of `ConvertibleFromRawBytes`, therefore types outside the standard library can only conform with an unsafe conformance.
+
+```swift
+extension myType: @unsafe ConvertibleFromRawBytes {}
```
apple-fcloutier: `MyType`?

glessard: Indeed.
```suggestion
extension MyType: @unsafe ConvertibleFromRawBytes {}
```

proposals/nnnn-rawspan-safe-loading-api.md
```diff
@@ -0,0 +1,733 @@
...
+  subscript(_ byteOffset: Int) -> UInt8 { get set }
+
+  @unsafe
+  subscript(unchecked byteOffset: Int) -> UInt8 { get set }
+}
+```
+
+##### `MutableRawSpan` and `OutputRawSpan`
+
+`MutableRawSpan` will gain a new overloads of `storeBytes()`:
```
xwu:
```suggestion
`MutableRawSpan` will gain new overloads of `storeBytes()`:
```

Assuming it wasn't accidental that you have more than one overload below :)

proposals/nnnn-rawspan-safe-loading-api.md
```diff
@@ -0,0 +1,733 @@
...
+    as type: T.Type,
+    _ byteOrder: ByteOrder
+  ) where T: ConvertibleToRawBytes & BitwiseCopyable & FixedWidthInteger
+  
+  mutating func storeBytes<T>(
+    repeating repeatedValue: T, count: Int, as type: T.Type
+  ) where T: BitwiseCopyable
+}
+```
+The existing `storeBytes` function constrained to `T: BitwiseCopyable` does not need to be marked `@unsafe`;  all the underlying bytes are already initialized, and therefore optimizations cannot result in uninitialized bytes. The addition of a repeating variant corrects an omission.
```
xwu: Is it an oversight here that there is a byte order overload, a repeated overload, but not _also_ a repeated-with-byte-order overload? (And same question below for `OutputRawSpan.append`.)

proposals/nnnn-rawspan-safe-loading-api.md
```diff
@@ -0,0 +1,733 @@
...
+```
+The existing `append` function constrained to just `T: BitwiseCopyable` will be marked `@unsafe`.
+
+##### Loading and storing in-memory types
+
+The memory layout of many of the types eligible for `ConvertibleFromRawBytes` and/or `ConvertibleToRawBytes` is not guaranteed to be stable across compiler and library versions. This is not an issue for the use case envisioned for these API, where data is sent among running processes, or stored for later use by the same process. For more elaborate needs such as serializing for network communications or file system storage, the API proposed here can only be considered as a building block.
+
+##### `Span` and `MutableSpan`
+
+`Span` will have a new initializer `init(viewing: RawSpan)` to allow viewing a range of untyped memory as a typed `Span`, when `Span.Element` conforms to `ConvertibleFromRawBytes`. These conversions will check for alignment and bounds.
```
xwu: > will check for alignment and bounds.

Can you append a clause here as to what happens if the checks fail? Assuming (since you didn't contradict the comment in the pitch thread) that it will trap at runtime for alignment. In the pitch thread, you said you hadn't decided whether this API would trap or truncate for a failed bounds check--have you decided now?

proposals/nnnn-rawspan-safe-loading-api.md
```diff
@@ -0,0 +1,733 @@
...
+
+  @_lifetime(copy mutableBytes)
+  init(_ mutableBytes: consuming MutableRawSpan)
+    where Element: ConvertibleFromRawBytes
+}
+```
+
+The conversions from `RawSpan` to `Span` only support well-aligned views with the native byte order. The [swift-binary-parsing][swift-binary-parsing] package provides a more fully-featured `ParserSpan` type for use cases beyond reinterpreting memory in-place.
+
+The existing `bytes` and `mutableBytes` accessors will have safe overloads for when `Element` conforms to `ConvertibleToRawBytes`.
```
xwu: In the pitch thread, you said:

> This being said, we might want to use a different name for the old unsafe ones, because there is no indicator unless one uses “-strict-memory-safety”

Did that mean you do want to incorporate the renaming as part of this proposal, or as future direction?

proposals/nnnn-rawspan-safe-loading-api.md
```diff
@@ -0,0 +1,733 @@
...
+  mutating func append<T, E: Error>(
+    elements n: Int,
+    as type: T.self,
+    initializingWith initializer: (inout OutputSpan<T>) throws(E) -> Void
+  ) throws(E) where T: ConvertibleToRawBytes & BitwiseCopyable
+}
+```
+`append(byteCount:as:initializingWith)` will perform bounds-checking and alignment-checking before executing the closure.
+
+Similarly,, `OutputSpan` will provide a way to initialize a portion of its uninitialized storage using an `OutputRawSpan`, when its `Element` type conforms to `ConvertibleFromRawBytes`.
```
xwu:
```suggestion
Similarly, `OutputSpan` will provide a way to initialize a portion of its uninitialized storage using an `OutputRawSpan`, when its `Element` type conforms to `ConvertibleFromRawBytes`.
```

proposals/nnnn-rawspan-safe-loading-api.md
```diff
@@ -0,0 +1,733 @@
...
+```swift
+extension OutputSpan where Element: ConvertibleFromRawBytes {
+  @_lifetime(copy self)
+  mutating func append<E: Error>(
+    elements n: Int,
+    initializingWith initializer: (inout OutputRawSpan) throws(E) -> Void
+  ) throws(E)
+}
+```
+`append(elements:initializingWith:)` will perform bounds-checking before executing the closure and, after it returns, will ensure that the number of bytes initialized is correct for the the type of `Element`.
```
xwu: Similarly here, I assume this means if any checks fail, it will be a runtime trap?

proposals/nnnn-rawspan-safe-loading-api.md
```diff
@@ -0,0 +1,733 @@
...
+extension InlineArray: ConvertibleFromRawBytes where Element: ConvertibleFromRawBytes {}
+
+extension CollectionOfOne: ConvertibleToRawBytes where Element: ConvertibleToRawBytes {}
+extension CollectionOfOne: ConvertibleFromRawBytes where Element: ConvertibleFromRawBytes {}
+
+extension ClosedRange: ConvertibleToRawBytes where Bound: ConvertibleToRawBytes {}
+extension Range: ConvertibleToRawBytes where Bound: ConvertibleToRawBytes {}
+extension PartialRangeFrom: ConvertibleToRawBytes where Bound: ConvertibleToRawBytes {}
+extension PartialRangeFrom.Iterator: ConvertibleToRawBytes
+  where Bound: ConvertibleToRawBytes {}
```
xwu: Question: is there a reason why only this iterator and not other range iterators, or is this an editorial thing?

xwu (commented):
Minor editorial comments and follow-up questions based on pitch feedback. I think once addressed it's ready for review discussion going into the weekend.

proposals/nnnn-rawspan-safe-loading-api.md
glessard: Some of the other range iterators contain an `Optional`, which can't be `ConvertibleToRawBytes` (when the value is `nil`, some of the bytes can be left uninitialized). `Range`'s iterator is simply `IndexingIterator`, which should also have a conditional conformance which I forgot to include.

Separately, `AnyObject` won't conform because the way it is defined prevents its conformance. That will need to wait for some further compiler support.

proposals/nnnn-rawspan-safe-loading-api.md
glessard: I've added these clarifications to the doc-comments.

proposals/nnnn-rawspan-safe-loading-api.md
glessard: I added a paragraph in future directions.

proposals/nnnn-rawspan-safe-loading-api.md
glessard: For the bounds-checking, yes. For the number of elements initialized, we calculate the number of elements after returning; being strict here would mean that thrown errors would often cause crashes. I have clarified this in the doc-comment.

proposals/nnnn-rawspan-safe-loading-api.md
glessard: Whatever I thought at first no longer feels valid! Fixed.

proposals/nnnn-rawspan-safe-loading-api.md
glessard: `IndexingIterator` can easily have internal padding even if both its associated types were `ConvertibleToRawBytes`, so it goes away again. Generic types are tricky.

PR #3191, opened 2026-03-17
SE-0521: Add Review header with links to pitch and review discussions

dempseyatgithub:
The review period for SE-0521 has started but the proposal does not appear on the evolution dashboard because it is missing the Review header.

This PR adds the review header with links to the pitch and review threads in the forums.

// @allevato @Jumhyn


Jumhyn (approved):
Thanks!

PR #3199, opened 2026-03-19
Normalize discussion names in Review header field

dempseyatgithub:
The names of the discussion links in the Review header should not be capitalized, the links should not be separated with commas, just a space.

This PR normalizes the names/spacing of the Review header field.

Normalizing helps aid the generation of metadata. The evolution dashboard may soon begin linking to the review thread of proposals in Active Review.

// @mikaelacaron


rjmccall (approved):
Thanks!

PR #3200, opened 2026-03-19
Add acceptance discussion link for SE-0501, normalize links

dempseyatgithub:
Add the acceptance discussion link for SE-0501.

Also normalize all links in the Review header to match the standardized thread naming and format.

// @daveyc123


rjmccall (approved):
Thanks!

PR #3208, opened 2026-03-24
Per-test-case repetition behavior

harlanhaskins:
This proposal would change the test repetition behavior to only repeat those test cases that fail, rather than repeating all tests when any fails.

[View rendered proposal](https://github.com/harlanhaskins/swift-evolution/blob/harlan/for-repeats-sake/proposals/testing/NNNN-per-test-case-repetitions.md)


briancroom:
Thanks, Harlan! Per discussion with the workgroup, we're going to begin review of this now (as ST-0024.)

PR #3239, opened 2026-04-13
[ST-0025] Tag-based test execution filtering

gmedori:
For convenience, here's a link to the implementation of the feature described in this proposal: https://github.com/swiftlang/swift-testing/pull/1531

### DIFF
--- a/proposals/testing/0025-tag-based-test-execution-filtering.md
+++ b/proposals/testing/0025-tag-based-test-execution-filtering.md
@@ -3,7 +3,7 @@
 - Proposal: [ST-0025](0025-tag-based-test-execution-filtering.md)
 - Authors: [Gustavo Medori](https://github.com/gmedori)
 * Review Manager: [Paul LeMarquand](https://github.com/plemarquand)
-* Status: **Active Review (May 21 - June 4, 2026)**
+* Status: **Active Review (Jun 11 - June 26, 2026)**
 - Bugs: [swiftlang/swift-testing#591](https://github.com/swiftlang/swift-testing/issues/591), rdar://132989780
 - Implementation: [swiftlang/swift-testing#1531](https://github.com/swiftlang/swift-testing/pull/1531)
 - Review: [pitch](https://forums.swift.org/t/pitch-tag-based-test-execution-filtering/86001)
--- a/proposals/testing/0025-tag-based-test-execution-filtering.md
+++ b/proposals/testing/0025-tag-based-test-execution-filtering.md
@@ -67,7 +67,7 @@ As of [SE-0451](https://github.com/swiftlang/swift-evolution/blob/main/proposals
 @Test func `tag:uiTest`() { /* ... */ }
 ```
 
-To continue to allow matching for such function names, I propose introducing a _separate_ prefix called `id:` which behaves much like the `tag:` prefix in that everything that follows it is a regular expression. It's job, however, is to disambiguate and allow the user a mechanism to explicitly say "match on test symbol names please." 
+To continue to allow matching for such function names, I propose introducing a _separate_ prefix called `id:` which behaves much like the `tag:` prefix in that everything that follows it is a regular expression. Its job, however, is to disambiguate and allow the user a mechanism to explicitly say "match on test symbol names please." 
 
 ```sh
 swift test --skip 'id:tag:uiTest'
@@ -118,7 +118,7 @@ This introduces a new mechanism that can be used by any existing tools to filter
 
 ## Future Directions
 
-Filtering based on tags is quite broad and general purpose. Because you can define any tag to stick on any test or suite, and because tags exists orthogonally to the test graph, you can arbitrarily include/skip any test based solely on the semantics of your tags. However, this change does raise the question of what _else_ we could filter/skip on and how we can be more expressive about it.
+Filtering based on tags is quite broad and general purpose. Because you can define any tag to stick on any test or suite, and because tags exist orthogonally to the test graph, you can arbitrarily include/skip any test based solely on the semantics of your tags. However, this change does raise the question of what _else_ we could filter/skip on and how we can be more expressive about it.
 
 For example, you may wish to filter/skip tests based on protocol conformance and/or inheritance. A suite's ancestor types can be a useful, and perhaps more natural, signal indicating whether it should run in a given context or not because the ancestor types carry with them behaviors and contracts that have powerful semantic meaning. In the future, we may seek to expand the prefix operators we allow beyond just `tag:`.
 
--- a/proposals/testing/0025-tag-based-test-execution-filtering.md
+++ b/proposals/testing/0025-tag-based-test-execution-filtering.md
@@ -4,6 +4,7 @@
 - Authors: [Gustavo Medori](https://github.com/gmedori)
 * Review Manager: [Paul LeMarquand](https://github.com/plemarquand)
 * Status: **Active Review (May 21 - June 4, 2026)**
+- Bugs: [swiftlang/swift-testing#591](https://github.com/swiftlang/swift-testing/issues/591), rdar://132989780
 - Implementation: [swiftlang/swift-testing#1531](https://github.com/swiftlang/swift-testing/pull/1531)
 - Review: [pitch](https://forums.swift.org/t/pitch-tag-based-test-execution-filtering/86001)
 
--- a/proposals/testing/0025-tag-based-test-execution-filtering.md
+++ b/proposals/testing/0025-tag-based-test-execution-filtering.md
@@ -1,11 +1,11 @@
 # Tag-based Test Execution Filtering
 
-- Proposal: [ST-NNNN](nnnn-tag-based-test-execution-filtering.md)
+- Proposal: [ST-0025](0025-tag-based-test-execution-filtering.md)
 - Authors: [Gustavo Medori](https://github.com/gmedori)
 - Review Manager: TBD
 - Status: **Awaiting review**
 - Implementation: [swiftlang/swift-testing#1531](https://github.com/swiftlang/swift-testing/pull/1531)
-- Review: ([pitch (not yet live)](https://forums.swift.org/...))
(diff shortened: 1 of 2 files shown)


proposals/testing/nnnn-tag-based-test-execution-filtering.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Integration with Supporting Tools
+
+If a codebase has test functions or suites that contain the string `tag:`, then any filtering/skipping arguments that begin with `tag:` behave differently with this proposal. They are treated as tags rather than regular expressions.
+
+It's worth noting that string `tag:` can only appear in symbol names when using raw identifiers as described above. I feel that this scenario is relatively rare and that this improvement is worth the edge case.
+
+## Future Directions
+
+N/A
```
grynspan: Consider whether there are other aspects of tests on which devs may want to filter?

proposals/testing/nnnn-tag-based-test-execution-filtering.md
```diff
@@ -0,0 +1,111 @@
...
+ 
+ As of [SE-0451](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0451-escaped-identifiers.md), Swift has raw identifiers which means the following is valid Swift:
+
+```swift
+@Test func `tag:uiTest`() { /* ... */ }
+```
+
+To continue to allow matching for such function names, we will also allow the colon character to be escaped such that the entire argument will be treated as a single regular expression:
+
+```
```
grynspan, line 74:
```suggestion
```sh
```

proposals/testing/nnnn-tag-based-test-execution-filtering.md
```diff
@@ -0,0 +1,111 @@
...
+swift test --skip 'tag\:uiTest'
+```
+
+The example above would behave as though the string `tag:uiTest` were passed as a regular expression, omitting the escaping backslash in the final regular expression.
+
+> **Note**: Most shells treat the backslash character `\` as a special character used for escaping. In order for the application to receive it, the argument needs to either be wrapped in quotes like `'tag\:uiTest'`, or the backslash itself needs to be escaped, `tag\\:uiTest`.
```
grynspan:
```suggestion
> [!NOTE]
> Most shells treat the backslash character `\` as a special character used for escaping. In order for the application to receive it, the argument needs to either be wrapped in quotes like `'tag\:uiTest'`, or the backslash itself needs to be escaped, `tag\\:uiTest`.
```

See https://github.com/orgs/community/discussions/16925

grynspan: Should we just make it a policy that `\` escapes the next character? If not, consider adding some discussion in this area to **Alternatives Considered**.

Or, if that's impractical or undesirable, should we document how to write `\:` verbatim (i.e. we probably also need to allow escaping `\` itself)?

proposals/testing/0025-tag-based-test-execution-filtering.md
```diff
@@ -0,0 +1,111 @@
...
+
+> **Note**: Most shells treat the backslash character `\` as a special character used for escaping. In order for the application to receive it, the argument needs to either be wrapped in quotes like `'tag\:uiTest'`, or the backslash itself needs to be escaped, `tag\\:uiTest`.
+
+## Source Compatibility
+
+The change is an implementation detail and makes no changes to source compatibility.
+
+## Integration with Supporting Tools
+
+If a codebase has test functions or suites that contain the string `tag:`, then any filtering/skipping arguments that begin with `tag:` behave differently with this proposal. They are treated as tags rather than regular expressions.
```
grynspan, line 111: This discussion might belong in **Source Compatibility** rather than this section because it doesn't really affect e.g. VS Code or Xcode integrations.

proposals/testing/0025-tag-based-test-execution-filtering.md
```diff
@@ -0,0 +1,111 @@
...
+
+```
+swift test --skip tag:uiTest --skip tag:integrationTest
+```
+
+It is not currently possible to create a filter/skip based on _all_ tags specified, as opposed to _any_ of the tags specified. In other words, supplying multiple `--filter`/`--skip` options is an **or** operation, not an **and** operation.
+
+### Handling Raw Identifiers
+ 
+ As of [SE-0451](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0451-escaped-identifiers.md), Swift has raw identifiers which means the following is valid Swift:
```
allevato: You should also consider what needs to happen (if anything different) if the tag itself is a raw identifier (unless anything else in Swift Testing prohibits that, which I don't know off the top of my head):

```swift
@Test(.tags(.`some tag with spaces`))
func `my test`() { ... }
```

I think I would expect to write `swift test --skip 'tag:some tag with spaces'` here.

grynspan: Then there's an open question about whether or not developers should specify backticks (which also affect shell processing).

proposals/testing/nnnn-tag-based-test-execution-filtering.md
gmedori: Ah fair point! I actually wrote this a little while ago to address an open github issue and am just putting up the PR now, and I didn't stop to think about future directions. Lemme give it a little thought.

proposals/testing/0025-tag-based-test-execution-filtering.md
gmedori: Yeah I read through that last night and think I agree.

proposals/testing/nnnn-tag-based-test-execution-filtering.md
gmedori: _Personally_ I think that if we want to make that a policy, the policy should exist at SPM level for every command. I'm not sure it's useful to have a policy that only applies to swift testing. Probably skews impractical IMO. Think this is something we want to pursue anyway?

> Or, if that's impractical or undesirable, should we document how to write \: verbatim (i.e. we probably also need to allow escaping \ itself)?

Do you mean in the usage description of the `--skip` and `--filter` options?

proposals/testing/0025-tag-based-test-execution-filtering.md
gmedori: ha, yep that complicates things. I think the most correct answer here is to write it as @allevato put it. Involving backticks opens the command up to different interpretations in different shells, where as single-quoting is relatively universal, even going back to shells like `sh` and `csh`.

proposals/testing/nnnn-tag-based-test-execution-filtering.md
gmedori: Added some thoughts in 61e2f4b2

proposals/testing/nnnn-tag-based-test-execution-filtering.md
gmedori: Resolved in d80d77b7

proposals/testing/nnnn-tag-based-test-execution-filtering.md
gmedori: Resolved in d80d77b7

proposals/testing/0025-tag-based-test-execution-filtering.md
gmedori: Added some more discussion around this scenario in 90054874

proposals/testing/0025-tag-based-test-execution-filtering.md
gmedori: I moved this discussion to **Source Compatibility** and added some clearer thoughts around tool integrations in 90054874

proposals/testing/0025-tag-based-test-execution-filtering.md
allevato: As a parallel, when using raw identifiers in the `-module-alias` or `-module-abi-name` command line flags of the compiler, you _don't_ need to include the backticks: you'd write `-module-abi-name 'foo bar'`, not ``-module-abi-name '`foo bar`'``. So I think it's reasonable to have the same policy here.

In other words, the backticks are required to delimit the identifier in _source code_ (and they also appear in the mangling of the symbol for compatibility reasons), but it's reasonable to drop them elsewhere.

grynspan: I don't have an opinion either way, but we should be clear about it. (And if the developer _does_ type the backticks, does that make it _wrong_?)

gmedori: > if the developer does type the backticks, does that make it _wrong_?

I could go either way on this, so IMO we should just fall back to precedent. From what @allevato said:

> you'd write `-module-abi-name 'foo bar'`, not ```-module-abi-name '`foo bar`'```

it _sounds_ like using backticks would be incorrect. So I'd opt to not do any special handling around stripping them for the user. But please correct me if I'm misinterpreting.

allevato: If you include the backticks in `-module-abi-name`, it looks like it checks whether the symbol needs to be escaped (which it does, because it contains a space) and then inserts a second pair of backticks, so the mangled symbol ends up being ` ``foo bar`` `. This is actually wrong because raw identifiers are not supposed to be able to contain backticks (syntactically, at least). The frontend should prevent that instead of letting it through.

So maybe that flag wasn't the best one to cite, but my vote here would be "don't require backticks and fail if the user provides them".

gmedori: I think this might be a common enough occurrence that there's actually value in detecting whether backticks were supplied and presenting a targeted error message. I'll work on adding that to the implementation later today.

gmedori: I was able to sort this out! Apologies for the delay. I've update the proposal with my suggestion here: a2b375a4

tl;dr: we'll alert the user that the symbol name shouldn't have backticks surrounding and strip them transparently. So it'll match what the user probably intended.

PR #3244, opened 2026-04-14
proposals: propose protocol metatype extensions

compnerd:
Propose protocol metatype extensions

### DIFF
--- a/proposals/NNNN-protocol-metatype-extensions.md
+++ b/proposals/NNNN-protocol-metatype-extensions.md
@@ -1,0 +1,398 @@
+# Protocol Metatype Extensions
+
+* Proposal: [SE-NNNN](NNNN-protocol-metatype-extensions.md)
+* Authors: [Saleem Abdulrasool](https://github.com/compnerd)
+* Review Manager: TBD
+* Status: **Awaiting review**
+* Implementation: [swiftlang/swift#88445](https://github.com/swiftlang/swift/pull/88445)
+* Upcoming Feature Flag: `ProtocolMetatypeExtensions`
+
+## Summary of changes
+
+Allows `extension P.Protocol`, where `P` is a protocol, to declare static members that live on the protocol itself and are not inherited by conforming types.
+
+## Motivation
+
+Swift protocols can have static members declared in extensions:
+
+```swift
+protocol Plugin {
+  static var name: String { get }
+  func run()
+}
+
+extension Plugin {
+  static var searchPaths: [String] { ["/usr/lib/plugins", "~/.plugins"] }
+}
+```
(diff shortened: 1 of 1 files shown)


proposals/NNNN-protocol-metatype-extensions.md
```diff
@@ -0,0 +1,447 @@
...
+members need to dispatch on `self`, which requires them to be instance members.
+
+By using instance members for `P.Protocol` extensions in this proposal, we
+establish a uniform model: metatype extensions add instance members to metatype
+values. The only difference between `P.Protocol` and `P.Type` extensions is how
+many values the metatype has, one for `P.Protocol`, many for `P.Type`.
+
+We consider this a natural future direction but out of scope for this proposal.
+
+### Protocol metatype conformances
```
allevato, line 335: We discussed this proposal in the Language Steering Group this week. One of the questions we had was whether this proposal was a complete enough story to do what folks might want to do with things like COM interop (or even non-COM use cases). This would be adding a capability to the language where we can use the protocol as a namespace, but as currently written, it ends there—it doesn't permit the kind of generic programming that a user might otherwise expect.

This section points out that a natural thing to want to do with these extensions would be to pass the metatype to a generic function that would operate on that IID. And without the ability to conform metatypes to protocols, there's no way to express that today:

```swift
func someFunc<T: /* ??? */>(_ comInterface: T) {
  let iid = T.IID
  /* do something with iid */
}
someFunc(IDispatch.self)
```

Is this something you foresee users of COM interop wanting to do, and if so, does it follow naturally from what is being proposed here? Did you have plans to work on this space as well, or is this more of an open question?

compnerd: Yeah, this is something that I could see as being valuable (and eventually needed). I think that that this would be something to revisit in the future once more of the COM infrastructure is built out. But, yes, I do feel like it would flow from this work. As to the work itself, that would be more of an open question and something that would come back up when the WinRT layer starts getting looked at.

allevato:
The Language Steering Group discussed this last week and we think there are some unanswered design questions that need to be figured out before this can be run.

The main concern surrounded the inability to operate on metatypes generically. Even though you acknowledge this as a future direction, we want to avoid backing ourselves into a language design corner if we allow folks to inject members with the current "namespace" approach, especially around the choice that `Q.Protocol` doesn't inherit members from `P.Protocol` when `Q: P`.

For the COM interop use cases you had in mind like generating IIDs for interface protocols, we were generally more comfortable with just having those be synthesized via compiler magic for the time being, rather than add a metatype feature that isn't fully realized.

PR #3249, opened 2026-04-16
Vision: add a vision for COM interop

compnerd:
Add an initial draft of the COM interop vision. This should describe the high level shape that we want the feature to take. Adding this early to allow editorial tweaks and collaboration on the content from others.


rjmccall:
Most of this vision seems to be about things that are firmly in the LSG's evolution domain. We'd be happy to collaborate with the PSG on it.

PR #3260, opened 2026-04-22
[SE-0525] remove default parameters from load and storeBytes

glessard:
- Remove default parameters, from `load` and `storeBytes` functions, as discussed in the LSG proposal acceptance meeting.
- Add an “alternatives considered” paragraph to document this.
- Added a missing "Implications on adoption" section.
- Fixed many inconsistencies in doc-comments and between sections.

- Had erroneously proposed `@unsafe` attribute on conformances, instead of `@unchecked`.
- Had erroneously proposed an argument label of `MutableSpan.init(bytes: MutableRawSpan)` instead `MutableSpan.init(mutableBytes: MutableRawSpan)`.
- The first argument label of `OutputSpan.append(elements: Int, initializingWith: (OutputRawSpan) -> R)` is changed to `elementCount` for clarity.

### DIFF
--- a/proposals/0525-rawspan-safe-loading-api.md
+++ b/proposals/0525-rawspan-safe-loading-api.md
@@ -114,7 +114,7 @@ The `load(as:)` functions will not have equivalents with unchecked byte offset. 
 
 ##### Subscripts for the `RawSpan` family
 
-As a convenience for the specific case of `UInt8`, we will define subscripts for `RawSpan`, `MutableRawSpan`, and `OutputRawSpan`, similar to the existing `Span`, `MutableSpan` and `OutputSpan` subscripts:
+As a convenience for the specific case of `UInt8`, we will define subscripts for `RawSpan`, `MutableRawSpan` and `OutputRawSpan`, similar to the existing `Span`, `MutableSpan` and `OutputSpan` subscripts:
 ```swift
 extension RawSpan {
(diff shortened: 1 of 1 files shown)


xwu (commented):
Nice clean-ups!

I discussed the API-level bugfixes with the group. They pointed out that we have slightly different spellings for `append(elementCount:...)` here versus `append(addingCount:...)` for the in-flight `UniqueArray` and `RigidArray` that should probably be reconciled, so I'll merge with acceptance and prompt the community for a focused mini-review on the API bugfixes (the remaining ones being pretty uncontroversial).

proposals/0525-rawspan-safe-loading-api.md
```diff
@@ -599,6 +599,9 @@ extension OutputRawSpan {
   ///     must be greater than or equal to zero, and less than `byteCount`.
   @unsafe
   subscript(unchecked byteOffset: Int) -> UInt8 { get set }
+
+  /// The offsets valid for subscripting the span, in ascending order.
+  public var byteOffsets: Range<Int>
```
xwu, line 550: Is this always `0..<byteCount`?

glessard: Yes. RawSpan and MutableRawSpan already had this property, so this was a gap.

PR #3264, opened 2026-04-24
Update SE-0337 to reflect @preconcurrency working on decls in Swift 6

a-viv-a:
Prompted by bringing up swiftlang/swift#88637, my understanding is that the intention is for `@preconcurrency` to apply to declarations and downgrade errors in Swift 6 the same way it did in Swift 5. In most cases, this is already the way preconcurrency is implemented in the compiler, and cases where it doesn't is a bug and an issue for incremental adoption, since Swift 6 consumers of your types would receive errors, even if you use `@preconcurrency`. SE-0337 currently doesn't reflect this intention or implementation reality, which is confusing when trying to understand this feature.

We should update SE-0337 to be explicit about `@preconcurrency` working (in ways other than`@preconcurrency import`) in Swift 6.


rjmccall:
The LSG approved this change a few weeks ago. No need for further review, this is just making the de facto rules explicit.

PR #3287, opened 2026-05-09
[Proposal] Enhancing-Async{Throwing}Stream

NotTheNHK:


NotTheNHK:
@rjmccall The "ABI compatibility" section isn’t done yet, as I’m not fully certain what the impact is. Could you have a look at it?

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
```diff
@@ -0,0 +1,328 @@
...
+The `@Sendable` requirement on the `onCancel` closure is removed and replaced with the `sending` keyword.
+
+```swift
+let locationStream = Async{Throwing}Stream { // consistent API
+  ...
+} onCancel: {
+  ...
+}
+
+for {try} await location in locationStream { // executes on the caller's actor
```
ktoso: It's more complicate than that though, isn't it? The AsyncSequence's next() still isn't `nonisolated(nonsending)` so there will be hops here involved. The iteration ofc comes back to the surrounding actor but there's more nuance here.

The example should also perhaps show the code inside an actor so the "caller's actor" is more clear.

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
```diff
@@ -0,0 +1,328 @@
...
+}
+```
+
+### Unfolding Initializer
+
+This proposal adds an `onCancel` parameter to the unfolding initializer of `AsyncThrowingStream`, aligning it with `AsyncStream` and with the original variant proposed in SE-0314.
+
+Additionally, this proposal adopts `nonisolated(nonsending)`. As described in [SE-0461](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0461-async-function-isolation.md), this allows the `produce` closure to run on the caller’s actor, avoiding unnecessary actor hops.
+
+The `@Sendable` requirement on the `onCancel` closure is removed and replaced with the `sending` keyword.
```
ktoso: Let's move this to detailed design, we're not explaining or showing it here yet

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
```diff
@@ -0,0 +1,328 @@
...
+
+```swift
+let locationStream = Async{Throwing}Stream { // consistent API
```
ktoso: Unclear comment, remove?

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
```diff
@@ -0,0 +1,328 @@
...
+for {try} await location in locationStream { // executes on the caller's actor
+  ...
+}
+```
+
+### Stream termination when its continuation is discarded
+
+The continuation-based variant is updated to track outstanding references to the stream’s continuation, including the continuation itself and any copies of it. When the last reference to the continuation is discarded, the stream is canceled. 
+
+The change is staged in via an upcoming feature flag (`StreamContinuationTracking`).
```
ktoso: This needs more explanation I think? The code listing should include `// with StreamContinuationTracking` perhaps to make it clear

ktoso: The name should IMHO be the full type name and be more explicit, the "Tracking" doesn't tell me that this changes cancellation behavior.

Maybe `AsyncStreamCancelOnContinuationDeinit` ? We can talk about it in review.

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
```diff
@@ -0,0 +1,328 @@
...
+```swift
+// AsyncStream
+
+extension AsyncStream: Hashable {
+  public static func == (lhs: Self, rhs: Self) -> Bool {
+    return lhs.context === rhs.context
+  }
+
+  public func hash(into hasher: inout Hasher) {
+    hasher.combine(ObjectIdentifier(self.context))
```
ktoso:
```suggestion
    // ...
```

no need to  include full impls in propsal

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
```diff
@@ -0,0 +1,328 @@
...
+```
+
+`Hashable` conformance:
+
+```swift
+// AsyncStream
+
+extension AsyncStream: Hashable {
+  public static func == (lhs: Self, rhs: Self) -> Bool {
+    return lhs.context === rhs.context
```
ktoso:
```suggestion
    // ... 
```

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
```diff
@@ -0,0 +1,328 @@
...
+
+  public static func makeStream(
+    of elementType: Element.Type = Element.self,
+    throwing failureType: Failure.Type = Failure.self,
+    bufferingPolicy limit: Continuation.BufferingPolicy = .unbounded
+  ) -> (stream: AsyncThrowingStream<Element, Failure>, continuation: AsyncThrowingStream<Element, Failure>.Continuation)
+}
+```
+
+`Hashable` conformance:
```
ktoso, line 261: Please explain that the hashable impl is identity based, rather than show impl

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
```diff
@@ -0,0 +1,328 @@
...
+  }
+
+  public func hash(into hasher: inout Hasher) {
+    hasher.combine(ObjectIdentifier(self.context))
+  }
+}
+
+extension AsyncThrowingStream.Continuation.BufferingPolicy: Hashable {}
+
+extension AsyncThrowingStream.Continuation.YieldResult: Equatable, Hashable where Element: Equatable, Element: Hashable {}
```
ktoso, line 294:
```suggestion
extension AsyncThrowingStream.Continuation.YieldResult: Equatable, Hashable where Element: Hashable {}
```

Hashable refines Equatable already

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
```diff
@@ -0,0 +1,328 @@
...
+
+## Source compatibility
+
+This proposal changes the behavior around stream termination when the stream’s continuation is discarded. To avoid silently changing behavior, this change is gated behind an upcoming feature flag (`StreamContinuationTracking`).
+
+The `sending` keyword on `onCancel` will allow a wider range of functions and closures to be passed to it.
+
+## ABI compatibility
+
+Adopting `nonisolated(nonsending)` for `produce` and replacing `@Sendable` on `onCancel` is an ABI change. // TODO: Finish this
```
ktoso: Basically:

This isn't a "change" as much as it is an addition of new overloads; the old apis must remain available, so this is ABI additive.

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
```diff
@@ -0,0 +1,328 @@
...
+
+## Implications on adoption
+
+Terminating the stream implicitly when the stream’s continuation is discarded would break code that relies on the current behavior, for example to create an indefinite suspension point.
```
ktoso: This needs to explain "so that's why it is guarded with the upcoming feature flag"

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
```diff
@@ -0,0 +1,328 @@
...
+1. Typed throws support for `AsyncThrowingStream`.
+2. Update the unfolding initializer by adopting `nonisolated(nonsending)` and replacing `onCancel`’s `@Sendable` requirement with `sending`.
+3. Terminate the stream when its continuation is discarded.
+4. `Hashable` conformance for `Async{Throwing}Stream` and nested types.
+
+## Motivation
+
+### Typed Throws
+
+Thrown errors are type-erased to `any Error`, requiring additional boilerplate to preserve the thrown error's type and integrate into typed contexts.
```
ktoso:
```suggestion
Currently errors are type-erased to `any Error` when AsyncStream is finished with an error (`finish(throwing:)`), requiring additional boilerplate to preserve the thrown error's type and integrate into typed contexts.
```

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
```diff
@@ -0,0 +1,328 @@
...
+// AsyncThrowingStream
+public init(
+  unfolding produce: @escaping () async throws -> Element?, 
+  onCancel: (@Sendable () -> Void)? = nil
+)
+```
+
+However, the `AsyncThrowingStream` variant was never implemented with an `onCancel` parameter, creating a discrepancy between the two APIs. 
+
+Furthermore, [SE-0338](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0338-clarify-execution-non-actor-async.md#proposed-solution) clarified the execution semantics of `nonisolated` asynchronous functions by specifying that such functions formally run on the Global Concurrent Executor (GCE), potentially introducing unnecessary actor hops.
```
ktoso:
```suggestion
Furthermore, [SE-0338](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0338-clarify-execution-non-actor-async.md#proposed-solution) clarified the execution semantics of `nonisolated` asynchronous functions by specifying that such functions formally run on the global concurrent executor, potentially introducing unnecessary actor hops. 
```

"GCE" is not a phrase we use

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
NotTheNHK: This is definitely a placeholder name, open for any name suggestions.

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
NotTheNHK: We get a compiler error if we don’t explicitly do it for each protocol. I was surprised by it too.

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
NotTheNHK: Thanks for the info!

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
NotTheNHK: Doesn’t the synthesized for await-in loop implementation use the next(isolation:) variant? I think this is only an issue when explicitly calling next().

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
NotTheNHK: I got an error similar to this: "error: conditional conformance of type 'S<Value>' to protocol 'B' does not imply conformance to inherited protocol 'A'".
```swift
protocol A {}

protocol B: A {}

struct S<Value> {}

extension S: B where Value: B {} // error ...
```
I don't think this is a bug either.

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
NotTheNHK: I expanded this a little further to this: 
Currently errors are type-erased to `any Error` when `AsyncThrowingStream` is finished with an error (`finish(throwing:)` or when the `unfolding` closure throws), requiring additional boilerplate to preserve the thrown error's type and integrate into typed contexts.

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
NotTheNHK: I will change it to `AsyncStreamCancelOnContinuationDeinit`, but let's see what alternatives review comes-up with.

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
FranzBusch: Yes that's correct. The `next(isolation: )` method should be used in the synthesized `for await` loops and should avoid any hops if implemented by the underlying `AsyncSequence`

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
NotTheNHK: yes.

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
NotTheNHK: I think we should keep it as is. This mirrors the three motivation paragraphs well. It also doesn’t leave the reader confused about why the motivation section explains execution semantics while the proposed solution section doesn’t mention it at all.

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
NotTheNHK: We could add an execution semantics section to the documentation later on if we find the need to explain this further.

NotTheNHK:
@ktoso Could we backdeploy the `onCancel` change? I expect this will be brought up in review.

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
```diff
@@ -0,0 +1,333 @@
...
+This proposal introduces the following changes:
+
+1. Typed throws support for `AsyncThrowingStream`.
+2. Update the unfolding initializer by adopting `nonisolated(nonsending)` and replacing `onCancel`’s `@Sendable` requirement with `sending`.
+3. Terminate the stream when its continuation is discarded.
+4. `Hashable` conformance for `Async{Throwing}Stream` and nested types.
+
+## Motivation
+
+### Typed throws
```
ktoso, line 23: Revisiting this proposal now after a while.

I think this part can be done without SE review, John can confirm soon

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
```diff
@@ -0,0 +1,333 @@
...
+)
+
+// AsyncThrowingStream
+public init(
+  unfolding produce: @escaping () async throws -> Element?, 
+  onCancel: (@Sendable () -> Void)? = nil
+)
+```
+
+However, the `AsyncThrowingStream` variant never implemented the `onCancel` parameter, creating a discrepancy between the two APIs.
```
ktoso, line 90: This part could fall under "implement what was missed by accident"

proposals/NNNN-Enhancing-Async{Throwing}Stream.md
```diff
@@ -0,0 +1,333 @@
...
+    ...
+  }
+}
+```
+
+### Unfolding initializer
+
+This proposal adds the missing `onCancel` parameter to the unfolding initializer of `AsyncThrowingStream`, aligning it with `AsyncStream` and with the original variant proposed in SE-0314.
+
+Additionally, this proposal adopts `nonisolated(nonsending)`. As described in [SE-0461](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0461-async-function-isolation.md), this allows the `produce` closure to run on the caller’s actor, and avoids unnecessary actor hops.
```
ktoso, line 171: We do not generally require proposals to introduce `nonisolated(nonsending)`, though doing so in ABI stable way is pretty tricky

ktoso (commented):
I re-reviewed this a bit and talked with @rjmccall earlier today, maybe we can do a bunch of this without SE review; 

I would be concerned about the behavior change bits though, we've seen changes of behavior in this type have been very troublesome to roll out safely without breaking folks.

NotTheNHK:
@ktoso Hm, we could carefully start with typed throws once the 6.4 process is done. In the worst case, we can always revert and think it over.
What do you think?

ktoso:
typed throws are pretty high risk to break in some places... inference doesn't quite work, but we can give it a shot.

ktoso:
That's the general theme with many of these changes, they're easy on paper but hard to land without breakage, if you're up for some frustrating adventures we can give it a go.

NotTheNHK:
Took the first step here: https://github.com/swiftlang/swift/pull/92464 and immediately ran into an issue with embedded.

PR #3289, opened 2026-05-11
[SE-0526] Revision for cooperative cancellation and deadline composition

phausler:
Major highlights of the changes to the proposal:

1) The DeadlineExceededError was removed in favor of avoiding creating a wrapper error. This makes the error handling considerably more ergonomic and avoids existentials
2) The currentDeadline was removed since that is not needed for composition, but also caused severe ergnomic issues accessing the `any InstantProtocol`.
3) A new interface for CancellationError was added to provide a reasoning for why a task or child task was cancelled. This also includes new cancellation methods for Task, TaskGroup, ThrowingTaskGroup, DiscardingTaskGroup and ThrowingDiscardingTaskGroup.
4) Additional references were added to other related proposals
5) Contrasting analysis on the naming and apis were added comparing other languages (Go and Kotlin).

### DIFF
--- a/proposals/0526-deadline.md
+++ b/proposals/0526-deadline.md
@@ -15,12 +15,6 @@ continuous clock instant representing the deadline by which the operation must
 complete. If the operation completes before the deadline, the function returns 
 the result; if the deadline expires first, the operation is cancelled.
 
-The typed throws signature was altered to avoid an extra error type, in conjunction
-with removing the restriction around the instant requiring the duration type
-to be `Swift.Duration`, the accessor for the current deadline was removed,
-and a new interface on CancellationError was added to handle the reasons
-for why a task or child task is cancelled (including a deadline exceeded reason).
-
 ## Motivation
 
 Asynchronous operations in Swift can run indefinitely, which creates several
@@ -145,16 +139,10 @@ The fundamental entry point for working with deadlines is a single function: `wi
 /// - Returns: The result of the operation if it completes successfully before or after the deadline expires.
 ///
 /// - Throws: The error thrown by the operation
-nonisolated(nonsending) public func withDeadline<Return, Failure: Error, C: Clock>(
+nonisolated(nonsending) public func withDeadline<Return: ~Copyable, Failure: Error, C: Clock>(
   _ expiration: C.Instant,
   tolerance: C.Instant.Duration? = nil,
-  clock: C,
-  body: nonisolated(nonsending) () async throws(Failure) -> Return
-) async throws(Failure) -> Return
-
-nonisolated(nonsending) public func withDeadline<Return, Failure: Error>(
-  _ expiration: ContinuousClock.Instant,
-  tolerance: ContinuousClock.Instant.Duration? = nil,
+  clock: C = ContinuousClock(),
   body: nonisolated(nonsending) () async throws(Failure) -> Return
 ) async throws(Failure) -> Return
 ```
@@ -178,7 +166,11 @@ let (userData, prefsData) = try await (user, prefs)
 
 These absolute deadlines are composable and nestable to any set scope of a deadline. This means that when more than 
 one `withDeadline` is nested the minimum of the expiration is taken. If any nested cases are differing clocks the 
-deadline is adjusted to the minimum by approximating the current deadline with the offset of the proposed expiration.
+deadline expires determined by the clock, so no inter-clock conversions need to be computed. This nesting case works
+by the outer executing with a given deadline expiration while the inner also executes with its own given deadline
+expiration. These two expirations will execute independently to whichever cancels the operation first. Practically
+this means that the expiration then is the minimum of the two deadlines, without needing to compare or calculate 
+between them.
 
 
 ```swift
@@ -214,14 +206,14 @@ Constructing an instant every time is not per se the most terse; so a simple ext
 with the same compositional advantage as the primary entry point.
 
 ```
-nonisolated(nonsending) public func withDeadline<Return, Failure: Error, C: Clock>(
+nonisolated(nonsending) public func withDeadline<Return: ~Copyable, Failure: Error, C: Clock>(
   in timeout: C.Instant.Duration,
   tolerance: C.Instant.Duration? = nil,
-  clock: C,
+  clock: C = ContinuousClock(),
   body: nonisolated(nonsending) () async throws(Failure) -> Return
 ) async throws(Failure) -> Return
 
-nonisolated(nonsending) public func withDeadline<Return, Failure: Error>(
+nonisolated(nonsending) public func withDeadline<Return: ~Copyable, Failure: Error>(
   in timeout: ContinuousClock.Instant.Duration,
   tolerance: ContinuousClock.Instant.Duration? = nil,
   body: nonisolated(nonsending) () async throws(Failure) -> Return
@@ -231,7 +223,7 @@ nonisolated(nonsending) public func withDeadline<Return, Failure: Error>(
 The implementation of this is trivially:
 
 ```
-try await withDeadline(clock.now.advanced(by: timeout), tolerance: tolerance, clock: clock, body: body)
+try await withDeadline(clock.now.advanced(by: timeout), tolerance: tolerance, body: body)
 ```
 
 #### Non-escaping nonisolated(nonsending) operation closure 
@@ -265,7 +257,7 @@ guarantees.
 
 #### Cancellation
 
-This API uses the base cancellation error to communicate the expiration of the deadline. 
+This API uses the base cancellation to communicate the expiration of the deadline.
 The information to differentiate a cancellation due to normal task cancellation is
 expanded to handle two new forms of cancellation; a cancellation due to deadline expiration,
 and a custom cancellation with a specified string for a reason. Since this is not a closed
@@ -286,6 +278,7 @@ cancelled.
 
 ```
 public struct CancellationError: Error {
+  @nonexhaustive
   public enum Reason {
     case taskCancelled
     case deadlineExpired
@@ -710,3 +703,11 @@ nesting. This approach was not taken because:
    reason about.
 3. Nothing in this proposal precludes a future task-installed deadline mechanism; the 
    explicit `withDeadline` API would remain useful even if such a mechanism were added.
+
+## Changelog
+- 1.1 Returned for revision
+  - The typed throws signature was altered to avoid an extra error type
+  - Removed the restriction around the instant requiring the duration type to be `Swift.Duration`
+  - The accessor for the current deadline was removed due to difficulty for using any InstantProtocol
+  - A new interface on CancellationError was added to handle the reasons for why a task or child task is cancelled (including a deadline exceeded reason).
+ - 1.0 Initial revision
--- a/proposals/0526-deadline.md
+++ b/proposals/0526-deadline.md
@@ -15,6 +15,12 @@ continuous clock instant representing the deadline by which the operation must
 complete. If the operation completes before the deadline, the function returns 
 the result; if the deadline expires first, the operation is cancelled.
 
+The typed throws signature was altered to avoid an extra error type, in conjunction
+with removing the restriction around the instant requiring the duration type
+to be `Swift.Duration`, the accessor for the current deadline was removed,
+and a new interface on CancellationError was added to handle the reasons
+for why a task or child task is cancelled (including a deadline exceeded reason).
+
 ## Motivation
 
 Asynchronous operations in Swift can run indefinitely, which creates several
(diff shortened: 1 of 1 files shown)


proposals/0526-deadline.md
```diff
@@ -15,6 +15,12 @@ continuous clock instant representing the deadline by which the operation must
 complete. If the operation completes before the deadline, the function returns 
 the result; if the deadline expires first, the operation is cancelled.
 
+The typed throws signature was altered to avoid an extra error type, in conjunction
```
ktoso: This isn't really the right spot for "was altered" since this proposal is _introducing_ this API. If this is to keep track of changes done in the proposal it should be at the bottom in a changelog like here https://github.com/swiftlang/swift-evolution/blob/60750ba3cc503035819e4be7990d0b64963ee619/proposals/0336-distributed-actor-isolation.md?plain=1#L1806

proposals/0526-deadline.md
```diff
@@ -608,7 +592,20 @@ composability and semantics. Duration-based timeouts accumulate drift when
...
+than intended. With an absolute deadline, every layer in the stack sees the 
+same instant and no time is lost in translation.
+
+This is the same reasoning behind Go's `context.WithDeadline` - Go provides
+both `WithTimeout` (relative) and `WithDeadline` (absolute), but recommends 
+deadlines for composable, multi-layer operations because the absolute instant 
+propagates without drift. Kotlin's `withTimeout` is duration-based, but 
+Kotlin's coroutine scope carries a single deadline internally and computes the 
+minimum against any new timeout, which is effectively what the nested 
+`withDeadline` composition in this proposal achieves explicitly.
```
ktoso, line 601: 👍

proposals/0526-deadline.md
```diff
@@ -639,12 +644,69 @@ Since the closure may itself use `withTaskCancellationHandler` or catch cancella
...
+From a nomenclature standpoint, `withDeadline` would be a term of art for Swift. By its
+nature has an implication of cooperative cancellation due to the design of Swift's 
+concurrency runtime and by that implication also interacts solely with tasks. This 
+follows suit with other languages like Kotlin - the naming in that case is withTimeout
+because the timeout in that case is an elapsed duration instead of a deadline instant.
+The name `withDeadline` also reads naturally at the call site - `try await withDeadline(...)` 
+immediately communicates to the reader that a temporal bound is in effect, which aids code 
+review and debugging. Names centered on the mechanism (`withAutomaticTaskCancellation`) 
+require the reader to infer the temporal aspect, while names centered on the concept 
+(`withDeadline`) let the reader infer the mechanism from context.
```
ktoso, line 656: 👍

proposals/0526-deadline.md
```diff
@@ -273,108 +263,91 @@ If the closure were `@Sendable`, it couldn't access actor-isolated state like
...
-/// ```
-public struct DeadlineError<OperationError: Error>: Error, CustomStringConvertible, CustomDebugStringConvertible {
-  /// The underlying cause of the deadline error.
-  public enum Cause: Sendable, CustomStringConvertible, CustomDebugStringConvertible {
-    /// The operation was cancelled due to deadline expiration and subsequently threw an error.
+```
+public struct CancellationError: Error {
+  public enum Reason {
+    case taskCancelled
     case deadlineExpired
```
ktoso, line 284: Wouldn't we want to include the instant here? Since it's not generic now we could do so without much hassle; it would help show better error messages I think?


I'm split about this, I'm also arguing below that failures should be logged at _failure sites_ and therefore we need not propagate more information about the cancellation anywhere, as we're coupling unrelated pieces of code this way... 

At the same time, I kindof am wondering if the instant is harmless enough here, as it would be hard to abuse, vs an open string or any object. 🤔

proposals/0526-deadline.md
```diff
@@ -273,108 +263,91 @@ If the closure were `@Sendable`, it couldn't access actor-isolated state like
...
-
-    /// The operation threw an error before the deadline expired.
-    case operationFailed
+    case custom(String)
```
ktoso, line 285: Oh boy, that's a really big one... Perhaps it'd be best avoided to add this "custom", as this effectively turns the communication entirely bidirectional and I'm a bit worried about coupling this can cause.

There's a few things bad about this, so I'd suggest we skip this entirely.

1. the bi-directional coupling by parsing exact messages has me worried here; the deadline number is kinda defensible, but more arbitrary information is starting to get worrying.
2. It opens up the debate if this should be `Sendable & Error`, but I think that would be equally bad:
   1. the logging of errors should be done at _failure sites_, so I think we're right to _not_ propagate more information into the cancellation. If we did, we then couple potentially unrelated pieces of code and expect some "random" cancellation handler to be able to do something with the information...

Proposal: let's not include `custom(...)` at all; but keep the `deadlineExpired` (nit I personally prefer "exceeded" but expired is fine as well if people like it). There is no inherent need for this, and it opens a lot of unconformable design questions I'm not sure we ought to be solving.

proposals/0526-deadline.md
```diff
@@ -273,108 +263,91 @@ If the closure were `@Sendable`, it couldn't access actor-isolated state like
...
-///     }
-/// }
-/// ```
-public struct DeadlineError<OperationError: Error>: Error, CustomStringConvertible, CustomDebugStringConvertible {
-  /// The underlying cause of the deadline error.
-  public enum Cause: Sendable, CustomStringConvertible, CustomDebugStringConvertible {
-    /// The operation was cancelled due to deadline expiration and subsequently threw an error.
+```
+public struct CancellationError: Error {
+  public enum Reason {
```
ktoso, line 282:
```suggestion
  @nonexhaustive
  public enum Reason {
```

How about spelling it out explicitly as nonexhaustive (https://github.com/swiftlang/swift-evolution/blob/main/proposals/0487-extensible-enums.md), even though "we know" (not everyone reading proposals knows) that's how the stdlib works?

proposals/0526-deadline.md
```diff
@@ -183,7 +178,7 @@ let (userData, prefsData) = try await (user, prefs)
 
 These absolute deadlines are composable and nestable to any set scope of a deadline. This means that when more than 
 one `withDeadline` is nested the minimum of the expiration is taken. If any nested cases are differing clocks the 
-deadline is adjusted to the minimum by aproximating the current deadline with the offset of the proposed expiration.
+deadline is adjusted to the minimum by approximating the current deadline with the offset of the proposed expiration.
```
FranzBusch: Does this still apply?

proposals/0526-deadline.md
```diff
@@ -273,108 +263,91 @@ If the closure were `@Sendable`, it couldn't access actor-isolated state like
...
 guarantees.
 
-#### Failures and expiration
-
-The mechanism this API uses to communicate the expiration or the failure of an executing deadline 
-is through a generic concrete error type: `DeadlineError`. This allows the throwing of the specific
-underlying error but also containing the applied deadline and reasoning for the failure.
+#### Cancellation
+
+This API uses the base cancellation error to communicate the expiration of the deadline.
```
FranzBusch: This doesn't use the cancellation error but task cancellation. In the end, it is up to the operation to decide what to do with that and potentially throw a `CancellationError`
```suggestion
This API uses the base cancellation to communicate the expiration of the deadline. 
```

proposals/0526-deadline.md
FranzBusch: The `withDeadline` method is still generic over a clock and instant. In my opinion, we don't need to include it. The reason I first wrapped the error was mostly because I wanted to differentiate between the two separate failures. I researched what most other languages do for deadline/timeouts and most of them don't provide any additional information besides that the deadline expired.

proposals/0526-deadline.md
FranzBusch: I argued for the custom cancellation reason because this has been a common request. We have many users that asked for being able to pass a cancellation reason when cancelling a task or a task group to aid with understanding where a cancellation comes from. It is not about where the logging is done which often already happens but that logging most often is just `Something threw an error: CancellationError()`. This allows developers to add a cancellation reason that will help them understand why a certain task got cancelled and it can be propagated up again through the `CancellationError`.

ktoso: I always am pretty allergic to those both-ways couplings, in reactive-streams we specifically avoided this back-channel and managed to avoid the same in the JDK still. But perhaps this is different enough from the streams chain case; cancellation there happens a lot, and making them heavy was a concern.

The way I see the String here is that we would not actually encourage libraries to make much use of it, but rather it'd be more like a debugging tool... 

There's another problem to discuss though here; if this implies that task eventually people will ask to gain the "task.cancellationCause" which would be really bad as we'd then have to keep those errors around in memory for a lot longer... I think we should immediately argue that we should not offer such cause "storage" in task; similarly, I'd be worried if we then start to grow APIs like `isCancelled(because: .deadlineExceeded)` which might then grow into something to "get the cause (error)" which would force us into storing the errors -- something I'd rather want to avoid.

FranzBusch: I would argue that we want to store those in the task and keep them around until the task is cleaned-up. I would like to understand your hesitance around this more. As I tried to allude to above, many Swift server adopters have asked me if we could add such a cancellation reason API including being able to retrieve the cancellation reason from the i.e. `Task.cancellationReason: CancellationError.Reason?`. This comes from a real need not only when debugging but also for deployed application. It is very common that a `CancellationError` is thrown and it is often very unclear why that happened. Most of the time it is due to the fact that something cancelled your task e.g. a deadline being exceeded or other business logic determining that a task should be cancelled. Having a cancellation reason makes it a lot easier to understand what is going on in your system both during development and in production.

I am not concerned with the performance of this for those reasons:
- Cancellation is often the error edge was is not expected to run in hot paths
- We are already storing the value or the error of a task. Storing a string in a record doesn't seem like it is going to cause a lot of overhead here.

phausler: one note: the payload for that is actually really difficult to encapsulate into an implementation. So if we do remove that it would make the implementation MUCH easier.

proposals/0526-deadline.md
phausler: sounds good; will add

proposals/0526-deadline.md
phausler: no; will change

proposals/0526-deadline.md
ktoso: > I would argue that we want to store those in the task and keep them around until the task is cleaned-up.

This is really stepping outside the scope of this proposal and doesn't feel right to introduce a half baked part of an API like this by smuggling in the cause string here, but not offering the rest of the story -- if we are to move towards heavier cancellation than just a plain signal. It might be added in the future if we decide to do so, separately.

--

I did so some more research on failure reasons; and indeed many runtimes have a string for it (seems Go was pressured into adding it as well huh, and Kotlin/JS both have reasons, though Tokio remains just a signal). So, I'm warming up to the idea but not necessarily the storage part, because as noted earlier this isn't "free" and we should discuss this separately from introducing deadlines IMHO. 

--

> We are already storing the value or the error of a task. Storing a string in a record doesn't seem like it is going to cause a lot of overhead here.

That's completely unrelated storage though. It's not like we can use the result storage for this; and it cannot be just an on-demand record either, so it would grow _all tasks_ by another record. So yeah it does have wider impact and deserves its own discussion, which doesn't belong to the deadlines introduction.

Let's remove the custom case here, and revisit separately.

FranzBusch: Leaving aside the custom storage for a second which I can get behind pulling into a separate proposal. I am wondering how we are implementing this proposal without being able to retrieve the cancellation reason of a task. In particular, let's assume this code:

```swift
try await withDeadline(in: .seconds(5)) {
  try await Task.sleep(for: .seconds(10))
}
```

`Task.sleep` throws a `CancellationError`. I would expect that with this proposal the reason on this error is `deadlineExceeded`. Without being able to retrieve the cancellation reason of your current task though, I don't see how we are able to implement this. Maybe I am missing something here.

A way to side-step all of this is to put the entire reason into a future direction which would mean that for now users aren't able to distinguish a normal cancellation from a deadline based cancellation.

proposals/0526-deadline.md
```diff
@@ -273,108 +255,92 @@ If the closure were `@Sendable`, it couldn't access actor-isolated state like
...
-```
 
-The safe current deadline accessor is trivially the following:
+extension TaskGroup {
+  public func cancelAll(reason: CancellationError.Reason)
+}
 
-```swift
-extension Task where Success == Never, Failure == Never {
-  public static var currentDeadline: any InstantProtocol? {
```
ktoso, line 361: I missed this removal? That's not okey to leave the proposal with no way of reading the deadline

PR #3290, opened 2026-05-12
Default Package Settings

mattmassicotte:
Allow SwiftPM package manifests to specific default settings for Swift/C/C++/Linker with per-target inheritance control.

### DIFF
--- a/proposals/0NNN-default-target-settings.md
+++ b/proposals/0NNN-default-target-settings.md
@@ -1,6 +1,6 @@
 # Default Target Settings
 
-* Proposal: [SE-0NNN](0NNN-default-target-sesttings.md)
+* Proposal: [SE-0NNN](0NNN-default-target-settings.md)
 * Authors: [Matt Massicotte](https://github.com/mattmassicotte)
 * Review Manager: TBD
 * Status: **Awaiting implementation**
@@ -9,7 +9,7 @@
 
 ## Introduction
 
-It is very common for packages to using the same settings flags across all their targets.
+It is very common for Swift packages to using the same settings flags across all their targets.
 A built-in mechanism to apply these base settings
 offers improved readability and convenience for package manifests.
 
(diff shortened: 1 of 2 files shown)


proposals/0NNN-default-target-settings.md
```diff
@@ -0,0 +1,301 @@
...
+## Implications on adoption
+
+This change impacts manifest authors, but should have no effects at all on package consumers.
+Authors will be able to adopt default settings freely without concern for compatibility.
+
+## Future directions
+
+The two most obvious avenues for future work are supporting `unsafeFlags` and inheritance conditions.
+
+Unsafe flags themselves already impose restrictions on package use,
```
dschaefer2: That's actually not the case any more since 6.3. Unsafe flags are really just other flags now. The restriction on package resolution has been removed.

mattmassicotte: Ohh wow I didn't realize! The SwiftPM code still treated them specially when I was working on this, but updating the proposal text should be easy. Thank you so much for letting me know.

owenv:
@swiftlang/build-and-packaging-workgroup discussed this in our meeting today. There were a few smaller bits of feedback that folks will try to share individually like Doug's comment above about unsafeFlags, but they were mostly things which might benefit from a broader review discussion and the general consensus was we support running a review of this proposal.

cc @swiftlang/ecosystem-steering-group

proposals/0NNN-default-target-settings.md
ole: I wasn't aware either that this policy had changed. This seems to be a (the?) relevant forum discussion: https://forums.swift.org/t/pitch-disable-checks-for-unsafe-flags-in-swiftpm/80698

dschaefer2: Yes, that happened before we had the working group set up and a better policy around evolution proposals. The change actually went into 6.2.

```
// unsafe flags check disabled in 6.2
usesUnsafeFlags: manifest.toolsVersion >= .v6_2 ? false : manifestTarget.usesUnsafeFlags,
```

FranzBusch:
@mikaelacaron will be the review manager for this and will reach out to you.

PR #3297, opened 2026-05-15
Amend SE-0519 to include `Sendable` and `BitwiseCopyable` conformances for `Ref` and `MutableRef`

jckarter:


DougGregor (approved):
Approved by the LSG

PR #3300, opened 2026-05-19
Propose a `Disconnected` type 

FranzBusch:


FranzBusch:
@rjmccall I opened a PR for the implementation https://github.com/swiftlang/swift/pull/89597 and addressed all the pitch feedback so far. Could you add this to the LSGs proposal review inbox?

proposals/NNNN-disconnected.md
```diff
@@ -0,0 +1,515 @@
+# Disconnected
+
+* Proposal: [SE-NNNN](NNNN-disconnected.md)
+* Authors: [Franz Busch](https://github.com/FranzBusch)
+* Review Manager: TBD
+* Status: **Awaiting implementation**
+* Implementation: [swiftlang/swift#NNNNN](https://github.com/swiftlang/swift/pull/NNNNN)
```
NotTheNHK:
```suggestion
* Implementation: https://github.com/swiftlang/swift/pull/89597
```

proposals/NNNN-disconnected.md
```diff
@@ -0,0 +1,515 @@
...
+isolation boundaries. As an example, let's look at a hypothetical `UniqueDeque`:
+
+```swift
+struct UniqueDeque<Element: ~Copyable>: ~Copyable {
+  func append(_ element: consuming Element) { ... }
+  func popFirst() -> Element? { ... }
+}
+```
+
+One use-case might want to use the `UniqueDeque` to append non-`Sendable`
```
NotTheNHK:
```suggestion
One use-case might be to use `UniqueDeque` to append non-`Sendable`
```

proposals/NNNN-disconnected.md
```diff
@@ -0,0 +1,515 @@
...
+introduced region-based isolation which leverages control flow sensitive
+diagnostics to determine whether non-`Sendable` values are safe to send across
+isolation boundaries.
+[SE-0430](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0430-transferring-parameters-and-results.md)
+introduced the `sending` parameter and result annotation to explicitly mark
+values that must be in a disconnected region at function boundaries.
+
+This proposal introduces a `Disconnected` type that preserves the disconnected
+property of a value through storage in data structures, allowing generic types
+to safely transfer non-`Sendable` values across isolation regions without
```
NotTheNHK: I would consider changing the wording "across isolation regions" to "across isolation domains".

proposals/NNNN-disconnected.md
```diff
@@ -0,0 +1,515 @@
...
+[SE-0414](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0414-region-based-isolation.md)
+introduced region-based isolation which leverages control flow sensitive
+diagnostics to determine whether non-`Sendable` values are safe to send across
+isolation boundaries.
+[SE-0430](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0430-transferring-parameters-and-results.md)
+introduced the `sending` parameter and result annotation to explicitly mark
+values that must be in a disconnected region at function boundaries.
+
+This proposal introduces a `Disconnected` type that preserves the disconnected
+property of a value through storage in data structures, allowing generic types
```
NotTheNHK: I think we should expand this section a bit more and place less emphasis on generic types. While data structures are probably the most common use case, you can also use this type to store disconnected values in actors, for example. It's somewhat awkward ergonomically, but it can be useful at times.

proposals/NNNN-disconnected.md
```diff
@@ -0,0 +1,515 @@
...
+```swift
+struct UniqueDeque<Element: ~Copyable>: ~Copyable {
+  func append(_ element: consuming Element) { ... }
+  func popFirst() -> Element? { ... }
+}
+```
+
+One use-case might want to use the `UniqueDeque` to append non-`Sendable`
+disconnected values and when popping an element send it to a different isolation
+region.
```
NotTheNHK:
```suggestion
domain.
```

proposals/0538-disconnected.md
```diff
@@ -0,0 +1,515 @@
...
+    print(element) // Error: Element is assumed to be in the same isolation region as uniqueDeque
+}
+```
+
+To make this work we would need to consume the element in `append` and
+return it from `popFirst` as `sending`; however, this would significantly limit
+this type for other important use-cases where users want to store non-`Sendable`
+but **not disconnected** elements.
+
+The fundamental limitation is that `sending` is a property of function
```
NotTheNHK, line 62: Nit: This sentence feels a bit awkward. I would suggest something like this instead:

The fundamental limitation is that `sending` is a property of function boundaries, not of types. Generic types such as `UniqueDeque` cannot conditionally control whether their stored elements should remain disconnected. Requiring `append` and `popFirst` to be `sending` would therefore only allow disconnected values to be stored in `UniqueDeque`.

proposals/NNNN-disconnected.md
```diff
@@ -0,0 +1,515 @@
+# Disconnected
+
+* Proposal: [SE-NNNN](NNNN-disconnected.md)
+* Authors: [Franz Busch](https://github.com/FranzBusch)
+* Review Manager: TBD
+* Status: **Awaiting implementation**
+* Implementation: [swiftlang/swift#NNNNN](https://github.com/swiftlang/swift/pull/NNNNN)
+* Review: ([pitch](https://forums.swift.org/...))
```
NotTheNHK:
```suggestion
* Review: ([pitch](https://forums.swift.org/t/pitch-disconnected-type-for-modeling-disconnected-values/86815))
```

proposals/0538-disconnected.md
```diff
@@ -0,0 +1,515 @@
...
+Rather than introducing a wrapper type, we could attempt to parameterize generic
+types over whether their elements are `sending`. This would require significant
+language changes to support conditional application of `sending` based on
+generic constraints, and would complicate generic type signatures. The wrapper
+type approach provides equivalent functionality with no language changes beyond
+the library addition.
+
+### Making `Disconnected` a protocol
+
+A `Disconnected` protocol could be applied to existing types. However, this
```
NotTheNHK, line 356: 👍 And simply offering a `Disconnected` protocol wouldn't solve the problem, since you still need to implement it yourself, no different than today.

NotTheNHK (commented):
I left a little bit of commentary. Overall, I think the proposal is going in the right direction. As much as I would prefer this to be handled directly by RBI, any potential solution would probably end up making things more complicated in the end.

proposals/NNNN-disconnected.md
FranzBusch: I clarified this with @hborla so quoting her here:

> "isolation region" and "isolation domain" are related but distinct concepts. Isolation domains are provided by actors and tasks. Isolation domains have associated regions but some isolation regions are not associated with any domain - that's what a "disconnected" region is.

proposals/NNNN-disconnected.md
FranzBusch: Reframed this a bit

proposals/NNNN-disconnected.md
NotTheNHK: Yes. The isolation region of a non-Sendable value can be associated with an isolation domain (actor-isolated, task-isolated). Regions without such an association are considered to be disconnected.

I found the phrase "across isolation regions" a little odd, since the non-Sendable value always remains in a disconnected region. Although, AFAIK, each sending operation assigns a fresh disconnected region (hence the need for `~Copyable` here) rather than the region itself being "transferred". While the non-Sendable value remains wrapped in a Disconnected instance, it can be moved across arbitrary isolation domains (or none at all) without the wrapped value itself becoming associated with an isolation domain.

But you're right that "across isolation domains" would be incorrect from an RBI perspective, because a transfer operation does not necessarily imply associating a disconnected region with an isolation domain (as is the case when using the `sending` keyword).

Perhaps we can find a different term for it. Although, on second thought, I think it's probably fine.

PR #3301, opened 2026-05-19
Add missing metadata fields to SE-0531

dempseyatgithub:
This PR adds missing metadata fields to SE-0531:

- Add 'Proposal' field to identify the proposal ID and link
- Add 'Review' field for pitch and review discussion threads
- Add GitHub link for review manager
- Use standard formatting for Status field
- Use recently adopted 'Summary of changes' heading instead of 'Introduction'

Without key fields like the proposal ID, the proposal will not appear on the evolution dashboard.

// @airspeedswift


airspeedswift:
Thank you!

PR #3306, opened 2026-05-20
[SE-0526] Feedback and refinements for withTaskCancellationHandler, and CancellationError.Reason

phausler:
This adds in a bit more refinements for the CancellationError.Reason and explains the future directions and constraints around that along with an overload for withTaskCancellationHandler


ktoso:
Sharing some implementation notes which don't necessarily have to be part of the proposal but this seems like a good place to post it as from this we should form the parts that do end up in the proposal update.

### Reading deadlines

We definitely want a way to read a deadline, it's a hard requirement, so let's see how we can achieve it.

There's two approaches we can take; The first one Philippe argues is more useful, however it is more complex. It has to convert all instants into the ContinousClock.Instant -- it is "easier" for users, however slower to lookup and it may answer slightly "different" times because of conversions as it needs to convert other clocks deadlines into the ContinousClock.

It is an open question how this approach would work for entirely custom clocks still a little bit... One option is to say they just say  "now" and we leave it at that, but this oculd be VERY confusing I worry...

----

**First**: Philippe has an approach where we have to iterate all deadlines and get all clocks to convert into an "best approximation" of ContiniousClock.Instant.
  - To implement this, we need to get a `visit<C: Clock>(c: C, instant: C.Instant, tolerance: ...)` call which has to visit all clocks.
  - Our implementation would call this so users can actually use methods on the clock and pass the instants to it.

To implement this visitor, we cannot do it in pure Swift (to be efficient, no horrible super hacks):

- withDeadline inserts a record {}

The visitor would have to be a protocol, since we don't have generic function literals, so something like this:


```swift
// NOT necessarily exposed as API
protocol DeadlineVisitor {
  func visit<C: Clock>(clock: C, deadline: C.Instant) -> ()
}
```

> Sidenote: This is exactly how Distributed works with its encoder/decoder protocols, and why some of it has to be in the runtime.

From runtime this is called as `void (*)(void *clock, void *deadline, SWIFT_CONTEXT void *self, Metadata *C, WitnessTable *C_conforms_to_Clock);` which informs what the withDeadline needs to store in a record:

```swift
// This "existential" box has to be hand-defined in the
// runtime because we don't have existentials like this in
// Swift. And we can take advantage of ownership a bit here,
// as if our existential actually were:
//     \exists C: Clock . (Ref<C>, Ref<C.Instant>)
struct {
  Metadata *C;
  Clock_WitnessTable *C_conforms_to_Clock;
  void *clock;  // pointer to a borrowed C
  void *deadline; // pointer to a borrowed C.Instant
  // maybe tolerance too?
}
```

So we implement this iteration for the purpose of implementing this "adjusted" API:

```swift
extension Task where Success == Never, Failure == Never {
  public static var currentEarliestContinuousDeadline: ContinuousClock.Instant? { get }
}
```

The implementation of this would need to iterate all deadlines though... and then convert all instants into the ContinuousClock.Instant and call  a new requirement on Clock:

```swift
protocol Clock ... {
  // New addition:
  func earliestContinuousInstant(from instant: Instant) -> ContinousClock.Instant
}

extension Clock {
  func earliestContinuousInstant(from instant: Instant) -> ContinousClock.Instant {
     .now // the default... also for custom clocks
  }
}
```

This is kinda true for all clocks. Any deadline may trigger at the earliest "now" but it is also kinda potentially be misleading.

This would be possible to implement.


-----


**Second** The second is simpler, but will require us to give clocks Identity. It's a miss that we didn't do this before

```
protocol Clock { // like Identifiable
  var id: ID { ... }
}
```

and thanks to this we can then offer the

```
func deadline(on: clock) -> C.Instant?
```

The benefit of this approach is:

- no adjustments, always the actual true value of the deadline for a given clock
- lookup is trivial - 
  - during entering a withDeadline() we lookup the last deadline for this clock and compare `that > newOne`; 
  - if true, we don't write a record at all; 
  - if not false, we write a record with the new deadline before it is "less in the future" than the existing one.

Lookups need to traverse records only until they hit the first record where the clock type **and** the clock identity match the lookup. This often can be a single quick lookup; it never needs to scan past a hit, like the "adjusting lookup" would need to do.


This is way cheaper to implement, but we do need to solve clock identity.

---

Thanks for collaborating on figuring this out @FranzBusch, @phausler and @rjmccall;

NotTheNHK:
@ktoso Interesting write-up. I think the second approach would be the sensible solution here. But do I understand you correctly that "give clocks Identity" would be something public-facing that each clock implementation can customize?

phausler:
I don't think we can really give clocks identity. Especially retroactively.

ktoso:
> I don't think we can really give clocks identity. Especially retroactively.

Of course we can, as we did for executors. New protocol requirement with default impl, without the `Identifiable` conformance but the `id` we can still get and make use of.

proposals/0526-deadline.md
```diff
@@ -705,6 +816,8 @@ nesting. This approach was not taken because:
    explicit `withDeadline` API would remain useful even if such a mechanism were added.
 
 ## Changelog
+- 1.2 Revisde for feedback
```
ktoso:
```suggestion
- 1.2 Revised for feedback
```

proposals/0526-deadline.md
```diff
@@ -342,6 +358,93 @@ This error is propagated from whenever the task (or child task) is cancelled via
 The reason specified will then be available to the `CancellationError` and can be retrieved from the `reason`
 property on the cancellation error.
 
+#### Accessing active deadlines
+
+External systems may need to interoperate with active deadlines. This means that the
+applied deadline needs to be retrievable, however this particularly becomes
+tricky since the clock is generic for the deadlines. To that end
+the accessor for the active deadline is generic for a clock type
```
ktoso:
```suggestion
the accessor for the active deadline accepts a generic clock instance:
```

proposals/0526-deadline.md
```diff
@@ -259,9 +259,9 @@ guarantees.
 
 This API uses the base cancellation to communicate the expiration of the deadline.
 The information to differentiate a cancellation due to normal task cancellation is
-expanded to handle two new forms of cancellation; a cancellation due to deadline expiration,
-and a custom cancellation with a specified string for a reason. Since this is not a closed
-set of possible reasons for future development, this reason is left as an open enumeration.
+expanded to handle two new forms of cancellation; a cancellation due to normal task cancellation
```
ktoso: Since it's not two new reasons, but one new reason really

```suggestion
expanded to handle a new reason for cancellation: 
```

proposals/0526-deadline.md
```diff
@@ -259,9 +259,9 @@ guarantees.
...
+and a cancellation due to deadline expiration. Since this is not a closed set of possible reasons
```
ktoso:
```suggestion
due to deadline expiration. Since this is not a closed set of possible reasons
```

proposals/0526-deadline.md
```diff
@@ -301,6 +299,11 @@ statements require an `@unknown default` case. Since previous cancellation was s
 has been already written the developer already has handled the cases of cancellation without a 
 given reason; this will continue to be the case.
 
+Note: For now the `Reason` type is restricted to known simple enumeration values without any
```
ktoso: Needs the > to render correctly.

```suggestion
> Note: The `Reason` type is restricted to known simple enumeration values without any
```

Not implying anything future looking

proposals/0526-deadline.md
```diff
@@ -342,6 +358,93 @@ This error is propagated from whenever the task (or child task) is cancelled via
...
+External systems may need to interoperate with active deadlines. This means that the
+applied deadline needs to be retrievable, however this particularly becomes
+tricky since the clock is generic for the deadlines. To that end
+the accessor for the active deadline is generic for a clock type
+
+```swift
+extension Task where Success == Never, Failure == Never {
+  public static var hasActiveDeadline: Bool { get }
+
+  public static func activeDeadline<C: Clock & Identifiable>(for clock: C) -> C.Instant?
```
ktoso, line 389: Signature looks good, though we should have the same defaulted parameter here as we do on the write side

proposals/0526-deadline.md
```diff
@@ -342,6 +358,93 @@ This error is propagated from whenever the task (or child task) is cancelled via
...
+}
+```
+
+If any deadline is active then the static property `hasActiveDeadline` returns true.  
+This only applies to the current task if the execution of that task is within a call
```
ktoso: A little confusing wording about current task only? what if were a child task where the parent had a deadline set and therefore we do too

proposals/0526-deadline.md
```diff
@@ -342,6 +358,93 @@ This error is propagated from whenever the task (or child task) is cancelled via
...
+  }
+  if let deadline = Task.activeDeadline(for: SuspendingClock()) {
+    // use the deadline as a SuspendingClock.Instant
+  }
+}
+```
+
+When the call to `activeDeadline(for:)` is made, the query looks up the most narrow
+application of any specified deadline with that clock, if the current nesting of 
+`withDeadline` calls does not used the specified clock type then the next nesting
```
ktoso:
```suggestion
`withDeadline` calls does not use the specified clock type then the next nesting
```

proposals/0526-deadline.md
```diff
@@ -342,6 +358,93 @@ This error is propagated from whenever the task (or child task) is cancelled via
...
+If the nesting of `withDeadline` is stacked with a `ContinuousClock` deadline of 
+"in two seconds" and then a `SuspendingClock` of "in three seconds" and a new
+nesting is made of a `ContinuousClock` is made for "in 10 seconds" the last
+10 seconds is known to be less narrow than the outer 2 seconds continuous clock 
+deadline. This means that within the scope of the "in 10 seconds" deadline the query
+for the `activeDeadline(for: ContinuousClock.self)` would return the deadline of within 
+2 seconds and the `activeDeadline(for: SuspendingClock.self)` would return the deadline
+of within 3 seconds. Since clock instants cannot be compared without potentially
+arbitrarily lossy conversions it means that the query for the current applied deadline
+is only accurate to the specific clock type given.
```
ktoso, line 427: Good example but this should be a code snippet and not just text for clarity

proposals/0526-deadline.md
```diff
@@ -342,6 +358,93 @@ This error is propagated from whenever the task (or child task) is cancelled via
...
+arbitrarily lossy conversions it means that the query for the current applied deadline
+is only accurate to the specific clock type given. 
+
+#### Identification of Clocks for coalescing 
+
+The expected behavior when setting a deadline is that any active deadline, given a specific clock,
+will always apply with the most narrow deadline available. Specifically if a deadline is 
+active for an expiration of `in 10 seconds` and a new deadline is applied for `in 5 seconds` 
+relative both relative to the continuous clock, then the applied deadline within the new
+scope is the `in 5 seconds`. Likewise if the reverse was applied; where it is already at `in 5 seconds`
```
ktoso: “in 5 seconds” isn’t code, so it’s unusual to use a code styling to it, use italics instead? To not confuse people like this is some code syntax they don’t know about

proposals/0526-deadline.md
```diff
@@ -342,6 +358,93 @@ This error is propagated from whenever the task (or child task) is cancelled via
...
+and a new scope is applied to `in 10 seconds` both on the continuous clock, then the internal
+logic will effectively skip the `in 10 seconds` since that deadline is known to beyond the current
+active deadline. This must have some way of determining if a given clock passed in to 
+the `withDeadline` functions is that same specific clock. To that end, the clocks are 
+required to be identified. The two major clocks; `ContinuousClock` and `SuspendingClock`
```
ktoso:
```suggestion
required to be identifiable. The two major clocks; `ContinuousClock` and `SuspendingClock`
```

proposals/0526-deadline.md
```diff
@@ -342,6 +358,93 @@ This error is propagated from whenever the task (or child task) is cancelled via
...
+both will gain a new conformance to `Identifiable` and each of which will have a new ID
+type of `SystemClockID`. As a side effect this means that new APIs can be written as:
```
ktoso: It’s not a side effect, it just means that

proposals/0526-deadline.md
```diff
@@ -342,6 +358,93 @@ This error is propagated from whenever the task (or child task) is cancelled via
...
+active deadline. This must have some way of determining if a given clock passed in to 
+the `withDeadline` functions is that same specific clock. To that end, the clocks are 
+required to be identified. The two major clocks; `ContinuousClock` and `SuspendingClock`
+both will gain a new conformance to `Identifiable` and each of which will have a new ID
+type of `SystemClockID`. As a side effect this means that new APIs can be written as: 
+`where C: Clock & Identifiable, C.ID == SystemClockID`. That particular refinement not only
+allows for the direct identification of the specific clock but also a constraint to the 
+standard system clocks.
+
+Note: since the system clock may grow additional identifiers it is left as non-exhaustive.
```
ktoso:
```suggestion
> Note: Since the system clock may grow additional identifiers it is left as non-exhaustive.
```

proposals/0526-deadline.md
ktoso: Im not sure why we’re writing this at all? I don’t think we need to write “constraint can be used” whenever we add some type somewhere. This might be confused with some of our APIs actually doing this?

proposals/0526-deadline.md
```diff
@@ -342,6 +358,93 @@ This error is propagated from whenever the task (or child task) is cancelled via
...
+
+extension ContinuousClock: Identifiable {
+  public var id: SystemClockID { .continuous }
+}
+
+extension SuspendingClock: Identifiable {
+  public var id: SystemClockID { .suspending }
+}
+```
+
```
ktoso, line 479: We should document that in theory this is to support instance clocks as well, if users choose to write such

proposals/0526-deadline.md
phausler: Im not sure that is really absolutely needed; since the query side should likely consider all available clocks (or at the very least all system clocks) a default value doesnt really add much, where as the side that is setting a deadline is more so just interested in a time frame; and that is going to be the lesser amount of needed knowledge. So I would actually say that not having a default value is more a feature than a shortcoming because it forces folks on the reading side to consider what clocks to support.

proposals/0526-deadline.md
phausler: fair; will remove

proposals/0526-deadline.md
phausler: ill change to italics, that should read better.

proposals/0526-deadline.md
ktoso: Hmm, we can leave this as a topic for SE review then, let's see what folks think

proposals/0526-deadline.md
```diff
@@ -279,14 +279,14 @@ normal task cancellation.
 public struct CancellationError: Error {
   @nonexhaustive
   public enum Reason {
-    case taskCancelled
+    case userRequested
```
ktoso, line 282: I don't think that's right; it's not user requested per se, throws out of groups and async lets also cause this;

We can discuss this in SE review though.

proposals/0526-deadline.md
ktoso: This isn't addressed yet? // edit: now it is

proposals/0526-deadline.md
```diff
@@ -330,6 +333,19 @@ extension ThrowingDiscardingTaskGroup {
...
+```swift
+public nonisolated(nonsending) func withTaskCancellationHandler<Return, Failure>(
+  operation: nonisolated(nonsending) () async throws(Failure) -> Return,
+  onCancel handler: sending (CancellationError.Reason) -> Void
+) async throws(Failure) -> Return 
+```
+
+This function works exactly as the existing `withTaskCancellationHandler` does today, 
+except that the `onCancel` handler is passed the reason for cancellation.
+
```
FranzBusch, line 348:
```suggestion

Furthermore, the cancellation reason can be retrieved from the current Task similar to how cancellation is checked.

```swift
extension Task where Success == Never, Failure == Never {
  /// Returns the current task's cancellation reason if the task was cancelled.
  ///
  /// This property will be set if `Task.isCancelled` returns `true`.
  ///
  /// - Important: The cancellation reason can only change once during the lifetime of a task from `nil` to a specific reason. After this initial change the reason will not change.
  public static var cancellationReason: CancellationError.Reason? { get }
}
```
```

ktoso: Missing the `UnsafeCurrentTask` version;

Seems we raced doing the update; I prepared a PR just now over here updating the same, including the missing bits 😅 

https://github.com/phausler/swift-evolution/pull/1

ktoso:
Not sure how you applied the change @phausler but it didn't work; the code snippet is not present. Just the sentence "Furthermore, the cancellation reason can be retrieved from the current Task similar to how cancellation is checked." without showing the API is not sufficient really.

Easiest would be to take https://github.com/phausler/swift-evolution/pull/1

ktoso:
Thanks that works too :)

ktoso (approved):

PR #3312, opened 2026-05-29
[ST-0027] Macro for getting the current source location

grynspan:
See [here](https://github.com/swiftlang/swift-evolution/blob/jgrynspan/sourcelocation-macro/proposals/testing/0027-sourcelocation-macro.md).

### DIFF
--- a/proposals/testing/0027-sourcelocation-macro.md
+++ b/proposals/testing/0027-sourcelocation-macro.md
@@ -1,9 +1,9 @@
 # Macro for getting the current source location
 
-* Proposal: [ST-NNNN](NNNN-sourcelocation-macro.md)
+* Proposal: [ST-0027](0027-sourcelocation-macro.md)
 * Authors: [Jonathan Grynspan](https://github.com/grynspan)
-* Review Manager: TBD
-* Status: **Awaiting review**
+* Review Manager: [Stuart Montgomery](https://github.com/stmontgomery)
+* Status: **Active Review (August 3...August 13, 2026)**
 * Bug: rdar://178259171
 * Implementation: [swiftlang/swift-testing#1733](https://github.com/swiftlang/swift-testing/pull/1733)
 * Review: ([pitch](https://forums.swift.org/t/pitch-swift-testing-macro-for-getting-the-current-source-location/87025))
--- a/proposals/testing/NNNN-sourcelocation-macro.md
+++ b/proposals/testing/NNNN-sourcelocation-macro.md
@@ -6,7 +6,7 @@
 * Status: **Awaiting review**
 * Bug: rdar://178259171
 * Implementation: [swiftlang/swift-testing#1733](https://github.com/swiftlang/swift-testing/pull/1733)
-* Review: ([pitch](https://forums.swift.org/...))
+* Review: ([pitch](https://forums.swift.org/t/pitch-swift-testing-macro-for-getting-the-current-source-location/87025))
 
 ## Introduction
 
--- a/proposals/testing/NNNN-sourcelocation-macro.md
+++ b/proposals/testing/NNNN-sourcelocation-macro.md
@@ -1,0 +1,174 @@
+# Macro for getting the current source location
+
+* Proposal: [ST-NNNN](NNNN-sourcelocation-macro.md)
+* Authors: [Jonathan Grynspan](https://github.com/grynspan)
+* Review Manager: TBD
+* Status: **Awaiting review**
+* Bug: rdar://178259171
+* Implementation: [swiftlang/swift-testing#1733](https://github.com/swiftlang/swift-testing/pull/1733)
+* Review: ([pitch](https://forums.swift.org/...))
+
+## Introduction
+
+Swift Testing includes a type, [`SourceLocation`](https://developer.apple.com/documentation/testing/sourcelocation),
+that represents the precise location of something in a file (typically a .swift
+file). Various Swift Testing API takes an instance of this type in order to
+correctly attribute diagnostics and test issues that occur at test time. This
+proposal covers introducing a macro that can be used as a default argument to
+such functions.
+
+## Motivation
+
+The Swift standard library includes macros to get the current file ID, file
+path, line, and column at compile time. These macros can then be used as default
+function arguments to allow automagical capture of the caller's location in
+source. For example, [`fatalError()`](https://developer.apple.com/documentation/swift/fatalerror(_:file:line:))
+takes the file and line number and prints them to `stderr` when called.
+
+Swift Testing needs to capture all four of these values, which is quite verbose
+and somewhat tedious to work with, so various Swift Testing APIs encapsulate all
+of them in a single argument of type `SourceLocation`. Swift Testing provides a
+`#_sourceLocation` macro that expands, at compile time, to an appropriate
+expression:
+
+```swift
+public func withKnownIssue(
+  _ comment: Comment? = nil,
+  isIntermittent: Bool = false,
+  sourceLocation: SourceLocation = #_sourceLocation,
+  _ body: () throws -> Void
+)
+```
+
+This macro, being underscored, is not formally supported, nor does it appear in
+Swift Testing's documentation. It is also not sufficient to use something like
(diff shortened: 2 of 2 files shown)


proposals/testing/NNNN-sourcelocation-macro.md
```diff
@@ -0,0 +1,174 @@
...
+## Alternatives considered
+
+- **Formally supporting the existing `#_sourceLocation` macro.** This symbol is
+  underscored and does not appear in documentation, and the use of underscored
+  symbols is normally a "tell" for developers that they're using something in
+  Swift that isn't guaranteed to exist in future Swift releases.
+
+- **Naming the macro something different.** We considered alternatives such as
+  `#here` and `#currentSourceLocation`, but `#sourceLocation` seems the most
+  appropriate name for it.
```
stmontgomery: In this alternative, I'd find it helpful to reference back to the earlier section of the proposal where you mentioned the need to fully-qualify the macro with a module selector, and say that is the primary reason why an alternate name might be desirable. Just to clarify that this is more than run-of-the-mill bike shedding, there's a usability consideration. (Not that I think we _should_ use a different name, I just think the bullet could provide more context.)

proposals/testing/NNNN-sourcelocation-macro.md
```diff
@@ -0,0 +1,174 @@
...
+  isIntermittent: Bool = false,
+  sourceLocation: SourceLocation = #_sourceLocation,
+  _ body: () throws -> Void
+)
+```
+
+This macro, being underscored, is not formally supported, nor does it appear in
+Swift Testing's documentation. It is also not sufficient to use something like
+[`SourceLocation.init()`](https://developer.apple.com/documentation/testing/sourcelocation/init(fileid:filepath:line:column:))
+as it will capture the _wrong_ source location. Thus, test authors have no
```
stmontgomery: You could mention [SE-0422](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0422-caller-side-default-argument-macro-expression.md) here as the underlying reason why a macro makes this possible

proposals/testing/0027-sourcelocation-macro.md
```diff
@@ -0,0 +1,174 @@
...
+  _ comment: Comment? = nil,
+  isIntermittent: Bool = false,
+  sourceLocation: SourceLocation = #_sourceLocation,
+  _ body: () throws -> Void
+)
+```
+
+This macro, being underscored, is not formally supported, nor does it appear in
+Swift Testing's documentation. It is also not sufficient to use something like
+[`SourceLocation.init()`](https://developer.apple.com/documentation/testing/sourcelocation/init(fileid:filepath:line:column:))
```
jerryjrchen, line 45: It's not obvious to me why this init would have the wrong source location. It seems like it would be incumbent on the caller to pass in the correct arguments (as mentioned below) as the initialiser doesn't have any default values right?

proposals/testing/NNNN-sourcelocation-macro.md
```diff
@@ -0,0 +1,174 @@
...
+/// }
+/// ```
+@freestanding(expression) public macro sourceLocation() -> SourceLocation
+```
+
+Note that, as indicated in the documentation for this macro, you must specify
+the module name when using this macro to avoid conflicting with the [`#sourceLocation(file:line:)`](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/statements/#Line-Control-Statement)
+statement built into the Swift language.
+
+The existing `#_sourceLocation` macro will be marked deprecated and renamed, but
```
stmontgomery: > …will be marked deprecated **and renamed**, …

The existing one will be renamed? (To what?)

stmontgomery: Oh, `renamed: "Testing::sourceLocation"` in `@available`. That's fair, but I found the phrasing a bit confusing since it sounded like you planned to rename the old macro. Maybe avoid that by just saying it will be deprecated?

proposals/testing/0027-sourcelocation-macro.md
```diff
@@ -0,0 +1,174 @@
...
+/// func cookBurger(sourceLocation: SourceLocation = #Testing::sourceLocation) {
+///   // ...
+/// }
+/// ```
+@freestanding(expression) public macro sourceLocation() -> SourceLocation
+```
+
+Note that, as indicated in the documentation for this macro, you must specify
+the module name when using this macro to avoid conflicting with the [`#sourceLocation(file:line:)`](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/statements/#Line-Control-Statement)
+statement built into the Swift language.
```
stmontgomery, line 97: This may be orthogonal to the proposal, and this would be a language/compiler change I realize. But I wonder if, in principle at least, it would be possible for the language to interpret bare `#sourceLocation` _in expression position_ as a third-party macro, and thus not require module selector (`::`) qualification, but continue to interpret other usages, such as those in statement or code-item position, the current way.

proposals/testing/0027-sourcelocation-macro.md
```diff
@@ -0,0 +1,174 @@
...
+ /// [`#line`](https://developer.apple.com/documentation/swift/line()), and
+ /// [`#column`](https://developer.apple.com/documentation/swift/column()) as a
+ /// default argument to a function.
+ ///
+ /// ```swift
+ /// func cookBurger(sourceLocation: SourceLocation = #_sourceLocation) {
+ ///   // ...
+ /// }
+ /// ```
++@available(swift, deprecated: 100000.0, renamed: "Testing::sourceLocation")
```
stmontgomery, line 126: I think we ought to deprecate it "for real" and not just use the `10000.0` "to-be-deprecated" marker. (That's what this is, right? Hope I didn't get that wrong.)

Do you think there's a compelling reason to not do that?

proposals/testing/NNNN-sourcelocation-macro.md
```diff
@@ -0,0 +1,174 @@
...
+## Integration with supporting tools
+
+No additional integration with tools is required.
+
+## Future directions
+
+- In the future, we likely want to adjust the Swift compiler to distinguish the
+  use of `#sourceLocation` in expression position from its use in statement
+  position, and to only use the compiler statement if `#sourceLocation`
+  unambiguously refers to it rather than to a macro.
```
stmontgomery: Oh lol. I swear I wrote [this comment](https://github.com/swiftlang/swift-evolution/pull/3312/changes#r3336088975) above before I noticed this paragraph which basically says the same thing 🙃 

Anyway yeah, sounds like we're on the same page

proposals/testing/0027-sourcelocation-macro.md
stmontgomery: Never mind, you covered it [here](https://github.com/swiftlang/swift-evolution/pull/3312/changes#r3336116160) (which I just found)

proposals/testing/0027-sourcelocation-macro.md
stmontgomery: See [here](https://github.com/swiftlang/swift-evolution/pull/3312#discussion_r3336044730) for the proposal which explains it. And I agree it's non-obvious to readers who aren't aware of this macro nuance. I'd suggest mentioning that proposal

proposals/testing/NNNN-sourcelocation-macro.md
grynspan: Will rephrase.

proposals/testing/0027-sourcelocation-macro.md
grynspan: We generally don't do formal deprecation in the same release where we introduce a new replacement API as it causes undue noise/churn for developers. Which is why we have `API_TO_BE_DEPRECATED` in the first place.

stmontgomery: Users wanting to maintain support for _building_ using an older toolchain is a reasonable justification, thanks. 

How about we clarify in the proposal that it will be deprecated using the special "to-be-deprecated" marker initially, but that later (after some settling period TBD), the intention is to lower the deprecated-in version down to a specific Swift version retroactively. That way, the proposal states the intent which is to deprecate it fully—it's just that we won't do so in the immediate term.

stmontgomery (approved):

PR #3315, opened 2026-06-03
Update 0462-task-priority-escalation-apis.md

mattmassicotte:
The Mutex variable is actually `m` in this snippet.

The sentence immediately following does not read right to me at all. I think this change matches the authors intent closely while also feeling more grammatical.

### DIFF
--- a/proposals/0462-task-priority-escalation-apis.md
+++ b/proposals/0462-task-priority-escalation-apis.md
@@ -96,7 +96,7 @@ await withTaskPriorityEscalationHandler {
         Task.escalatePriority(of: task, to: newPriority)
     }
   } onPriorityEscalated: { oldPriority, newPriority in
-    state.withLock { state in
+    m.withLock { state in
       switch state {
       case .initialized, .priority:
         // priority was escalated just before we managed to store the task in the mutex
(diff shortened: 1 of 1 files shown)


NotTheNHK:
I suggest we rename `m` to `state` instead. And while we're at it already, why not rename all the single-letter variables/abbreviations to more descriptive names?
I noticed there are quite a few of these present.

proposals/0462-task-priority-escalation-apis.md
```diff
@@ -109,7 +109,7 @@ await withTaskPriorityEscalationHandler {
 }
 ```
 
-The above snippet handles edge various ordering situations, including the task escalation happening after
+The above snippet handles various edge ordering situations, including the task escalation happening after
```
ktoso: Or just remove the word edge. Pretty sure this was supposed to say edge cases but then changed half of it

mattmassicotte: ahhh well that makes more sense to me! But you're right it works well just removing it too.

mattmassicotte:
I'm not opposed to more descriptive names at all! I'm just not sure I want to personally take that on for this particular change.

NotTheNHK:
We can do this in a follow-up PR. I think it would be worth it.

rjmccall (approved):

PR #3322, opened 2026-06-07
[SE-0488] adjustments to scope

glessard:
Since SE-0488 has been accepted, it's become clear that the proposed extensions to `Slice` are generally not needed. Furthermore, the proposal had the wrong return type for the `Slice` extensions. Given that, this edit moves the `Slice` extensions to an "alternatives considered" item, with an explanation of the change.


rjmccall:
I've put this on the agenda to talk about in the LSG.

allevato:
Thanks for the update! I'll post this on the review announcement thread.

PR #3324, opened 2026-06-08
Update toolchain link with Ubuntu

natecook1000:
Adds a link to the Linux toolchain – still looking for where the Windows toolchain ended up…


rjmccall:
We don't generally link toolchains from the proposal document at all; they're just linked from the review thread.

PR #3347, opened 2026-07-06
[SE-0525] amend to add conformances for SIMD types

glessard:
We added conformances to `ConvertibleToBytes` and `ConvertibleFromBytes` for concrete SIMD types in https://github.com/swiftlang/swift/pull/90352. This amendment adds these types to the proposal's conformance list.

SE-0525 is now marked as implemented in Swift 6.4.


xwu:
The language steering group agreed that this addition is fine to accept summarily. Will post an updated announcement shortly.

PR #3349, opened 2026-07-07
[proposal] lazy initializers for accessor macros

nilsgrabenhorst:
A proposal to enable `self` access for property initializers if re-contextualized by a macro. The compiler currently type-checks these initializer expressions in their original context before macro expansion, where `self` is not available. The proposed implementation removes this limitation.

[Pitch](https://forums.swift.org/t/pitch-lazy-accessor-macros/87515)

[Rendered proposal](https://github.com/nilsgrabenhorst/swift-evolution/blob/lazy-accessor-macros/proposals/NNNN-self-access-for-property-initializers.md)

This is my first contribution. The implementation still needs a bit of cleanup and fixing of tests. I will finish it as soon as possible, but I wanted to make the proposal available first to get some feedback. I hope this is OK!


proposals/NNNN-lazy-accessor-macros.md
```diff
@@ -0,0 +1,399 @@
...
+* Proposal: [SE-NNNN](NNNN-lazy-accessor-macros.md)
+* Authors: [Nils Grabenhorst](https://github.com/nilsgrabenhorst)
+* Review Manager: TBD
+* Status: **Awaiting review**
+* Implementation: [swiftlang/swift#89757](https://github.com/swiftlang/swift/pull/89757)
+* Review: ([pitch](https://forums.swift.org/t/pitch-lazy-accessor-macros/87515))
+
+## Summary of changes
+
+Adds an optional `initialization: lazy` parameter for `accessor` macro role declarations. It enables macro authors to declare that property initializers will be evaluated lazily after macro expansion. The compiler will type-check the initializer expression accordingly, allowing `self` access for `lazy` property initializers.
```
allevato: Based on the discussion in the pitch thread, here and throughout the rest of this document, it probably makes sense to lean away from describing the initialization as "lazy" but to focus on the actual underlying change that's being made. I would even hesitate to use words like "deferred" because that doesn't accurately describe what's happening here. We aren't lazily type-checking the initializer expression or deferring its evaluation. Rather, type-checking still happens before the macro is expanded but you're proposing specifically to _change whether instance members are allowed to be referenced by the initializer expression._

And likewise, I would recommend updating the title of the proposal to focus on this aspect.

I think it still makes sense to use `lazy` in the Motivation section because that truly is a language feature that can't be duplicated correctly by a macro without the proposed feature, and concrete examples are great. But front-loading so much of the discussion with `lazy` might distract readers and lead them to think the proposal is doing something that it isn't (like replacing `lazy` with a macro).

proposals/NNNN-lazy-accessor-macros.md
```diff
@@ -0,0 +1,399 @@
...
+    let mice = 21
+    
+    @Lazy
+    var theAnswer = mice * 2
+    //              ^ ❌ cannot use instance member 'mice' within property initializer; property initializers run before 'self' is available
+}
+```
+The error is unnecessarily limiting, because `@Lazy` moves the initializer into a `get` accessor, where `self` access is allowed. 
+
+### Concrete Example
```
allevato: With the `@Lazy` example already present, I wonder if this second example is actually needed. It's complex enough that it means someone reading the proposal linearly won't get to the meat-and-potatoes of the change being proposed until halfway into the document. Could it be shortened?

proposals/NNNN-lazy-accessor-macros.md
```diff
@@ -0,0 +1,399 @@
...
+            }
+    }
+}
+```
+
+Clearly, this approach is awkward to work with. It exposes the risk of bugs if the dependency is not injected correctly. The cleanest and most convenient solution would be to inject the viewModel in the property initializer, allowing `self` access with lazy evaluation.
+
+## Proposed solution
+
+Add an optional `initializion:` parameter to the accessor macro role declaration. Possible values are `lazy` or `eager`. A macro author uses `lazy` to declare that the initializer is invoked lazily after macro expansion, which enables access to `self` in the initializer expression. `eager` is the default value, and resembles the current behavior of accessor macros.
```
allevato: If you're considering the naming advice given in the pitch thread, feel free to update this and the remaining examples/discussion to use new naming.

But ultimately the naming here is your choice as the author and it can be debated in the review thread.

proposals/NNNN-lazy-accessor-macros.md
```diff
@@ -0,0 +1,399 @@
...
+
+## Source compatibility
+
+Additive. Existing macro role declarations may remain unchanged. In this case, today's behavior remains unaffected.
+
+Newly written accessor macros may elect to specify `lazy` initializer usage. This does not affect existing source code.
+
+A macro author may elect to specify `lazy` initializer usage for an existing macro. This change may affect existing code:
+
+```swift
```
allevato: There was some discussion in the pitch thread about whether we could just change the way we typecheck property initializer expressions _in general_ to always allow instance member access but diagnose the invalid ones later. You pointed out that this example with same-named instance and static properties would be problematic because it would change the meaning. Discussing this in Alternatives Considered would be helpful, to highlight why this isn't something we can just do by default, and why we need an explicit knob for it.

In fact, as I write this, I think this part of the discussion is better served as an Alternatives Considered rather than Source Compatibility. Source compatibility is more about, if we make this change, does it change the behavior of any _existing code that is otherwise unchanged_? That's not the case here: a macro author changing the implementation of their macro isn't a source compatibility issue that _this proposal_ needs to care about; that's solely the responsibility of the macro author.

allevato (commented):
I'll be review managing this; I've left some editorial comments. Please take a look!

proposals/NNNN-lazy-accessor-macros.md
ole: > And likewise, I would recommend updating the title of the proposal to focus on this aspect.

If you do end up changing the title, I'd like to ask you to also rename the file to reflect the new title. We don't want to rename proposal files retroactively to avoid broken links, so we should take care to use a good filename while we still can (before the proposal gets assigned a number).

nilsgrabenhorst: Hi Tony,
thanks a lot for your review.

I will rework the proposal by focusing on `selfAvailable`. This should make the idea more clear.

> [...]I would even hesitate to use words like "deferred" because that doesn't accurately describe what's happening here. We aren't lazily type-checking the initializer expression or deferring its evaluation. Rather, type-checking still happens before the macro is expanded but you're proposing specifically to _change whether instance members are allowed to be referenced by the initializer expression._

Oh, I see where the confusion came from. I never intended `lazy` to refer to the type-checking. Instead, it was meant as a way for the author to declare that the init expression is moved to a new context, where it is called later when `self` is available.

nilsgrabenhorst: Done, using `selfAvailable` throughout the proposal.
Also updated the title and file name, plus updated the link in my initial comment to match.

proposals/NNNN-lazy-accessor-macros.md
nilsgrabenhorst: Thanks for the suggestion—done!

I was a bit concerned myself.
I still believe the second example is a compelling motivation, but it truly is looooong...
I just removed it and added a link to the example in the pitch to allow anyone interested to explore it without wasting anybody else's time.

proposals/NNNN-lazy-accessor-macros.md
nilsgrabenhorst: Done.
I updated the naming throughout the proposal.

proposals/NNNN-lazy-accessor-macros.md
nilsgrabenhorst: Good point! Done.

proposals/NNNN-self-access-for-property-initializers.md
```diff
@@ -0,0 +1,306 @@
...
+
+Since the initial initializer type-check happens before macro expansion, there is currently no information about how the initializer will be used by the macro. For example, it could be used in an `init` accessor. The initializer would be evaluated during initialization of the enclosing type, when there is no access to `self`. For such a situation, the current implementation is correct: `self` is not available during type-checking in the original context.
+
+Other macros such as `@Lazy` may result in an expansion where `self` access would be valid for the re-contextualized initializer. Currently, this fact is unknown during the initial check; therefore, `self` access is assumed to be illegal.
+
+### Role Declaration
+
+This proposal adds an optional `initialization:` parameter to the macro role declaration of accessor macros. It can have one of two possible arguments: either `selfAvailable` or `selfUnvailable`.
+
+- `selfUnvailable`: The current behavior – no `self` access for the property initializer
```
allevato: I see a handful of places where you have `selfUnvailable` instead of `selfUnavailable` (missing "a"). Can you fix those up?

proposals/NNNN-self-access-for-property-initializers.md
```diff
@@ -0,0 +1,306 @@
...
+    //              ^ 🛑 cannot use instance member 'mice' within property initializer; property initializers run before 'self' is available
+}
+```
+The error is unnecessarily limiting, because `@Lazy` moves the initializer into a `get` accessor, where `self` access is allowed. 
+
+There is another [example](https://forums.swift.org/t/pitch-lazy-accessor-macros/87515#p-399664-concrete-example-2) in the original pitch, which shows how the proposed change could enable access to SwiftUI environment values when initializing `Observable` objects.
+
+## Proposed solution
+
+Add an optional `initializion:` parameter to the accessor macro role declaration. Valid arguments are `selfAvailable` or `selfUnavailable`. A macro author uses `selfAvailable` to declare that the new context of the initializer after macro expansion enables access to `self`. `selfUnavailable` is the default value, and resembles the current behavior of accessor macros.
```
allevato: sp: `initialization` (not `initializion`)

proposals/NNNN-self-access-for-property-initializers.md
nilsgrabenhorst: Indeed, thanks!

Should be fixed.

proposals/NNNN-self-access-for-property-initializers.md
nilsgrabenhorst: Fixed. Thanks!

allevato (approved):

PR #3352, opened 2026-07-09
[SE-0508] Update source compatibility to note `init` accessor breakage

hamishknight:
There is an additional source compatibility impact with `init` accessors (https://github.com/swiftlang/swift/issues/90280), note it in the proposal text for posterity.


hborla (approved):
The LSG discussed this and we agree this is a niche enough case that the source break is acceptable.

One minor bit of editorial feedback: it's left to the reader to connect the dots that this only affects `init` accessors because you cannot write an initial value on a computed property that only has `get`/`set` accessors. To have an initial value on a computed property, you must have an `init` accessor. I think it's worth stating that explicitly in the section.

PR #3359, opened 2026-07-15
Introduce SV-NNNN numbering convention for vision documents

stmontgomery:
This introduces 4-digit ID prefixes of the form `SV-NNNN` for [vision documents](https://github.com/swiftlang/swift-evolution/blob/main/process.md#proposals-roadmaps-and-visions), to retroactively number them matching the convention used by proposals (in the [`proposals/`](https://github.com/swiftlang/swift-evolution/tree/main/proposals) directory).

Since there are likely many existing links pointing to the unversioned filenames, the original files are left in place; each new versioned file just carries the title, its new ID, and a note redirecting to the original.

I chose the ID numbers based on the date each vision document was first added to the repo:

| ID | Filename | Creation Date |
|------|-----------------------------|---------------|
| 0001 | using-swift-from-c++.md | Dec 16, 2022 |
| 0002 | macros.md | Mar 22, 2023 |
| 0003 | using-c++-from-swift.md | May 22, 2023 |
| 0004 | embedded-swift.md | Oct 25, 2023 |
| 0005 | webassembly.md | Mar 9, 2024 |
| 0006 | swift-testing.md | May 22, 2024 |
| 0007 | memory-safety.md | Oct 1, 2024 |
| 0008 | approachable-concurrency.md | Feb 18, 2025 |
| 0009 | networking.md | Mar 6, 2026 |

I also added a description of the new document formatting details to the [process documentation](https://github.com/swiftlang/swift-evolution/blob/main/process.md).

> Thank you to @ktoso for suggesting this!

### DIFF
--- a/process.md
+++ b/process.md
@@ -520,6 +520,35 @@ Review Manager
 
 </details>
 
+## Vision documents
+
+Like proposals, vision documents should generally not be renamed once they've been added to the repository.
+It's okay to change the title in the text, but the filename should remain stable.
+
+Vision documents generally have fewer formatting guidelines to follow compared to proposal documents.
+However, they should start with the following basic structure and use standard fields:
+
+```text
+# Name of vision
+
+* Field1: value
+* Field2: value
+
+The remainder of the document has no specific requirements.
+```
+
+Standard fields include:
+* `Vision`:
+  exactly `[prefix-number](filename.md)`, where `prefix` is the appropriate document prefix (generally `SV` for all visions),
+  `number` is the sequence number of this vision within that prefix, and `filename.md` is the filename of this vision document.
+  Required in all vision documents.
+* `Proposal`:
+  a comma-separated list of links to proposal documents which are related to this vision.
+  This field is typically not used when a vision is first written, since there are no related proposals
+  for it yet, but it may be added retrospectively. Optional.
+
+These standard fields should be provided in this order.
+
 </details>
 
 [forums]: https://forums.swift.org/
--- a/visions/0001-using-swift-from-c++.md
+++ b/visions/0001-using-swift-from-c++.md
@@ -1,0 +1,10 @@
+# Using Swift from C++
+
+* Vision: [SV-0001](using-swift-from-c++.md)
+
+> [!NOTE]
+> This file exists only to reserve vision ID 0001. The `visions/` directory did
+> not originally use numbered identifiers; this ID was assigned retroactively
+> when numbering was introduced later on. The original document, which predates
+> versioning, can be found at
+> [`visions/using-swift-from-c++.md`](using-swift-from-c++.md).
--- a/visions/0002-macros.md
+++ b/visions/0002-macros.md
@@ -1,0 +1,9 @@
+# A Vision for Macros in Swift
+
+* Vision: [SV-0002](macros.md)
+
+> [!NOTE]
+> This file exists only to reserve vision ID 0002. The `visions/` directory did
+> not originally use numbered identifiers; this ID was assigned retroactively
+> when numbering was introduced later on. The original document, which predates
+> versioning, can be found at [`visions/macros.md`](macros.md).
--- a/visions/0003-using-c++-from-swift.md
+++ b/visions/0003-using-c++-from-swift.md
@@ -1,0 +1,10 @@
+# Using C++ from Swift
+
+* Vision: [SV-0003](using-c++-from-swift.md)
+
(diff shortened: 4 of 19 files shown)


FranzBusch (commented):
Thanks for raising the PR. Will discuss it in the next ESG meeting.

rjmccall:
The LSG talked about this, and we're weakly in favor of it. We think it's probably less useful to assign specific numbers to visions than it is for proposal documents, but there's still some value here.

There are a handful of legacy design documents (e.g. the ownership and error-handling manifestos) that are part of Swift's history and probably ought to be recognized as rough peers to the vision documents (although they generally weren't approved in any formalized way). Those would not fit chronologically into the numbering system unless we set aside room for them now.

rjmccall:
I think the basic process for acceptance here is that the ESG, LSG, and PSG each need to sign off, so right now you're sitting at 1 / 3.

stmontgomery:
> The LSG talked about this, and we're weakly in favor of it. We think it's probably less useful to assign specific numbers to visions than it is for proposal documents, but there's still some value here.

Great, thanks for the update and for representing it with LSG.

> There are a handful of legacy design documents (e.g. the ownership and error-handling manifestos) that are part of Swift's history and probably ought to be recognized as rough peers to the vision documents (although they generally weren't approved in any formalized way). Those would not fit chronologically into the numbering system unless we set aside room for them now.

Do you have links to the canonical versions of those non-vision-but-still-important documents handy? If so, I could attempt to find a creation date for them and reserve the numbers, even if we do not formally move them into the repo.

> I think the basic process for acceptance here is that the ESG, LSG, and PSG each need to sign off, so right now you're sitting at 1 / 3.

@FranzBusch, were you planning to represent this at ESG, or could you?

@al45tair, I wonder if you could get involved here and add this to the agenda at some upcoming PSG meeting? I'm happy to help represent this wherever it's helpful.

xwu:
> Do you have links to the canonical versions of those non-vision-but-still-important documents handy? If so, I could attempt to find a creation date for them and reserve the numbers, even if we do not formally move them into the repo.

Most of these that I can think of live in https://github.com/swiftlang/swift/tree/main/docs; among them:

* ABI Stability Manifesto
* Access Control
* Differentiable Programming Manifesto
* Error Handling Rationale
* Generics Manifesto
* Library Evolution Manifesto
* Ownership Manifesto
* String Manifesto
* _(even older)_ Sequences and Collections, String Design, Text Formatting

ktoso:
Thanks for making the idea into PR @stmontgomery 👍

FranzBusch:
Sorry for the delay in update here. The ESG discussed this and we are in favor of adopting a numbering scheme. While we agree with the LSGs feedback that there is only light value in this compared to numbering for evolution proposals we still see value and it doesn't impose any burden on vision document authors.

One personal review comment, should we flip the proposed linking from the new numbered document containing the content and the old document containing a link to the new one instead? This would result in more coherence for the future.

stmontgomery:
> Sorry for the delay in update here. The ESG discussed this and we are in favor of adopting a numbering scheme. While we agree with the LSGs feedback that there is only light value in this compared to numbering for evolution proposals we still see value and it doesn't impose any burden on vision document authors.

Great, thanks for the update!

> One personal review comment, should we flip the proposed linking from the new numbered document containing the content and the old document containing a link to the new one instead? This would result in more coherence for the future.

I thought about this, and remain open to whatever the community feels is best, but here was my thought process:

In an ideal world what I'd _really_ prefer is to set up an automatic redirect so that we could change the existing docs to be numbered and have redirects at the old locations. However, after some research I don't see any way to do that on GitHub today. (I investigated using HTML `<meta>` tags in Markdown, among other things, but nothing seemed to work.) Given that limitation, I felt we should prioritize keeping existing links working and pointing _directly_ at the proposal documents so that people don't have to perform an extra click to get where they're trying to go. Installing breadcrumb files at the existing locations would presumably also affect "link previews" in popular apps like Slack or the Swift Forums site by truncating their text content.

If we proceed with this new numbering scheme, any _new_ vision documents going forward could directly use the `NNNN-` prefix and not require a redirect file, of course. Again, I'm open to counterarguments here, but hopefully this rationale makes sense.

al45tair:
PSG is going to look at this today; I'll report back later.

al45tair:
PSG takes the same view as LSG; we're weakly in favour.

al45tair (approved):

PR #3367, opened 2026-07-21
InlineArray: Hashable.

stephentyrone:


xwu:
IMO, we should discuss whether this ought to be a test case for a 'mini-proposal' process.

PR #3370, opened 2026-07-23
[SE-0414]: Clerify task isolated isolation region terminology.

noamfreeman:
the terms `task isolated isolation domain` and `task isolated isolation region` seem much harder to read  and understand then `task-isolated isolation domain`

this also fits the use of `actor-isolated` that appears many times in this proposal.


rjmccall (approved):
Sure, this seems reasonable.

PR #3379, opened 2026-07-28
Introduce Aliased*Span and Aliased*Ref family of types

DougGregor:


DougGregor:
Moving back to draft because we're not sure this is the path we want to bring up for review

PR #3381, opened 2026-07-29
Normalize SE-0304 status field

dempseyatgithub:
The link to the acceptance discussion thread is already included in the standard 'Review' header field.

This PR normalizes the 'Status' header field by removing the additional link to the acceptance discussion.


rjmccall (approved):
Thanks!

PR #3386, opened 2026-07-31
Normalize and correct SE-0446 header fields

dempseyatgithub:
- Move acceptance link to Review header field
- Add missing review link to Review header field
- Remove erroneous Upcoming Feature Flag header field

This proposal incorrectly lists its *experimental* feature flag `NonescapableTypes` as an *upcoming* feature flag.

The flag `NonescapableTypes` appears in the 6.0 release branch [Features.def](https://github.com/swiftlang/swift/blob/release/6.0/include/swift/Basic/Features.def#L359) file as an experimental flag.

It currently appears as a baseline language feature in the 6.3 release [Features.def](https://github.com/swiftlang/swift/blob/release/6.3/include/swift/Basic/Features.def#L256) file.

This flag has never been in a release as an upcoming feature flag.  The proposal itself also lists no source compatibility issues that would require an upcoming feature flag.

Since upcoming feature flags are reported on the evolution dashboard as part of a proposal's metadata, this change will also fix the erroneous dashboard entry.

### DIFF
--- a/proposals/0446-non-escapable.md
+++ b/proposals/0446-non-escapable.md
@@ -6,9 +6,7 @@
 * Status: **Implemented (Swift 6.2)**
 * Roadmap: [BufferView Language Requirements](https://forums.swift.org/t/roadmap-language-support-for-bufferview)
 * Implementation: **Implemented** in `main` branch
-* Upcoming Feature Flag: `NonescapableTypes`
-* Review: ([pitch](https://forums.swift.org/t/pitch-non-escapable-types-and-lifetime-dependency/69865))
-* Decision Notes: [Acceptance](https://forums.swift.org/t/accepted-with-modifications-se-0446-nonescapable-types/75504)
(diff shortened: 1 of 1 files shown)


rjmccall:
Let me ask the LSG how we feel about removing the experimental feature flag. I agree it shouldn't be listed as an upcoming feature flag if we ended up not needing an upcoming feature for it.

rjmccall:
I talked about this with the LSG, and we think:
1. We should add an Experimental Feature Flag header field that we can use for features still undergoing review. This field should generally be removed once the feature is accepted. (This is something that your script could check, James.)
2. We should be explicit that Upcoming Feature Flag should only be used when you think the proposal will need one. So the existence of this field implies a non-trivial discussion in the Source Compatibility section.

We discussed whether we should have a documented name for identifying features for has-feature checks, and we agreed that we don't actually want people to check for always-on features this way. We want programmers to be checking tools versions, both so that they can recognize obvious associations (no need to support the combination of +X / -Y when there are no actual releases with that set of implemented features) and so that they can more easily eliminate unnecessary checks when they bump their minimum supported tools release.

I'll create a PR that makes these changes to the templates. Meanwhile, this PR can just be merged. Thanks!

rjmccall (approved):

rjmccall:
That PR is up [here](https://github.com/swiftlang/swift-evolution/pull/3419).

PR #3401, opened 2026-08-07
SE-NNNN: Allow Deprecating Package Products

bkhouri:
- [Pitch thread](https://forums.swift.org/t/pitch-deprecating-package-products-in-the-package-manifest/88468).
- Implementation
  - https://github.com/swiftlang/swift-package-manager/pull/10437
  - https://github.com/swiftlang/swift-package-manager/pull/10438

------

Adds a new SwiftPM evolution proposal, *Allow Deprecating Package Products*, which introduces a first-class way for package authors to mark a package `.library`, `.executable`, or `.plugin` product as unsupported directly in `Package.swift`.

Key points of the proposed design captured in the document:

- New `deprecated:` parameter on the `Product` factories, taking an opaque `Product.Deprecation` value constructed via `.unsupported(message:replacement:)`.
- Optional `replacement:` locator (`.renamed(_:package:)`) pointing consumers at the successor product, either in the same package or in a different one.
- SwiftPM emits a build-time warning whenever a consumer depends on an
unsupported product, surfacing the author-supplied message and replacement.
- New `swift package audit` subcommand that enumerates unsupported products reachable from the current package graph, with an exit status suitable for CI gating.
- Existing manifests are unaffected — `deprecated:` is optional and defaults to `nil`.

### DIFF
--- a/proposals/NNNN-swiftpm-allow-deprecating-package-products.md
+++ b/proposals/NNNN-swiftpm-allow-deprecating-package-products.md
@@ -1,0 +1,609 @@
+# SE-NNNN: Allow Deprecating Package Products
+
+* Proposal: [SE-NNNN](NNNN-swiftpm-product-deprecation.md)
+* Authors: [Sam Khouri](https://github.com/bkhouri)
+* Review Manager: TBD
+* Status: **Awaiting implementation**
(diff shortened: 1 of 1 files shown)


owenv:
@swiftlang/ecosystem-steering-group we discussed this proposal in the build and packaging workgroup today and consensus was it was ready to proceed to review

PR #3404, opened 2026-08-08
Revise SE-0542

clive819:
Address comments in https://forums.swift.org/t/se-0542-package-manager-condition-plugins/88820


benrimmington (changes requested):
@clive819 I don't think proposal filenames should be changed, because any links in the forums (or elsewhere) will be broken. There isn't an automatic redirection, unlike repository renames.

For example, the proposal for `InlineArray` is still named [`0453-vector.md`](<https://github.com/swiftlang/swift-evolution/blob/main/proposals/0453-vector.md>).

The only counterexample I can find is when [`0519-borrow-inout-types.md`](<https://github.com/swiftlang/swift-evolution/blob/main/proposals/0519-borrow-inout-types.md>) was renamed to [`0519-ref-mutableref-types.md`](<https://github.com/swiftlang/swift-evolution/blob/main/proposals/0519-ref-mutableref-types.md>). But that required a hard-coded workaround in swiftlang/swift-evolution-metadata-extractor#106.

dempseyatgithub:
> @clive819 I don't think proposal filenames should be changed, because any links in the forums (or elsewhere) will be broken. There isn't an automatic redirection, unlike repository renames.
> 
> For example, the proposal for `InlineArray` is still named [`0453-vector.md`](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0453-vector.md).
> 
> The only counterexample I can find is when [`0519-borrow-inout-types.md`](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0519-borrow-inout-types.md) was renamed to [`0519-ref-mutableref-types.md`](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0519-ref-mutableref-types.md). But that required a hard-coded workaround in [swiftlang/swift-evolution-metadata-extractor#106](https://github.com/swiftlang/swift-evolution-metadata-extractor/pull/106).

Yes, the name of the file should be kept the same.

The counterexample cited (SE-0519) needed to be special-cased in the swift-evolution-metadata-extractor tool to ignore the now-incorrect file. (For over a decade it has been an invariant that each proposal number has a single markdown file, and that invariant is relied on. From an automatic processing perspective, it would be much better for that to be a very rare exception as opposed to a common case which adds additional complexity for very little seeming benefit.)

The name of the file does not need to exactly match the title of the proposal.  People generally read the proposal title not the file name, allowing flexibility in renaming the title without having to rename the file.

proposals/0542-package-manager-conditional-plugin.md
```diff
@@ -1,28 +1,28 @@
-# Package Manager Conditional Plugin
+# Package Manager Build Tool Plugin Usage Conditionals
 
 * Proposal: [SE-0542](0542-package-manager-conditional-plugin.md)
 * Authors: [Clive Liu](https://github.com/clive819)
 * Review Manager: [David Cummings](https://github.com/daveyc123)
 * Status: **Active Review (August 6 - August 21)**
```
dempseyatgithub: Review date ranges should include the year.

```suggestion
* Status: **Active Review (August 6 - August 21, 2026)**
```

PR #3418, opened 2026-08-13
SE-NNNN: @noSanitize attribute for functions

ndrewh:
Pitch: https://forums.swift.org/t/pitch-nosanitize-attribute-for-functions/88972
Implementation: https://github.com/swiftlang/swift/pull/91137/changes

### DIFF
--- a/proposals/NNNN-nosanitize-attribute.md
+++ b/proposals/NNNN-nosanitize-attribute.md
@@ -1,0 +1,95 @@
+# @noSanitize attribute for functions
+
+* Proposal: [SE-NNNN](NNNN-nosanitize-attribute.md]
+* Authors: [Andrew Haberlandt](https://github.com/ndrewh)
+* Review Manager: TBD
+* Status: **Awaiting Review**
+* Implementation: https://github.com/swiftlang/swift/pull/91137/changes
+* Review: ([pitch](tbd))
+
+## Summary of changes
+
+This proposal introduces a new attribute `@noSanitize(<kind>)` which can be applied to functions and subscripts in order to disable certain types of sanitizer instrumentation. This can be useful when certain code is known to trigger sanitizer false positives or cause performance issues when sanitized.
+Just like the clang `__attribute__((no_sanitize("<kind>")))` attribute, the proposed Swift `@noSanitize(<kind>)` attribute prevents Swift from adding the `sanitize_address` (or `sanitize_thread`, `sanitize_memtag`, etc.) attribute to a function when emitting LLVM IR.
+
+## Motivation
+
+Sanitizers such as ASan and TSan rely on instrumentation to check the correctness of every memory access (e.g. by using shadow memory regions that encode whether a location is valid to access). Especially in embedded environments, sanitizers may fail to correctly track the state of certain memory locations (e.g. MMIO), causing false positives.
+In other cases, some functions may deliberately (i.e. some `strlen` implementations) read beyond the bounds of a memory object, triggering undesirable (yet real) sanitizer reports.
+Even when sanitizers do not produce unwanted reports, certain code may disproportionately contribute to the instrumentation overhead of a sanitizer.
+
+## Proposed solution
+
+Allow individual functions to be opted out of sanitizer instrumentation with an attribute. Each sanitizer is opted out independently.
+
+## Detailed design
+
+The `@noSanitize(<kind>)` attribute takes a single required argument naming the sanitizer to suppress. The initially supported kinds are:
+
+- `address` — suppresses ASan (`sanitize_address`) instrumentation.
+- `thread` — suppresses TSan (`sanitize_thread`) instrumentation.
+- `memtag` — suppresses [MemTag](https://llvm.org/docs/MemTagSanitizer.html) stack tagging (`sanitize_memtag`) instrumentation.
+
+Each kind opts out independently, so `@noSanitize(address)` on a function built with `-sanitize=thread` has no effect. Multiple `@noSanitize` attributes may be stacked on the same declaration to opt out of more than one sanitizer.
+
+The attribute is accepted on any function (top-level `func`, methods, initializers, deinitializers, and accessors such as `get`/`set`/`_read`/`_modify`) and on subscripts.
+
+```swift
+@noSanitize(address)
+func readsMMIO(_ p: UnsafePointer<UInt32>) -> UInt32 { p.pointee }
+
+struct Device: ~Copyable {
+  @noSanitize(address) init() { ... }
+
+  @noSanitize(address) deinit { ... }
+
+  var status: UInt32 {
+    @noSanitize(address) get { ... }
+    @noSanitize(address) set { ... }
+  }
+
+  @noSanitize(address)
+  subscript(i: Int) -> UInt8 { ... }
+}
+
+// Stacking is allowed to opt out of more than one sanitizer.
+@noSanitize(address)
+@noSanitize(thread)
+func hotPath() -> Int { ... }
+```
+
+### Interaction with inlining
+
+Inlining a `@noSanitize` callee into a caller that is still being instrumented would silently re-instrument the callee's body, defeating the attribute. To preserve the guarantee, the SIL performance inliner refuses to inline a `@noSanitize(<kind>)` callee into a caller that does not carry the same `@noSanitize(<kind>)` when that sanitizer is enabled for the current build.
+
+The restriction is one-directional: a regular (sanitized) callee may still be inlined into a `@noSanitize` caller (and thus may lose its instrumentation). Users who want a `@noSanitize` function to be inlined into ordinary sanitized code should either mark the caller with a matching `@noSanitize` or accept that the callee will remain an out-of-line call in sanitized builds. `@inline(__always)` does not override this restriction.
+
+## Source compatibility
+
+This is a pure extension with no source compatibility impact.
+
+## ABI compatibility
+
+This attribute is applied to deliberately disable sanitizer instrumentation on individual functions. In all currently supported sanitizers, functions compiled with a sanitizer are ABI-compatible with unsanitized functions.
+
+## Implications on adoption
+
+This feature can be freely adopted and un-adopted in source code and is not tied to any runtime support.
+
+## Future Directions
+
+### Closures
+
+The attribute is currently limited to declarations (`OnAbstractFunction | OnSubscript`), so it cannot be written on a closure expression:
+
+```swift
+registerCallback { @noSanitize(address) in ... }
+```
+
+### Globals
+
+Some sanitizers support clang's `no_sanitize` to disable sanitization of globals. We could support a similar attribute for Swift globals.
+
+### Additional sanitizers
+
+Additional sanitizer kinds may need to be added in the future when supported by swiftc's -sanitize option.


proposals/NNNN-nosanitize-attribute.md
```diff
@@ -0,0 +1,95 @@
...
+* Proposal: [SE-NNNN](NNNN-nosanitize-attribute.md]
+* Authors: [Andrew Haberlandt](https://github.com/ndrewh)
+* Review Manager: TBD
+* Status: **Awaiting Review**
+* Implementation: https://github.com/swiftlang/swift/pull/91137/changes
+* Review: ([pitch](tbd))
+
+## Summary of changes
+
+This proposal introduces a new attribute `@noSanitize(<kind>)` which can be applied to functions and subscripts in order to disable certain types of sanitizer instrumentation. This can be useful when certain code is known to trigger sanitizer false positives or cause performance issues when sanitized.
```
allevato, line 12: The "Summary of Changes" section should be 50 words or less. Can you please shorten it? Just focus specifically on what the proposal changes about the language; save the "why this is useful" or Clang comparisons for later in the document.

proposals/NNNN-nosanitize-attribute.md
```diff
@@ -0,0 +1,95 @@
...
+
+Allow individual functions to be opted out of sanitizer instrumentation with an attribute. Each sanitizer is opted out independently.
+
+## Detailed design
+
+The `@noSanitize(<kind>)` attribute takes a single required argument naming the sanitizer to suppress. The initially supported kinds are:
+
+- `address` — suppresses ASan (`sanitize_address`) instrumentation.
+- `thread` — suppresses TSan (`sanitize_thread`) instrumentation.
+- `memtag` — suppresses [MemTag](https://llvm.org/docs/MemTagSanitizer.html) stack tagging (`sanitize_memtag`) instrumentation.
```
allevato, line 31: Do you also want to add `coverage` here since it came up during the pitch discussion? (Again, this is your proposal, so it's your call whether you want to put that in or not.)

proposals/NNNN-nosanitize-attribute.md
```diff
@@ -0,0 +1,95 @@
...
+In other cases, some functions may deliberately (i.e. some `strlen` implementations) read beyond the bounds of a memory object, triggering undesirable (yet real) sanitizer reports.
+Even when sanitizers do not produce unwanted reports, certain code may disproportionately contribute to the instrumentation overhead of a sanitizer.
+
+## Proposed solution
+
+Allow individual functions to be opted out of sanitizer instrumentation with an attribute. Each sanitizer is opted out independently.
+
+## Detailed design
+
+The `@noSanitize(<kind>)` attribute takes a single required argument naming the sanitizer to suppress. The initially supported kinds are:
```
allevato, line 27: Would you like to update this to be a comma-delimited list of sanitizer names, as came up during the pitch discussion? (This is your proposal, so it's your call whether you want to put that in or not.)

proposals/NNNN-nosanitize-attribute.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Implications on adoption
+
+This feature can be freely adopted and un-adopted in source code and is not tied to any runtime support.
+
+## Future Directions
+
+### Closures
+
+The attribute is currently limited to declarations (`OnAbstractFunction | OnSubscript`), so it cannot be written on a closure expression:
```
allevato, line 83: The level of implementation detail you have here (mentioning the specific attribute flags from the .def file) should be removed. The audience for most of the content in a proposal is a Swift language user, not a compiler engineer.

proposals/NNNN-nosanitize-attribute.md
```diff
@@ -0,0 +1,95 @@
...
+
+This attribute is applied to deliberately disable sanitizer instrumentation on individual functions. In all currently supported sanitizers, functions compiled with a sanitizer are ABI-compatible with unsanitized functions.
+
+## Implications on adoption
+
+This feature can be freely adopted and un-adopted in source code and is not tied to any runtime support.
+
+## Future Directions
+
+### Closures
```
allevato, line 81: There are actually two separate questions about closures that we believe need more discussion (and decisions to be made):

1. Can closures be supported now, if there are no major implementation concerns that would give reason to subset them out?
2. Regardless of the answer to the above question, does disabling a sanitizer propagate down from a function into its closures? For example:

  ```swift
  @noSanitize(thread)
  func f() {
    callSomething {
      doSomethingElse()
    }
  }
  ```
  Given that example, are sanitizers also disabled for the closure passed to `callSomething` (and any closures that it might contain, recursively)?

There is some prior art we can consider:

* From my experimentation, Clang appears to *not* propagate `no_sanitize` from a function to its lambdas in C++.
* Swift recently had this same question when it extended `@section` to functions in SE-0537. There, we chose to propagate it to closures because there is such a proliferation of closure-based APIs in Swift that they're effectively more like control flow constructs than separate functions.

If the choice is to make it propagate, then we have to consider the fact that as proposed, there's no syntax to re-enable the sanitizer for a closure if that's desired. (For `@section`, we handled this by introducing `@section(default)` to turn off any propagated inference.)

proposals/NNNN-nosanitize-attribute.md
```diff
@@ -0,0 +1,95 @@
...
+@noSanitize(address)
+@noSanitize(thread)
+func hotPath() -> Int { ... }
+```
+
+### Interaction with inlining
+
+Inlining a `@noSanitize` callee into a caller that is still being instrumented would silently re-instrument the callee's body, defeating the attribute. To preserve the guarantee, the SIL performance inliner refuses to inline a `@noSanitize(<kind>)` callee into a caller that does not carry the same `@noSanitize(<kind>)` when that sanitizer is enabled for the current build.
+
+The restriction is one-directional: a regular (sanitized) callee may still be inlined into a `@noSanitize` caller (and thus may lose its instrumentation). Users who want a `@noSanitize` function to be inlined into ordinary sanitized code should either mark the caller with a matching `@noSanitize` or accept that the callee will remain an out-of-line call in sanitized builds. `@inline(__always)` does not override this restriction.
```
allevato, line 65: `@inline(always)` was made an official spelling in SE-0496, so use that here instead of the underscored version.

proposals/NNNN-nosanitize-attribute.md
```diff
@@ -0,0 +1,95 @@
...
+  subscript(i: Int) -> UInt8 { ... }
+}
+
+// Stacking is allowed to opt out of more than one sanitizer.
+@noSanitize(address)
+@noSanitize(thread)
+func hotPath() -> Int { ... }
+```
+
+### Interaction with inlining
```
allevato, line 61: Can you expand on the interaction with `@inline(always)`? You mention that it does override the restriction, but does that mean that it emits an error diagnostic, or it silently doesn't inline the function?

In particular, we're concerned about how this interacts with what was accepted in [SE-0496](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0496-inline-always.md), which describes the rules around when diagnostics are emitted if an `@inline(always)` function failed to be inlined, such as "[w]e only emit an error if the annotated function is directly referenced and something would cause it to be not inlined or if some property at the declaration site of the function would make it not possible in the common case."

We also have some concerns about whether this becomes a trap for some performance scenarios where you may apply some combination of these attributes in the call hierarchy.

`@inline(always)` in Swift is perhaps closer to Clang's `__attribute__((always_inline))` than just regular `(static) inline`, and Clang explicitly documents combining ` __attribute__((always_inline))` and ` __attribute__((no_sanitize(...)))` as [unsupported and [leading] to unpredictable results](https://clang.llvm.org/docs/AddressSanitizer.html#interaction-of-inlining-with-disabling-sanitizer-instrumentation). The official recommendation is to use the preprocessor to define a macro that expands to either ` __attribute__((always_inline))` or nothing depending on whether a sanitizer is enabled, but Swift has no such capability.

If there were something like a `sanitized(...)` compilation condition, one could at least write something like this:

```swift
#if !sanitized(address)
@inline(always)
#endif
func f() { ... }
```

So to summarize, there are (at least) two points that need to be considered here:

* Does mixing `@noSanitize` and `@inline(always)` break the latter's diagnostic contract?
* Can the unpredictability that Clang documents be avoided for Swift?

proposals/NNNN-nosanitize-attribute.md
```diff
@@ -0,0 +1,95 @@
...
+
+The `@noSanitize(<kind>)` attribute takes a single required argument naming the sanitizer to suppress. The initially supported kinds are:
+
+- `address` — suppresses ASan (`sanitize_address`) instrumentation.
+- `thread` — suppresses TSan (`sanitize_thread`) instrumentation.
+- `memtag` — suppresses [MemTag](https://llvm.org/docs/MemTagSanitizer.html) stack tagging (`sanitize_memtag`) instrumentation.
+
+Each kind opts out independently, so `@noSanitize(address)` on a function built with `-sanitize=thread` has no effect. Multiple `@noSanitize` attributes may be stacked on the same declaration to opt out of more than one sanitizer.
+
+The attribute is accepted on any function (top-level `func`, methods, initializers, deinitializers, and accessors such as `get`/`set`/`_read`/`_modify`) and on subscripts.
```
allevato, line 35: You can leave out the unsupported accessors here. I think it would be sufficient to just say "accessors" and leave out the "such as..." list.

proposals/NNNN-nosanitize-attribute.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Summary of changes
+
+This proposal introduces a new attribute `@noSanitize(<kind>)` which can be applied to functions and subscripts in order to disable certain types of sanitizer instrumentation. This can be useful when certain code is known to trigger sanitizer false positives or cause performance issues when sanitized.
+Just like the clang `__attribute__((no_sanitize("<kind>")))` attribute, the proposed Swift `@noSanitize(<kind>)` attribute prevents Swift from adding the `sanitize_address` (or `sanitize_thread`, `sanitize_memtag`, etc.) attribute to a function when emitting LLVM IR.
+
+## Motivation
+
+Sanitizers such as ASan and TSan rely on instrumentation to check the correctness of every memory access (e.g. by using shadow memory regions that encode whether a location is valid to access). Especially in embedded environments, sanitizers may fail to correctly track the state of certain memory locations (e.g. MMIO), causing false positives.
+In other cases, some functions may deliberately (i.e. some `strlen` implementations) read beyond the bounds of a memory object, triggering undesirable (yet real) sanitizer reports.
```
allevato, line 18: Since we're writing Swift, where reading beyond the bounds of a memory object is strictly forbidden, I wonder if there's a better example you can cite here.

proposals/NNNN-nosanitize-attribute.md
```diff
@@ -0,0 +1,95 @@
...
+
+Inlining a `@noSanitize` callee into a caller that is still being instrumented would silently re-instrument the callee's body, defeating the attribute. To preserve the guarantee, the SIL performance inliner refuses to inline a `@noSanitize(<kind>)` callee into a caller that does not carry the same `@noSanitize(<kind>)` when that sanitizer is enabled for the current build.
+
+The restriction is one-directional: a regular (sanitized) callee may still be inlined into a `@noSanitize` caller (and thus may lose its instrumentation). Users who want a `@noSanitize` function to be inlined into ordinary sanitized code should either mark the caller with a matching `@noSanitize` or accept that the callee will remain an out-of-line call in sanitized builds. `@inline(__always)` does not override this restriction.
+
+## Source compatibility
+
+This is a pure extension with no source compatibility impact.
+
+## ABI compatibility
```
allevato, line 71: If a function is `@inlinable`, uses `@noSanitize`, and is built in a module with library evolution enabled, is the `@noSanitize` attribute preserved in the textual .swiftinterface file? I believe it would have to be.

allevato (commented):
Hello! I'll be review managing this proposal. We discussed this proposal in the Language Steering Group today. I have some feedback from the group, some editorial and some more substantial. Could you please update the proposal to address the following?

proposals/NNNN-nosanitize-attribute.md
ndrewh: When I initially wrote the proposal, I had imagined that it would silently just not inline the function. It seems like that is not the right behavior.

> Does mixing `@noSanitize` and `@inline(always)` break the latter's diagnostic contract?

My read of SE-0496 is that we should emit an error if these attributes are used on the same function, based on "if some property at the declaration site of the function would make it not possible in the common case".

I would like to propose that we emit an error if both `@noSanitize` and `@inline(always)` are present on a function, and one of the noSanitize sanitizers is turned on.

> Can the unpredictability that Clang documents be avoided for Swift?

I am not sure where the documented unpredictability stems from, but the interaction between these two options seems complex enough that I am not sure it's worth trying to support the use of these options simultaneously.

Would a reasonable path forward be: this proposal will state that using `@inline(always)` with `@noSanitize` is   not supported, and results in an error. If some future proposal wishes to define this behavior, they can still do so?

> If there were something like a sanitized(...) compilation condition, one could at least write something like this:

I think we need this, and I don't really see a way around adding it before we can do `@noSanitize` -- because of the incompatibility with `@inline(always)`. What's the right path to doing this -- make another proposal for that feature, or can I roll it into this one?

allevato: > Would a reasonable path forward be: this proposal will state that using @inline(always) with @noSanitize is not supported

I think so, as long as it also comes with the compiler condition. The fundamental concern is that when it comes to inlining and disabling sanitizers, you _must_ sacrifice one: if you had a hypothetical `@noSanitize(...) @inline(always)` function and it's called from a function with a different `@noSanitize` "color", then:

* inlining it would cause it to inherit the sanitization behavior of the caller, violating the `@noSanitize` attribute, or
* the compiler would have to refuse to inline it, violating the `@inline(always)` attribute

This problem doesn't exist if the caller and callee have compatible `@noSanitize` "colors" but that requires you to know all the potential callers of the function, and that moves the diagnostic from a declaration site issue to a use site issue. In other words, one of the attributes would have to be silently ignored in some cases or you would have to ban certain callers, which might put you in a situation where a function is uncallable in certain contexts (or worse, only in certain compilation modes!). The compiler condition gives you the escape hatch.

Being more conservative and disallowing the combination seems reasonable, and could potentially be lifted later.

> I think we need this, and I don't really see a way around adding it before we can do @noSanitize -- because of the incompatibility with @inline(always). What's the right path to doing this -- make another proposal for that feature, or can I roll it into this one?

Since they're related features, I think it would be fine to roll it into the same proposal.

proposals/NNNN-nosanitize-attribute.md
ndrewh: I decided to support closures, and to not propagate.

proposals/NNNN-nosanitize-attribute.md
ndrewh: Yes, I like that idea. Done.

proposals/0550-nosanitize-attribute.md
```diff
@@ -0,0 +1,115 @@
...
+Because `@inline(always)` and `@noSanitize` cannot be combined on the same declaration, a mechanism is needed to support functions that are `@inline(always)` in non-sanitized builds but out-of-line-and-uninstrumented in sanitized builds.
+A new `sanitized(<kind>)` compilation condition, with the same supported kinds as `noSanitize`, will evaluate true if the specified sanitizer is enabled. Each `sanitized` condition may only list a single sanitizer kind, but can
+be combined using the usual logical operators.
+
+```swift
+#if !sanitized(address)
+@inline(always)
+#else
+@noSanitize(address)
+#endif
```
allevato, line 89: FYI, I made a small tweak here because this needs to be an `#if/#else/#endif` so that the two attributes are mutually exclusive, not just guarding the `@inline` itself.

allevato (approved):

PR #3419, opened 2026-08-13
Break out a separate, temporary field for experimental feature flags

rjmccall:
See https://github.com/swiftlang/swift-evolution/pull/3386#issuecomment-5184944726.


dempseyatgithub:
I've created an issue for the metadata extraction tool to validate that the experimental feature flag is not present when the status of the proposal is 'Implemented'. https://github.com/swiftlang/swift-evolution-metadata-extractor/issues/126

A question: Should the experimental feature flag remain when the proposal moves to other end states like 'Rejected' or 'Withdrawn'?

rjmccall:
> A question: Should the experimental feature flag remain when the proposal moves to other end states like 'Rejected' or 'Withdrawn'?

LSG consensus: it should be removed when the proposal reaches any end state.

PR #3421, opened 2026-08-17
Fix code example in Embedded vision

kkebo:
The typo I fixed in PR #3415 was not a typo of `Any`. Looking at the surrounding code, it should be `any P` rather than just `Any`.


rjmccall:
Thanks!

PR #3423, opened 2026-08-18
Fix incorrect filename for SE-0542 in proposal link

dempseyatgithub:
This PR corrects the filename in the proposal link for SE-0542.

The incorrect link causes a 404 error when clicking the proposal on the evolution dashboard instead of navigating to the proposal.

I have also filed an issue for the metadata extractor tool to validate the filename to catch similar errors in the future. https://github.com/swiftlang/swift-evolution-metadata-extractor/issues/124


dempseyatgithub:
cc @daveyc123

rjmccall (approved):

PR #3427, opened 2026-08-19
Fix links to Error Handling Rationale document

edymtt:
That has changed extension as a result of swiftlang/swift#62318


xwu:
Thanks!

PR #3428, opened 2026-08-20
[Proposal] resignRemoteID for remote distributed actor references

ktoso:
This is a small proposal to update some distributed actor system semantics.

Implementation incoming any moment now...


proposals/0546-resign-remote-id.md
```diff
@@ -0,0 +1,114 @@
+# resignRemoteID for remote distributed actor references
+
+* Proposal: [SE-0546](0546-resign-remote-id.md)
```
ktoso: Seems the PR validator is nowadays upset about the NNNN-... are we supposed to just keep it failing (seems like a meh pattern), or just guess a number, neither seems like optimal workflow?

allevato: I think it's fine if the workflow fails for the proposal author; it really only matters to make sure the _review manager doesn't merge it_ with a mistake in any of the required fields. Until the review begins, you should move it back to `SE-NNNN`.

I thought the workflow was set up in such a way that it would take the state into consideration, but maybe not?

ktoso: Yeah I got failures about the N not being numbers: ([this run](https://github.com/swiftlang/swift-evolution/actions/runs/32334157215/job/96320323155?pr=3428))

```
ISSUES FOUND IN 1 PROPOSAL

SE-NNNN 'resignRemoteID for remote distributed actor references'

	ERROR: Proposal ID must include four decimal digits. <<
	ERROR: Proposal ID link must be a relative link [SE-NNNN](NNNN-filename.md).
	ERROR: Missing Review field.
	ERROR: Invalid proposal status.
```

It is a bit weird to have proposal authors make PRs and expect the CI to be failing; The usual reaction to a failing CI I have is to make it pass, but that's the wrong thing to do in this case, as the number is only assigned later 🤔 

Is this something you're still polishing/working on @dempseyatgithub?

allevato: If the workflow doesn't do it already, I'd expect the behavior to be:

* If the status is "Awaiting review", ensure that review manager is TBD, number is "NNNN", fail otherwise
* Any other status, ensure review manager and number are assigned, fail otherwise

proposals/0546-resign-remote-id.md
```diff
@@ -0,0 +1,114 @@
...
+* Authors: [Konrad 'ktoso' Malawski](https://github.com/ktoso)
+* Review Manager: TBD
+* Status: **Awaiting review**
+* Implementation: [swiftlang/swift#91610](https://github.com/swiftlang/swift/pull/91610)
```
ktoso:
```suggestion
* Implementation: [swiftlang/swift#91610](https://github.com/swiftlang/swift/pull/91610), available under experimental feature flag `DistributedActorResignRemoteID `
```

ktoso:
This is now implemented under experimental flag `DistributedActorResignRemoteID`. 
Pitch also [posted](https://forums.swift.org/t/pitch-resignremoteid-for-remote-distributed-actor-references/89078) since a while.

This could proceed into a formal review when time allows, cc @rjmccall @hborla

proposals/0546-resign-remote-id.md
ktoso: Yeah it didn't work like that, here's a patch to fix it: https://github.com/swiftlang/swift-evolution-metadata-extractor/pull/137 cc @allevato @dempseyatgithub

dempseyatgithub: The goal of the CI really is to ensure the proposal isn't merged by the review manager with validation errors, which is a major cause of thrash with the evolution dashboard. (merge happens with errors > proposal is omitted or incorrect on dashboard > 1-line fix submitted > review manager needs to fix).

The tool itself is close to being extended to support validation and extraction of additional projects like swift-testing and swift-foundation, which would give the community much more visibility into the evolution process of these widely used projects. I'd prefer to do that first, before extending the validation logic to handle status-specific validation for PRs that are not intended to be committed in that state. (i.e would prefer to generalize for multiple projects before increasing complexity of the tool.)

@allevato @ktoso

DougGregor:
I'll be managing this review

proposals/0546-resign-remote-id.md
dempseyatgithub: > If the workflow doesn't do it already, I'd expect the behavior to be:
> 
> * If the status is "Awaiting review", ensure that review manager is TBD, number is "NNNN", fail otherwise
> * Any other status, ensure review manager and number are assigned, fail otherwise

I've been thinking about this for a while and the purpose of the CI validation is to ensure the proposal in the PR is valid to be committed.

Historically at least, a proposal is *always* be assigned a number before it is committed, so the CI should *always* fail without an assigned proposal number.

(If that longtime invariant changes, additional work would probably also need to be done to the evolution dashboard since I believe it is expecting numbered proposals.)

I think a better solution would be to change the message of the validation report when the status is 'awaiting review' so that it does not report specific issues, but instead presents a message that a proposal cannot pass CI until it is ready to be committed and moves to 'active review' status.

It lets the author know there is not something they need to do to get it to pass, it just isn't in a state where it is ready to be committed and the CI validation prevents that (as it should).

dempseyatgithub: Another thought might be to keep proposal PRs in draft status until a review manager is ready to commit a proposal. Until details like the proposal number and review manager are added, the proposal is essentially a draft. We could potentially only validate PRs that are not drafts.

ktoso: Thanks for chiming in James.

I understand wanting to keep the tool simple, but it does mean that folks opening a proposal PR cannot get it green basically and this is _very weird_ to a contributor to sit around on a failing PR and know that's "right".

When a review manager schedules a proposal they make an edit to give it a number and give it the "In review" status; so when they do this change all the existing checks would still work correctly. I do think we need to not apply those checks for "Awaiting review" status.

> [...] keep proposal PRs in draft status until a review manager is ready to commit a proposal.

So the draft idea would have to change how LSG look at proposals; currently a proposal in draft status is not being looked at basically; and removing it from draft is effectively asking for scheduling a review AFAIK.

@swiftlang/language-steering-group would need to decide what process they want here.

PR #3429, opened 2026-08-20
Proposal: `@called(once)` as a way to state that a function is called…

xedin:
… at-most once

This proposal introduces a `@called(once)` attribute to model at-most once execution of a function. Making function execution behavior visible to the compiler has two advantages. First, it makes it possible to diagnose incorrect usage of such functions. But, it also allows the compiler to understand how closure captures could affect variable lifetimes. This can improve compatibility with non-Copyable types.

### DIFF
--- a/proposals/NNNN-called-once-functions.md
+++ b/proposals/NNNN-called-once-functions.md
@@ -187,6 +187,8 @@ func consumingCapture(s: consuming S) {
   }
 }
 ```
+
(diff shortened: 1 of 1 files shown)


xwu:
FYI: [SE-0073](https://github.com/swiftlang/swift-evolution/blob/2074c85921e17b91ca807d8ebc942846ce09162e/proposals/0073-noescape-once.md).

PR #3448, opened 2026-08-26
SE-0538 Disconnected: Rename take/swap to consume/exchange

FranzBusch:
Per the "accepted with modifications" review outcome, rename `take()` to `consume()` and `swap(newValue:)` to `exchange(newValue:)`


hborla (approved):
Thank you!

PR #3454, opened 2026-08-28
[SE-0431] Clarify that closure thrown type inference works the same in Swift 5 and 6 modes

ole:
Delete "in Swift 5" from a sentence under Future directions / Closure thrown type inference that (wrongly, I believe) suggests that the described behavior is only applied in Swift 5 language mode.

Even though this sentence appears under Future directions, it describes the actual behavior as it's currently implemented.

I believe the original wording is a remnant of the originally accepted proposal with the `FullTypedThrows` upcoming feature that was later removed in #2535. Since the type inference story for typed throws is the same in Swift 5 and 6 modes (as far as I know), the wording should reflect this.

### DIFF
--- a/proposals/0413-typed-throws.md
+++ b/proposals/0413-typed-throws.md
@@ -1169,7 +1169,7 @@ This inference rule will change the thrown error types of existing closures that
 }
 ```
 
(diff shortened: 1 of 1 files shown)


DougGregor (approved):
Thank you, @ole! This is indeed a missed edit from when we removed `FullTypedThrows`.

PR #3478, opened 2026-09-08
[Proposal] Deployment target conditional compilation

Jiaxu-Li:
This proposal introduces `#if deploymentTargetAtLeast(...)`, a conditional compilation condition based on the minimum deployment target of the module being compiled.

Pitch discussion: https://forums.swift.org/t/pitch-conditional-compilation-for-the-deployment-target/89282

Prototype implementations:

- https://github.com/Jiaxu-Li/swift/tree/deployment-target-ifconfig
- https://github.com/Jiaxu-Li/swift-syntax/tree/deployment-target-ifconfig


proposals/NNNN-deployment-target-conditional-compilation.md
```diff
@@ -0,0 +1,313 @@
...
+target and keep the two values in sync. If they diverge, the selected source no
+longer reflects the binary's minimum OS requirement. The compiler already has
+this information in the target triple.
+
+This is distinct from an SDK check: two modules can use the same SDK while
+targeting different minimum OS versions.
+
+## Proposed solution
+
+Introduce `deploymentTargetAtLeast(...)` as a condition accepted by `#if`:
```
xwu, line 67: There's some question as to whether this terminology would be best aligned with existing usage as `minDeploymentTarget`, and/or if there might be confusion with "target" as it's used in SwiftPM and whether the use of "deployment" here breaks new ground within Swift itself. But this can all be discussed during review, so no need to hash it out at this point.

proposals/NNNN-deployment-target-conditional-compilation.md
```diff
@@ -0,0 +1,313 @@
...
+// corresponding requirement, or when the active platform is not listed.
+#else
+// Compiled when the active platform's minimum deployment target is earlier
+// than the corresponding requirement.
+#endif
+```
+
+The arguments use the platform-version spelling of availability queries. Only
+the active platform's requirement is tested. The final `*` leaves unlisted
+platforms unrestricted.
```
xwu: This should use the wording that `*` results in `true` for all other platforms.

proposals/NNNN-deployment-target-conditional-compilation.md
```diff
@@ -0,0 +1,313 @@
...
+deployment-target-requirements ->
+  deployment-target-requirements ',' deployment-target-requirement
+
+deployment-target-requirement -> identifier version
+version -> decimal-digits
+version -> version '.' decimal-digits
+```
+
+A trailing comma is permitted. The required wildcard may appear only once and
+must be last. `deploymentTargetAtLeast(*)` is valid and always evaluates to
```
xwu: I'm not sure that your grammar above allows for a trailing comma, nor (as far as I recall) is trailing comma allowed in other places that require trailing terminal wildcard `*`, so I don't quite understand why it's mentioned here.

proposals/NNNN-deployment-target-conditional-compilation.md
```diff
@@ -0,0 +1,313 @@
...
+non-Apple platforms.
+
+As with existing `anyAppleOS` availability, versions earlier than 26 produce a
+warning because Apple OS version numbers are aligned only beginning with version
+26.
+
+An unrecognized platform name produces a warning and cannot match a target
+known to that compiler, so the wildcard determines the result. This allows
+older compilers to encounter future platform names while still diagnosing
+likely misspellings.
```
xwu: It's not clear what, if anything, here diverges from `if #available(...)` syntax. If anything does diverge, that should be called out explicitly. Anything that doesn't diverge could be covered under just a statement along the lines that it's the same syntax and behavior as in `is #available`.

How Apple chooses to classify its OSes, including `anyAppleOS`, etc., is reserved to Apple itself and not subject to Swift Evolution, so it doesn't belong as a design detail here.

proposals/NNNN-deployment-target-conditional-compilation.md
```diff
@@ -0,0 +1,313 @@
...
+
+If a requirement explicitly names the active platform but the compiler cannot
+obtain a deployment version for it, compilation fails:
+
+```swift
+// Error when compiling for Linux because the target has no deployment version.
+#if deploymentTargetAtLeast(Linux 6, *)
+// ...
+#endif
+```
```
xwu, line 180: Similar comment here: does this diverge from `is #available(...)` behavior? if so, why?

Jiaxu-Li:
Thanks for the review. I've pushed an update.

I fixed the wildcard wording and removed the trailing comma sentence, along with the `anyAppleOS` details.

For platform matching, a requirement now applies only when its platform names the target, with the most specific one winning when several match. I've written out where this differs from `if #available(...)` and why.

The Linux case also explains why it's diagnosed.

Happy to revisit the spelling during review.

proposals/NNNN-deployment-target-conditional-compilation.md
```diff
@@ -0,0 +1,306 @@
...
+condition evaluates to `true`.
+
+This differs from `if #available(...)`, where a requirement written for one
+platform can also describe another that inherits its availability. Applying an
+iOS requirement to visionOS is meaningful there only because the compiler
+translates the version through a mapping supplied by the SDK. Conditional
+compilation is evaluated before that information is available, so this condition
+matches on the platform name alone. Compiling for visionOS with a requirement
+written only for iOS therefore falls to `*`, and a requirement naming
+`visionOS` is needed to test a visionOS deployment target.
```
xwu, line 153: Thanks @Jiaxu-Li! Overall, looks good.

I discussed this point with the rest of the steering group. Our understanding is that it ought to be possible to maintain the same user-facing behavior as `if #available(...)` by obtaining this mapping either with _an_ SDK (not necessarily a specific SDK version) and/or to hardcode it if necessary, but that in any case the exact hierarchy of platform inheritance and the engineering responsibility would be up to the platform owner (i.e., Apple) and not a concern for Swift Evolution.

So the suggestion here would be that this document then propose to maintain the same user-facing behavior as `if #available(...)` without further specification of how that is to be achieved on Apple platforms. If that seems reasonable to you, I think we are ready to kick off review with that change.

Jiaxu-Li: Thanks, I’ve updated the proposal to specify the same user-facing platform behavior as `if #available(...)`, leaving the platform hierarchy and any necessary version mapping to platform owners. I’ve updated both prototypes, and the targeted tests pass. The proposal is ready for review from my side.

xwu:
Great! Sorry last thing: would you mind creating pull requests against main with your prototypes. They can stay in draft mode and won't be merged as-is, but this would allow us to build test toolchains for people to try out. But I agree the proposal text looks ready! Thanks so much

Jiaxu-Li:
Thanks! I’ve opened draft PRs against `main` for both prototypes:

- Compiler: swiftlang/swift#92598
- SwiftSyntax: swiftlang/swift-syntax#3450

PR #3485, opened 2026-09-14
Correct SE-0549 review discussion formatting

dempseyatgithub:
Small correction to normalize formatting of the Review discussion links.

// @FranzBusch


FranzBusch (approved):
Would be great if the CI check would catch this.

PR #3493, opened 2026-09-16
Add release date for Swift 6.4

denizmaral:
Updated the release date for Swift 6.4 in the README.


xwu:
Thanks! Unfortunately, it's missing the terminating `|` for that table row, so we'll merge #3494.

PR #3495, opened 2026-09-16
Swift 6.4 release date

denizmaral:


xwu:
Thanks! As I said earlier, though, #3494 addressed the same issue correctly first, so that one's been merged.

